PR #18, opened 2019-02-18
ResultSet timeseries fix

keydunov:
Fix `ResultSet.timeSeries` for granularity hour.


keydunov:
@paveltiunov I've updated it to be aligned with code in Compiler.

PR #29, opened 2019-02-23
 Introduce option: devServer

dtslvr:


packages/cubejs-server-core/core/index.js
```diff
@@ -81,6 +81,7 @@ class CubejsServerCore {
       driverFactory: () => CubejsServerCore.createDriver(options.dbType),
       apiSecret: process.env.CUBEJS_API_SECRET,
       dbType: process.env.CUBEJS_DB_TYPE,
+      devServer: true,
```
paveltiunov: To ensure backward compatibility should be `devServer: process.env.NODE_ENV !== 'production'`

packages/cubejs-server-core/core/index.js
```diff
@@ -162,7 +163,9 @@ class CubejsServerCore {
       this.logger
     );
     apiGateway.initApp(app);
-    this.initDevEnv(app);
+    if (this.options.devServer) {
```
paveltiunov: There're already these devServer checks in place. I believe we should just replace all `process.env.NODE_ENV !== 'production'` occurrences with `this.options.devServer`.

dtslvr: I agree, please see https://github.com/statsbotco/cube.js/pull/29/files#diff-49fd186d45485663b0e641a0214512e2R102 (Line 102)

packages/cubejs-server-core/core/index.js
dtslvr: Done.

dtslvr:
Please check my improvements after the code review.

packages/cubejs-server-core/core/index.js
```diff
@@ -162,45 +163,45 @@ class CubejsServerCore {
...
+      this.initDevEnv(app);
+    }
   }
 
   initDevEnv(app) {
-    if (process.env.NODE_ENV !== 'production') {
```
dtslvr: I have removed this check, I prefer it at the function call.

paveltiunov:
Looks great!

PR #33, opened 2019-02-26
Harmonize quotes in README

dtslvr:


paveltiunov:
Hi @dtslvr ! Thanks for posting this! Quotes should be linted across documentation for sure! The only thing: in Cube.js schema default code style for quotes is a back tick. And reason for it is most of declarations reference other members using string template literals which require it. 

So could you please revert quote changes for schema? Otherwise it's fine.

dtslvr:
Okay. I have reverted the quotes in the schema, @paveltiunov.

My linter is always complaining about string literals that do not contain a variable :stuck_out_tongue_closed_eyes:

paveltiunov:
🙌

PR #48, opened 2019-03-13
add support for MongoBI Driver

krunalsabnis:
- add mongobi-driver
- support SSL configuration for client connection
- tested with mongodb Atlas running mongodb 3.6 with BI Connectors enabled


krunalsabnis:
@paveltiunov  I have added Driver dependencies but that's not enough, I need to see where does it link it. create project with dbtype = mongobi does not work

paveltiunov:
@krunalsabnis I believe it should be enough when it's published. What doesn't work for you? 
If it's 
```
$ cubejs create project -d mongobi
``` then it's ok for now. To test it you should just link all the dependencies manually through `npm link` or `yarn link`.

File: packages/cubejs-mongobi-driver/driver/MongoBIDriver.js [hunk 1]
```diff
@@ -0,0 +1,126 @@
...
+  constructor(config) {
+    super();
+    this.config = {
+      host: process.env.CUBEJS_DB_HOST,
+      database: process.env.CUBEJS_DB_NAME,
+      port: process.env.CUBEJS_DB_PORT,
+      user: process.env.CUBEJS_DB_USER,
+      password: process.env.CUBEJS_DB_PASS,
+      ssl: {
+	      ssl: process.env.USE_SSL
```
paveltiunov: I guess for naming consistency we should use `CUBEJS_DB_SSL` here instead or something like it.

packages/cubejs-mongobi-driver/driver/MongoBIDriver.js
```diff
@@ -0,0 +1,126 @@
...
+      },
+      authSwitchHandler: (data, cb) => {
+        var buffer = Buffer.fromprocess.env.CUBEJS_DB_PASS.concat('\0')();
```
paveltiunov: There's probably some parenthesis problem.

paveltiunov:
@krunalsabnis Looks great! Thanks for contributing this!

PR #51, opened 2019-03-16
Add example for how to disable security

dtslvr:


paveltiunov:
@dtslvr Hey Thomas! Makes sense to add it. Thanks for contribution!

PR #53, opened 2019-03-16
Make API path namespace configurable

dtslvr:
Now it's possible to customize your route like
- `http://localhost:8080/cubejs-api/v1/load?query=...` (default)
- `http://localhost:8080/my-custom-route/v1/load?query=...`

I'm still not fully satisfied with the versioning. In my opinion the endpoint should be like this `http://localhost:8080/my-custom-route/v1/cubejs/load?query=...` where `cubejs` should also be customizable.

But this format is not possible at the moment 🤔 How could we solve this?

### DIFF
--- a/docs/Cube.js-Backend/@cubejs-backend-server-core.md
+++ b/docs/Cube.js-Backend/@cubejs-backend-server-core.md
@@ -19,7 +19,7 @@ Create an instance of `CubejsServerCore` to embed it in an `Express` application
     * `logger(msg, params)` - Pass function for your custom logger.
     * `schemaPath` - Path to the `schema` location. By default, it is `/schema`.
     * `devServer` - Enable development server. By default, it is `true`.
-    * `baseRoute` - Route where _Cube.js_ is mounted to. By default, it is `/cubejs-api`.
+    * `basePath` - Path where _Cube.js_ is mounted to. By default, it is `/cubejs-api`.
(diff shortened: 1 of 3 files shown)


paveltiunov:
@dtslvr Hey Thomas! Seems good to me. The only thing: `basePath` would be probably more usual name for the option. 

Regarding versioning: I believe Cube.js API should be versioned separately from other APIs as it will be usually upgraded just by Cube.js version upgrade.

dtslvr:
Thanks for the feedback, @paveltiunov. What do you think about adding the version to the default base path?

`this.basePath = options.basePath || '/cubejs-api/v1';`

Then the consumer is completely free to override the version if needed.

I will change `baseRoute` to `basePath`.

paveltiunov:
@dtslvr It's a great question. Actually original idea was to optionally have several endpoints for different API versions at the same time. My guess is API versions in this case should be controlled by server core code. What do you think?

dtslvr:
Yes, I agree. I have refactored `baseRoute` to `basePath` now.

Adapting the url pattern from `http://localhost:8080/cubejs-api/v1/load?query=...` to `http://localhost:8080/api/v1/cubejs/load?query=...` would cause breaking changes because the frontend client also needs adaption. Let's keep it for the moment.

paveltiunov:
@dtslvr Hey Thomas! Looks great! Thanks for contributing this!

PR #65, opened 2019-04-01
Vue package

richipargo:
Wanted to also support Vue.js package API wrapper with same capabilities as React, although the implementation is a simple one it might need some work.

Also had to build ES modules version instead of commonJS since `vue-cli` didn't like it.

Any comments are welcome


paveltiunov:
@richipargo Oh. Wow! That's really great! Would love to see any doc examples / tutorials on how to use this. Thanks for contributing!

PR #70, opened 2019-04-05
feat: Vue.js reactivity on query update

richipargo:
Had some issues with query reactivity, fixed with computed properties and watchers


paveltiunov:
Hey Ricardo! Looks good! I'm curious why did you add `set` methods for members? Could you please elaborate on a use case? Is there any chance same behaviour can be achieved by updating query?

richipargo:
hey @paveltiunov some multiselect controls pass the whole array and would be easier to manage. Also I need to either update the whole query object or force reactivity methods for existing query. Either way i'll probably make some change further on

paveltiunov:
@richipargo Hey Ricardo! Yep. Makes sense. Very interesting case. Is this multi select used only for measures, dimensions? Or is it also for time dimensions or filters?

PR #72, opened 2019-04-05
fix typo in cubejs example docs

davidawad:


paveltiunov:
@davidawad Hey David! Thanks for fixing this! Actually repo name here's a little bit outdated: current is git@github.com:statsbotco/cube.js.git.

PR #74, opened 2019-04-06
Fix typo

benbankes:


paveltiunov:
@benbankes Nice catch! Thanks for fixing this!

PR #78, opened 2019-04-12
fix(mongobi-driver): fix ssl configuration

vladi-dev:
Fixes https://github.com/statsbotco/cube.js/issues/77

Added environment variables for MONGO BI SSL configuration:
- `CUBEJS_DB_SSL`
- `CUBEJS_DB_SSL_CA`
- `CUBEJS_DB_SSL_CERT`
- `CUBEJS_DB_SSL_CIPHERS`
- `CUBEJS_DB_SSL_PASSPHRASE`
- `CUBEJS_DB_SSL_REJECT_UNAUTHORIZED`


paveltiunov:
Hey @vladi-dev ! Everything looks great! Thanks for contributing this! The only thing I noted: can we revert back `process.env.CUBEJS_DB_SSL` variable behaviour for backward compatibility? So when it's passed there's `{}` passed as ssl object and when it isn't then nothing is passed as ssl? Otherwise it seems existing deployments without SSL will fail?

paveltiunov:
@vladi-dev Looks great! Thanks!

PR #85, opened 2019-04-17
fix(query-orchestrator): use bluebird to promisify redis client

### LINKED ISSUE #84: RedisQueueDriver mutates EventEmitter prototype indirectly
**Describe the bug**
Configuration of redisClient is using the package `util-promisifyall` to promisify all redisClient methods (on redisClient.prototype) but it also promisifies redisClient protototype parent one level up the chain, which is EventEmitter's prototype since redisClient inherits EventEmitter class.

Calling `promisifyAll`:
https://github.com/statsbotco/cube.js/blob/50f1bbb7aac34ae7c4fd82a86ecee8506b17f4ea/packages/cubejs-query-orchestrator/orchestrator/RedisQueueDriver.js#L2-L7

`redisClient` inherits `EventEmitter`:
https://github.com/NodeRedis/node_redis/blob/12265a5079a133d2003bef9cdb0f2deee0251518/index.js#L189

Implementation of `utils-promisifyAll` takes methods of the input class and creates copies with `Async` suffix that return a promisified version of the original method.

For me the problem arises when we are trying to use other libraries that use the concept of promisifying all.

For example `node-soap`:

https://github.com/vpulim/node-soap/blob/0efafa219be15b0efac84c9228c161b8662f720f/lib/client.js#L22-L35

Here you can see promisification with `Bluebird.promisifyAll` of the `Client` class which inherits `EventEmitter` but since `EventEmitter` has been mutated it adds `xxxAsync` methods to `Client`'s prototype and then `Bluebird.promisifyAll` errors out with message: `Cannot promisify an API that has normal methods` - http://bluebirdjs.com/docs/error-explanations.html#error-cannot-promisify-an-api-that-has-normal-methods

Here is the code that throws the error:
https://github.com/petkaantonov/bluebird/blob/17f69f3b0fa89a0b7a5cfb9a6546d1180a3610e0/src/promisify.js#L49-L65

So to solve that issue for me and possibly for other people that use `Bluebird.promisifyAll` I will open a PR to use `Bluebird` instead of `util-promisifyall` if that is okay with you. 
I hope it is not an issue since `Bluebird` is already in some of sub-dependencies of this package.

**To Reproduce**
Sandbox: https://codesandbox.io/embed/rmjmmnn73p

Or code:

```
'use strict';

const util = require('util');
const { EventEmitter } = require('events');

const promisifyAll = require('util-promisifyall');

function RedisClient() {}

util.inherits(RedisClient, EventEmitter);

require('bluebird').promisifyAll(RedisClient.prototype);

console.log(Object.keys(EventEmitter.prototype));

promisifyAll(RedisClient.prototype);

console.log(Object.keys(EventEmitter.prototype));
```

**Expected behavior**
Expect that using the library does not modify prototype of the global class `EventEmitter`.

**Version:**
0.7.0

vladi-dev:
fixes #84

### DIFF
--- a/packages/cubejs-query-orchestrator/orchestrator/RedisQueueDriver.js
+++ b/packages/cubejs-query-orchestrator/orchestrator/RedisQueueDriver.js
@@ -1,5 +1,5 @@
 const redis = require('redis');
-const promisifyAll = require('util-promisifyall');
+const { promisifyAll } = require('bluebird');
 const R = require('ramda');
 
 const BaseQueueDriver = require('./BaseQueueDriver');
--- a/packages/cubejs-query-orchestrator/package.json
+++ b/packages/cubejs-query-orchestrator/package.json
@@ -10,9 +10,9 @@
     "test": "mocha"
   },
   "dependencies": {
+    "bluebird": "^3.5.4",
     "ramda": "^0.24.1",
-    "redis": "^2.8.0",
-    "util-promisifyall": "^1.0.4"
+    "redis": "^2.8.0"
   },
   "devDependencies": {
     "eslint": "^4.14.0",
--- a/packages/cubejs-query-orchestrator/yarn.lock
+++ b/packages/cubejs-query-orchestrator/yarn.lock
@@ -82,6 +82,11 @@ balanced-match@^1.0.0:
   resolved "https://registry.yarnpkg.com/balanced-match/-/balanced-match-1.0.0.tgz#89b4d199ab2bee49de164ea02b89ce462d71b767"
   integrity sha1-ibTRmasr7kneFk6gK4nORi1xt2c=
 
+bluebird@^3.5.4:
+  version "3.5.4"
+  resolved "https://registry.yarnpkg.com/bluebird/-/bluebird-3.5.4.tgz#d6cc661595de30d5b3af5fcedd3c0b3ef6ec5714"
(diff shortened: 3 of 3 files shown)


paveltiunov:
Hey @vladi-dev ! Thanks for providing this PR! I believe bluebird fixes your original issue. On other hand bluebird will replace/patch native Promise implementation which entails other similar issues that you faced already. I believe it's better just not to use any `promisifyAll` stuff here and make async methods manually one by one as list of those used methods is known.

vladi-dev:
@paveltiunov do you mean having something like this?

```javascript
const { promisify } = require('util');
['brpop', 'del', 'exec', 'get', 'hget', 'rpop', 'set', 'zadd', 'zrange', 'zrangebyscore'].forEach(
  k => promisify(redis.RedisClient.prototype[k])
);
promisify(redis.Multi.prototype.exec);
```

paveltiunov:
@vladi-dev Yep. Exactly! Probably it also makes sense to `promisify` instances and not prototypes to avoid side effects.

vladi-dev:
@paveltiunov please take a look at https://github.com/statsbotco/cube.js/pull/89

vladi-dev:
Implemented manual promisify in https://github.com/statsbotco/cube.js/pull/89

PR #100, opened 2019-05-02
fix merging config with custom queryOptions which were not passing al…

cameronbraid:
Fixed the merging of the config to use the correct database name

### DIFF
--- a/packages/cubejs-clickhouse-driver/driver/ClickHouseDriver.js
+++ b/packages/cubejs-clickhouse-driver/driver/ClickHouseDriver.js
@@ -17,12 +17,14 @@ class ClickHouseDriver extends BaseDriver {
       ...config
     };
     this.pool = genericPool.createPool({
-      create: () => new ClickHouse(Object.assign({}, this.config, {
+      create: () => new ClickHouse({
+        ...this.config,
         queryOptions: {
           join_use_nulls: 1,
-          session_id: uuid()
+          session_id: uuid(),
+          ...this.config.queryOptions,
         }
-      })),
+      }),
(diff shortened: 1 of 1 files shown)


packages/cubejs-clickhouse-driver/driver/ClickHouseDriver.js
```diff
@@ -17,12 +17,14 @@ class ClickHouseDriver extends BaseDriver {
...
     };
     this.pool = genericPool.createPool({
-      create: () => new ClickHouse(Object.assign({}, this.config, {
+      create: () => new ClickHouse({
+        ...this.config,
         queryOptions: {
           join_use_nulls: 1,
-          session_id: uuid()
+          session_id: uuid(),
+          ...this.config.queryOptions,
```
inikolaev: This would allow to override `join_use_nulls` and `session_id`. Should it be moved above `join_use_nulls` instead to prevent this?

cameronbraid: I had thought about that, however I coded it with the view that it allows someone who knows what they are doing to to override this intentionally.

paveltiunov:
@cameronbraid Hey Cameron! Looks great! I'm curious is it for multi-tenancy? Are you passing database name in `queryOptions` instead of as env variable?

cameronbraid:
@paveltiunov queryOptions.database is how the clickhouse driver identifies the database for the current query.

paveltiunov:
@cameronbraid Nvm. I see these `queryOptions` just wasn't passed 👍

PR #118, opened 2019-06-03
Replacing ${TABLE} instances to ${CUBE}

avin3sh:
There is some inconsistency in the docs where at certain places the old `${TABLE}` reference is used, whereas at other places the current `${CUBE}` reference is used


paveltiunov:
@avin3sh Hey Avinesh! Thanks for providing this! Definitely need to fix this.

PR #122, opened 2019-06-05
Bug Fixes:

dgkmurthy:
1.  Schema generator bug with case sensitive Table and Column names
2.  Column name ambiguity issue during table joins with similar column names fix by forcing the alias for column names
3.  Query bug fix to mandatorily include double quotes for the alias when running query against table names with restricted keywords such as "case"


dgkmurthy:
To split the PRs

Issue #120, opened 2019-06-04
Postgres database schema generation not maintaining case sensitivity of names and not including the joins based on the referential integrity


dgkmurthy:
**Describe the bug**
When generating the schema in the Playground or CLI selecting multiple table names on Postgres is not generating the schema file having the joins automatically based on the referential integrity/foreign key columns.

> In addition, schema is not adding quotes to maintain case sensitivity in both the Table name and column names.

**To Reproduce**
Create tables and columns having case sensitive names with double quotes.

```
CREATE TABLE "public"."Account"
(
    "id" serial primary key,
    "departmentId" integer REFERENCES "public"."Department"("id") ON DELETE CASCADE,
    "firstName" text,
    "lastName" text,
    "dateOfBirth" date,
    "email" text,
    "status" text,
    "createdAt" timestamptz not null default now(),
    "updatedAt" timestamptz not null default now(),
    "archiveAt" timestamptz not null default now()
)

CREATE TABLE "public"."Department"
(
    "id" serial primary key,
    "name" text,
    "description" text,
    "createdAt" timestamptz not null default now(),
    "updatedAt" timestamptz not null default now(),
    "archiveAt" timestamptz not null default now()
)
```

**Steps to reproduce the behavior**

1. Connect cube.js to 'Postgres' database
2. Go to 'Playground'
3. Click on 'Database or Tables'
4. Select all the 'Tables' then click on generate schema
5. Click on 'Schema' and open the Schema file

there you see no entries for joins property.

**Expected behavior**

1. to maintain the case sensitivity of the table and column names by adding single quotes to the name property values in the schema.  

> E.g., "Account" for table name and "accountId", "firstName", "lastName", "dataOfBirth" for column names.

2. To populate the joins automatically primary, referential integrity columns in a mixed case sensitivity setup.  

> E.g., having table name as "Account" and reference column as "accountId" rather than "AccountId".

**Version:**
Latest

paveltiunov:
@dgkmurthy Hey Keshava! Thanks for posting this! Yep. Here we should have lowercase comparison: 
https://github.com/statsbotco/cube.js/blob/a53f0136e3efd2c4cf0ad52bec033319e6966438/packages/cubejs-schema-compiler/scaffolding/ScaffoldingSchema.js#L155

dgkmurthy:
Fixed in #122

craigharman:
This still seems to be an issue in v0.18.3 with time dimensions and Postgres.
[...]

PR #124, opened 2019-06-05
Schema generation with joins having case sensitive table and column names

dgkmurthy:
#120


packages/cubejs-schema-compiler/scaffolding/ScaffoldingSchema.js
```diff
@@ -148,10 +148,15 @@ class ScaffoldingSchema {
   };
 
   joins(tableName, tableDefinition) {
+    const key = "id";
     return R.unnest(tableDefinition
-      .filter(column => column.name.toLowerCase().indexOf('id') !== -1 && column.name.toLowerCase() !== 'id')
+      .filter(column => {
+        const columnName = column.name.toLowerCase();
+        const columnKeyPosition = columnName.length - columnName.lastIndexOf(key) - key.length;
+        return (columnKeyPosition === 0 && columnName !== 'id')
```
paveltiunov: This is right but missing `_id` case I guess.

packages/cubejs-schema-compiler/scaffolding/ScaffoldingTemplate.js
```diff
@@ -40,16 +40,16 @@ class ScaffoldingTemplate {
   schemaDescriptorForTable(tableSchema) {
     return {
       cube: tableSchema.cube,
-      sql: `SELECT * FROM ${tableSchema.schema}.${tableSchema.table}`, // TODO escape
+      sql: `SELECT * FROM "${tableSchema.schema}"."${tableSchema.table}"`, // TODO escape
```
paveltiunov: We can't use `"` here because not all backends support it. Need use escape identifier methods.

packages/cubejs-schema-compiler/scaffolding/ScaffoldingTemplate.js
```diff
@@ -40,16 +40,16 @@ class ScaffoldingTemplate {
...
       joins: tableSchema.joins.map(j => ({
         [j.cubeToJoin]: {
-          sql: `\${CUBE}.${j.thisTableColumn} = \${${j.cubeToJoin}}.${j.columnToJoin}`,
+          sql: `\${CUBE}."${j.thisTableColumn}" = \${${j.cubeToJoin}}."${j.columnToJoin}"`,
```
paveltiunov: Same as before

packages/cubejs-schema-compiler/scaffolding/ScaffoldingTemplate.js
```diff
@@ -40,16 +40,16 @@ class ScaffoldingTemplate {
...
           relationship: j.relationship
         }
       })).reduce((a, b) => ({ ...a, ...b }), {}),
       measures: tableSchema.measures.map(m => ({
         [this.memberName(m)]: {
-          sql: m.name,
+          sql: `"${m.name}"`,
```
paveltiunov: Same here. Would be also great to check if we actually need to escape before escaping to keep schema clean.

packages/cubejs-schema-compiler/scaffolding/ScaffoldingTemplate.js
```diff
@@ -61,7 +61,7 @@ class ScaffoldingTemplate {
       }),
       dimensions: tableSchema.dimensions.map(m => ({
         [this.memberName(m)]: {
-          sql: m.name,
+          sql: `"${m.name}"`,
```
paveltiunov: Same here.

packages/cubejs-schema-compiler/scaffolding/ScaffoldingSchema.js
paveltiunov: Probably just makes sense to use `RegExp` everywhere.

dgkmurthy: Correct, RegExp similar to done to the below would simplify.

packages/cubejs-schema-compiler/scaffolding/ScaffoldingTemplate.js
dgkmurthy: under BigQuery.js the way you doing escape for the column name is as follows:

```
escapeColumnName(name) {
    return `"${name}"`;
  }
```

which is no different to the way it is done now other than it's inline.  Are you looking anything differently.

packages/cubejs-schema-compiler/scaffolding/ScaffoldingSchema.js
dgkmurthy: Updated,  please check if it works.

packages/cubejs-schema-compiler/scaffolding/ScaffoldingTemplate.js
dgkmurthy: Updated,  please check if it works.

packages/cubejs-schema-compiler/scaffolding/ScaffoldingTemplate.js
dgkmurthy: Updated,  please check if it works.

packages/cubejs-schema-compiler/scaffolding/ScaffoldingTemplate.js
dgkmurthy: Updated,  please check if it works.

packages/cubejs-schema-compiler/scaffolding/ScaffoldingTemplate.js
dgkmurthy: Updated,  please check if it works.

packages/cubejs-schema-compiler/scaffolding/ScaffoldingTemplate.js
paveltiunov: Yep. Right. For BigQuery it's `"`. On other hand MySQL supports only `\``.

paveltiunov:
I think we're good to merge this. Will add some minor tweaks for MySQL support then.

PR #125, opened 2019-06-05
Query Join ambiguity with similar column names

dgkmurthy:
Column name ambiguity issue during table joins with similar column names fix by forcing the alias for column names


dgkmurthy:
Will be combined with #126

dgkmurthy:
Query Join ambiguity with similar column names

PR #126, opened 2019-06-05
Query join ambiguity while using similar column names and alias having table names with restricted keywords

dgkmurthy:
Query bug fix to mandatorily include double quotes for the alias when running query against table names with restricted keywords such as "case"


paveltiunov:
@dgkmurthy Hey Keshava! Could you please rebase it and drop already merged in changes?

PR #127, opened 2019-06-06
feat: Added SparkSQL and Hive support to the JDBC driver.

QuantamHD:
This will allow cubejs to connect directly to
SparkSQL and Hive datastores running big
data queries.

### DIFF
--- a/packages/cubejs-jdbc-driver/driver/JDBCDriver.js
+++ b/packages/cubejs-jdbc-driver/driver/JDBCDriver.js
@@ -56,6 +56,34 @@ const DbTypes = {
       PWD: process.env.CUBEJS_AWS_SECRET,
       S3OutputLocation: process.env.CUBEJS_AWS_S3_OUTPUT_LOCATION
     }
+  },
+  sparksql: {
+    driverClass: "org.apache.hive.jdbc.HiveDriver",
+    prepareConnectionQueries: [],
+    mavenDependency: {
+      "groupId": "org.apache.hive",
+      "artifactId": "hive-jdbc",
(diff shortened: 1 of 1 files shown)


paveltiunov:
@QuantamHD Hey Ethan! This is great! Thanks for contributing it! I'll merge it right away. The only thing we'd also need hive and sparksql dialects in order to make it work. This is how to do it: https://github.com/statsbotco/cube.js/blob/master/CONTRIBUTING.md#implementing-sql-dialect

kamranayub:
@QuantamHD interested in using this! Do you need help implementing the dialect? I don't know anything about querying Spark/Hive but I can help out if you need it.

QuantamHD:
@kamranayub Absolutely, I would love help with understanding how the code generation is organized on the cubejs side. SparkSQL and HiveQL support a pretty wide range of the SQL standard from CTEs, and windowing functions, etc. I was looking over the Postgres dialect and it seemed to only override timezone functions so it seems as though BaseQuery handles a lot of things. 

I'm just wondering if this is due to lack of support for that dialect in particular or if BaseQuery is just pretty good out of the box.

QuantamHD:
One thing of particular interest as well is that HiveQL and SparkSQL are nearly identical in terms of syntax so I was thinking of implementing a HiveQL dialect and then for sparkSQL just extending off of HiveQL to reduce the maintenance requirements

kamranayub:
@QuantamHD are you in the cubejs Slack? Perhaps could start a thread/channel/discussion there too.

QuantamHD:
I am not what's the link to it?

PR #130, opened 2019-06-09
Fix to discard the FR Columns from inclusion in the measures

dgkmurthy:
Fix to discard the Foreign Reference Columns from inclusion in the measures while generating the schema

### DIFF
--- a/packages/cubejs-schema-compiler/scaffolding/ScaffoldingSchema.js
+++ b/packages/cubejs-schema-compiler/scaffolding/ScaffoldingSchema.js
@@ -117,10 +117,9 @@ class ScaffoldingSchema {
   }
 
   numberMeasures(tableDefinition) {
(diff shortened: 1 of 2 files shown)


paveltiunov:
@dgkmurthy Hey Keshava! Thanks for posting this! Could you please provide some example use case which is fixed by this change? Preferably in the form of unit test.

dgkmurthy:
> @dgkmurthy Hey Keshava! Thanks for posting this! Could you please provide some example use case which is fixed by this change? Preferably in the form of unit test.

Updated the existing unit tests to cover this change

dgkmurthy:
The last change will ensure that the columns having 'id' suffix not included in the measures, and also, the measure fields to match dictionary for suffix only not between.

paveltiunov:
Looks good! Thanks!

PR #133, opened 2019-06-13
Extend the documentation about the refreshKey property and getting real time responsiveness to work

airhorns:
Prompted by https://github.com/statsbotco/cube.js/issues/132

While I was writing this too, it seems like understanding all of this to get the real time use case is asking users to do a fair amount of digging. That suggests it might be good to expose an option like `skipCaching: true` or `forceQuery: true` or something like that that just ensures the query is re-run every time. I also think that if developers know that their underlying data is rapidly changing (like in the real time use case), executing the `refreshKey` query is actually not very useful, because it will always return something different, so being able to control that with the same option would be useful.

### DIFF
--- a/docs/Cube.js-Backend/Caching.md
+++ b/docs/Cube.js-Backend/Caching.md
@@ -22,28 +22,40 @@ is found it executes the query in database, returns result back alongside writin
 If Cube.js detects existing cache entry for the key it schedules a background cache refresh check.
 
 
-### Refresh Strategy
+### Refresh Keys
 
-Background refresh check decides whether the cache entry should be refreshed or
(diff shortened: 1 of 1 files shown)


paveltiunov:
@airhorns Thanks for contributing this! Yep. We're thinking a lot about exposing force no cache but the hard part here is unexpected behaviour occurs when pre-aggregations are enabled. We just can't invalidate pre-aggregations based on single query because each pre-aggregation can serve multiple queries. And while it works well with in-memory step pre-aggregation users will get into the same trap of cached results even with force no cache enabled. On other hand there's `refreshKey` for pre-aggregations as well and `renewQuery` forces to update pre-aggregations if required.

PR #148, opened 2019-07-01
feat: Lean more on vue slots for state

richipargo:
Since Vue slots are useful to show different parts and states I'm leveraging on a better solution to how queries are loaded. Also covered some areas with unit testing

### DIFF
--- a/packages/cubejs-vue/src/QueryRenderer.js
+++ b/packages/cubejs-vue/src/QueryRenderer.js
@@ -58,6 +58,9 @@ export default {
         resultSet,
         sqlQuery,
         query: this.builderProps.query,
+        loading,
+        error,
+        ...this.builderProps,
       };
 
       slot = $scopedSlots.default ? $scopedSlots.default(slotProps) : slot;
--- a/packages/cubejs-vue/src/QueryRenderer.js
+++ b/packages/cubejs-vue/src/QueryRenderer.js
@@ -46,8 +46,13 @@ export default Vue.component('QueryRenderer', {
   },
   render(createElement) {
     const { $scopedSlots, resultSet, error, loading, sqlQuery } = this;
-    const empty = createElement('div', {});
+    const empty = createElement('div', { class: { 'no-content': true }});
     let slot = this.$slots.empty ? this.$slots.empty : empty;
+    let controls = createElement('div', { class: { 'no-information': true }});
+
+    if ($scopedSlots.builder && this.builderProps.measures) {
+      controls = $scopedSlots.builder({ ...this.builderProps });
+    }
 
     if (!loading && resultSet && !error) {
       const slotProps = {
@@ -68,7 +73,10 @@ export default Vue.component('QueryRenderer', {
           'cubejs-query-renderer': true,
         },
       },
-      slot,
+      [
+        controls,
+        slot,
+      ],
     );
   },
   methods: {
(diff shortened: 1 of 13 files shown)


richipargo:
Still have to document the changes

paveltiunov:
@richipargo Hey Ricardo! This is big one! Thanks for contributing it! Could you please elaborate a little bit what problem does this PR solve?

richipargo:
Hey @paveltiunov, more than a problem the intention of this PR is to give more flexibility on the usage of the `QueryBuilder` and `QueryRenderer` component. Using more `slots` to differentiate between the states of the query be it loaded, error or loading.
Also to try to cover the components a bit better I implemented Jest Unit testing on top of the components.

Still i'm missing a couple of items so I would leave it open still but comments and suggestions are welcome

richipargo:
@paveltiunov It's ready!

paveltiunov:
@richipargo Hey Ricardo! Cool! Are these changes backward compatible with previous version?

richipargo:
Mostly yes

packages/cubejs-vue/src/QueryRenderer.js
```diff
@@ -41,40 +44,47 @@ export default Vue.component('QueryRenderer', {
...
+      slot = $scopedSlots.default ? $scopedSlots.default(slotProps) : slot;
+    } else if (error) {
+      slot = $scopedSlots.error ? $scopedSlots.error({ error, sqlQuery }) : slot;
     }
+
+    return createElement(
+      'div',
+      {
+        class: {
+          'cubejs-query-renderer': true,
```
paveltiunov: Do you use this one for styling?

richipargo:
The idea is to make the object selectable if it requires styling, but of
course I do use it

paveltiunov:
@richipargo Is there any chance we can make it fully backward compatible? I know some guys already deployed it. It would be great if we can make every version upgrade as smooth as possible.

richipargo:
I wanted to avoid it cause it generated a lot of noise, but i've set back into the default slot the missing variables to mantain compatibility

packages/cubejs-vue/src/QueryRenderer.js
```diff
@@ -41,40 +44,50 @@ export default Vue.component('QueryRenderer', {
...
-        this.$scopedSlots.default({
-          loading,
-          error,
-          sqlQuery,
-          ...this.builderProps,
-        }),
-      );
+      slot = $scopedSlots.default ? $scopedSlots.default(slotProps) : slot;
+    } else if (error) {
+      slot = $scopedSlots.error ? $scopedSlots.error({ error, sqlQuery }) : slot;
```
paveltiunov: Do you think we should fallback here to default if only one slot is passed for backward compatibility?

richipargo: You're right i missed that one

paveltiunov:
@richipargo Hey Ricardo! Everything looks great! The only thing: is there any chance it can work without class names? Would be really great if we can maintain code to be class name free in order to prevent any class name clash.

richipargo:
@paveltiunov while i would like to keep it I agree with the possible clashing, so I removed it

paveltiunov:
@richipargo Cool! Thanks!

PR #160, opened 2019-07-11
Oracle driver

itrcz:
1) Oracle Driver
2) Oracle Query
3) Syntax "AS" for table&joins alias in BaseQuery was moved to getter


paveltiunov:
@itrcz Hey Ilya! I think we're good to merge this one as there're only minor issues we were talking about.

PR #161, opened 2019-07-15
fix primaryKey filter for method dimensionColumns

itrcz:


paveltiunov:
@itrcz Hey Ilya! Thanks for this one!

PR #162, opened 2019-07-15
Fix string concat for MySQL

davidlkennedy:
Unfortunately, MySQL uses CONCAT() for string concat instead of the more common '||'.  This fixes that in cube.js.


paveltiunov:
@davidlkennedy Hey David! Thanks for contributing this!

PR #169, opened 2019-07-24
dbtype should be prestodb not presto

rongfengliang:
dbtype for presto  should be `prestodb` not `presto`


paveltiunov:
@rongfengliang Hey Rong! Yep. Thanks for fixing this!

PR #175, opened 2019-07-31
Typo in joins.md

deem0n:
Obviously you can not use number type on TEXT strings.


paveltiunov:
@deem0n Hey Dmitriy! Nice catch! Thanks for fixing this one!

Issue #176, opened 2019-07-31
Postgres: ERROR: type "string" does not exist


shietpas:
**Describe the bug**
I have a pre-aggregate defined as an External Rollup, where Postgres is the database type for pre-aggregations.  When cube.js attempts to create the pre-aggregate table, it defines some fields as data type "String", which is not supported in Postgres.  Similar to following
CREATE TABLE stb_pre_aggregations.my_dimension_and_date_pavk001m_lhrjued2_1564600063903 ("myCol1" **STRING**, "myCol2" STRING, "myMeasure" FLOAT)"

Exception: textPayload:  "{"message":"Error querying db","query":"SELECT xxxx","params":["2019-01-01T00:00:00Z","2019-12-31T23:59:59Z"],"error":"Error: type \"string\" does not exist\n    at QueryQueue.parseResult (/srv/node_modules/@cubejs-backend/query-orchestrator/orchestrator/QueryQueue.js:75:13)\n    at QueryQueue.executeInQueue (/srv/node_modules/@cubejs-backend/query-orchestrator/orchestrator/QueryQueue.js:64:19)\n    at <anonymous>\n    at process._tickDomainCallback (internal/process/next_tick.js:229:7)"}"  
 timestamp:  "2019-07-31T19:08:12.527Z"  

**To Reproduce**
1. Define cube schema using Big Query as primary database type
2. Define pre-aggregate with "external: true" for External Rollup
3. Define external database type to be Postgres
4. Run query that would result in pre-aggregate creation

**Expected behavior**
Table should be created in Postgres under stb_pre_aggregations schema that reflects pre-aggregation table using valid Postgres data types.

**Screenshots**

**Version:**
0.10.16

**Additional context**
Using serverless-google-cloudfunctions

paveltiunov:
@shietpas Thanks for posting this! 

Need to implement this one: https://github.com/cube-js/cube.js/blob/353ed631bd9c227f4d9d77df63b2ebf900a680ed/packages/cubejs-query-orchestrator/driver/BaseDriver.js#L157

shietpas:
@paveltiunov I assume this means overriding the fromGenericType method in the PostgresDriver.js file.  Can I edit that in this GitHub project (fork/PR)?  I'm not familiar with how that ends up in npm under @cubejs-backend/postgres-driver package.

paveltiunov:
@shietpas Yep. You can fork and then provide PR. We'll release it for npm then.

PR #178, opened 2019-08-05
Fix postgres datatype lookup

shietpas:
I incorporated the toLowerCase() on datatype for consistency and added datetime and integer datatypes.  Resolves Issue #176

### DIFF
--- a/packages/cubejs-postgres-driver/driver/PostgresDriver.js
+++ b/packages/cubejs-postgres-driver/driver/PostgresDriver.js
@@ -2,7 +2,9 @@ const BaseDriver = require('@cubejs-backend/query-orchestrator/driver/BaseDriver
 const { Pool } = require('pg');
 
 const GenericTypeToPostgres = {
-  string: 'text'
+  string: 'text',
+  datetime: 'timestamp',
+  integer: 'int'
 };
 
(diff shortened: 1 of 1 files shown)


paveltiunov:
@shietpas Hey Scott! Thanks for providing this one! Could you please elaborate a little bit on `toLowerCase()` case? What's the case you need it?

shietpas:
@paveltiunov When the external rollup (pre-aggregate) table is created, the data types came through as "STRING", "DATETIME", etc.  Since the key lookup in the GenericTypeToPostgres object is case sensitive, it did not find a match on data type "string" and therefore did not convert to "text".  Automatically converting to lowercase will support matching regardless of case.

shietpas:
FYI, the scenario where this occurred is using PostgreSQL as External Rollup DB and Big Query as primary DB.  The Big Query data types are captured as uppercase.

paveltiunov:
@shietpas Gotcha. I believe these types should be fixed at BigQuery side because DBs other than Postgres would be affected as well.

paveltiunov:
@shietpas Do you think you can fix `BigQueryDriver` and `PostgresDriver` accordingly? Thanks!

Issue #200, opened 2019-09-03
Speed up postgres pre aggregations


landyrev:
I’m trying to use Postgres as external db for rollup pre aggregations in serverless mode, but facing following issue: creating of new pre-aggregations takes a lot of time — about one second per every 4-5 rows of pre aggregated data.

As I can see — it happens because cube uploads data to Postgres row-by-row, and it takes a lot of time. Is there any ways to speed up uploading?

paveltiunov:
@landyrev Thanks for posting this! Yep. It's because default driver implementation doesn't have any  DB specific optimizations. Thanks for providing this implementation!

PR #201, opened 2019-09-03
Speedup PG pre aggregations

landyrev:
**Check List**

#200

PostgresDriver not has it's own method uploadTable, which uploads all data using [single query](https://github.com/brianc/node-postgres/issues/957#issuecomment-295583050), which makes uploading process faster.

### DIFF
--- a/packages/cubejs-postgres-driver/driver/PostgresDriver.js
+++ b/packages/cubejs-postgres-driver/driver/PostgresDriver.js
@@ -51,6 +51,26 @@ class PostgresDriver extends BaseDriver {
     }
   }
 
+  async uploadTable(table, columns, tableData) {
+    if (!tableData.rows) {
+      throw new Error(`${this.constructor} driver supports only rows upload`);
+    }
+    await this.createTable(table, columns);
+    try {
+      const q = `INSERT INTO ${table}
+      (${columns.map(c => this.quoteIdentifier(c.name)).join(', ')})
+      SELECT * FROM UNNEST (${columns.map((c, columnIndex) => `${this.param(columnIndex)}::${this.toGenericType(c.type)}[]`).join(', ')})`;
+      console.log(q);
+      await this.query(
+        q,
+        columns.map(c => tableData.rows.map(r => r[c.name]))
+      );
+    } catch (e) {
+      await this.dropTable(table);
+      throw e;
+    }
+  }
+
   release() {
     return this.pool.end();
   }
--- a/packages/cubejs-postgres-driver/driver/PostgresDriver.js
+++ b/packages/cubejs-postgres-driver/driver/PostgresDriver.js
@@ -60,7 +60,7 @@ class PostgresDriver extends BaseDriver {
       await this.query(
         `INSERT INTO ${table}
       (${columns.map(c => this.quoteIdentifier(c.name)).join(', ')})
-      SELECT * FROM UNNEST (${columns.map((c, columnIndex) => `${this.param(columnIndex)}::${this.toGenericType(c.type)}[]`).join(', ')})`,
+      SELECT * FROM UNNEST (${columns.map((c, columnIndex) => `${this.param(columnIndex)}::${c.type}[]`).join(', ')})`,
         columns.map(c => tableData.rows.map(r => r[c.name]))
       );
     } catch (e) {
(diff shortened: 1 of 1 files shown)


packages/cubejs-postgres-driver/driver/PostgresDriver.js
```diff
@@ -51,6 +51,26 @@ class PostgresDriver extends BaseDriver {
...
+  async uploadTable(table, columns, tableData) {
+    if (!tableData.rows) {
+      throw new Error(`${this.constructor} driver supports only rows upload`);
+    }
+    await this.createTable(table, columns);
+    try {
+      const q = `INSERT INTO ${table}
+      (${columns.map(c => this.quoteIdentifier(c.name)).join(', ')})
+      SELECT * FROM UNNEST (${columns.map((c, columnIndex) => `${this.param(columnIndex)}::${this.toGenericType(c.type)}[]`).join(', ')})`;
+      console.log(q);
```
paveltiunov: I believe we can remove it.

packages/cubejs-postgres-driver/driver/PostgresDriver.js
```diff
@@ -51,6 +51,26 @@ class PostgresDriver extends BaseDriver {
...
 
+  async uploadTable(table, columns, tableData) {
+    if (!tableData.rows) {
+      throw new Error(`${this.constructor} driver supports only rows upload`);
+    }
+    await this.createTable(table, columns);
+    try {
+      const q = `INSERT INTO ${table}
+      (${columns.map(c => this.quoteIdentifier(c.name)).join(', ')})
+      SELECT * FROM UNNEST (${columns.map((c, columnIndex) => `${this.param(columnIndex)}::${this.toGenericType(c.type)}[]`).join(', ')})`;
```
paveltiunov: Could you please elaborate on scenario where this cast is required? Also is it `toGenericType` and not `fromGenericType`?

paveltiunov:
@landyrev Hey Alexey! Thanks for providing this! Looks great! There're 2 comments to resolve and we're good to merge.

packages/cubejs-postgres-driver/driver/PostgresDriver.js
landyrev: Oops, you're right

packages/cubejs-postgres-driver/driver/PostgresDriver.js
landyrev: Hmm, `toGenericType` is redundant here, you're right. 
Now I see that types are being casted in `createTableSql` method.

landyrev:
Done with comments, thank you

packages/cubejs-postgres-driver/driver/PostgresDriver.js
```diff
@@ -51,6 +51,24 @@ class PostgresDriver extends BaseDriver {
...
+  async uploadTable(table, columns, tableData) {
+    if (!tableData.rows) {
+      throw new Error(`${this.constructor} driver supports only rows upload`);
+    }
+    await this.createTable(table, columns);
+    try {
+      await this.query(
+        `INSERT INTO ${table}
+      (${columns.map(c => this.quoteIdentifier(c.name)).join(', ')})
+      SELECT * FROM UNNEST (${columns.map((c, columnIndex) => `${this.param(columnIndex)}::${c.type}[]`).join(', ')})`,
```
paveltiunov: @landyrev Cast in SQL is still there. I believe there's some case when `this.param(columnIndex)` doesn't work. Just curious what the case is it? If cast is required we need to use `this.fromGenericType(c.type)` because `c.type` is a generic type.

landyrev: Got it.

When you omit explicit type casting in UNNEST function you get `unnest(unknown) is not unique` error for some reason. Here are some related issues: [one](https://stackoverflow.com/questions/52678372/pq-function-unnestunknown-is-not-unique), [two](https://github.com/brianc/node-postgres/issues/1776)).

landyrev: Added `this.fromGenericType`

paveltiunov: @landyrev Gotcha. Yeah. Was thinking about the same. Thanks!

PR #211, opened 2019-09-18
fix empty array error

YEriin:
It will error in "TypeError: Reduce of empty array with no initial value" if result is empty array.


paveltiunov:
Hey @YEriin ! Nice catch! Thanks for fixing this!

PR #214, opened 2019-09-23
change playground urls to ralative urls

superbarne:
**Check List**

I changed the urls for the playground api in to relative urls. Becaus i mounted the playground as express subapp and its wanted to call the playound from the root level.

### DIFF
--- a/packages/cubejs-playground/src/App.js
+++ b/packages/cubejs-playground/src/App.js
@@ -44,7 +44,7 @@ class App extends Component {
       });
     });
 
-    const res = await fetch('/playground/context');
+    const res = await fetch('playground/context');
     const result = await res.json();
     if (window.analytics) {
       window.analytics.identify(result.anonymousId, {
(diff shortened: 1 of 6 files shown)


paveltiunov:
@superbarne Hey Barne! Thanks for contributing this! Could you please elaborate a little bit more on your use case? Are you embedding Playground somewhere as part of your solution to be able to work with it in production environment?

superbarne:
No not realy for production, its an internal tool. In my case i wanted to embed it in a litte express app, so i could access it directly for exploring the data.
I could also run it seperatly but i thougt it would be nice, if its in the same application.

paveltiunov:
@superbarne That totally makes sense. Change seems good to me. The only thing is using relative paths isn't backward compatible at the moment because we know about deployments where playground served not at the root. I think we can go with base path solution. What do you think?

superbarne:
and the base path configuration is passed via a configuration variable?

paveltiunov:
@superbarne Yep. Right. Expose react app variable and rebuild playground with it: https://create-react-app.dev/docs/adding-custom-environment-variables.

paveltiunov:
@superbarne We also just recently released new dashboard app which has the same Query builder as in Playground. Could you please check it and let us know if you still need to host Playground somewhere?

paveltiunov:
@superbarne We just published https://react-dashboard.cube.dev/ which can solve this problem for you. Closing for now. Please reopen if you still need these changes.

PR #217, opened 2019-09-25
Just fixed a small typo

janrenz:
This is only fixing a small typo in the docs


paveltiunov:
@janrenz Thanks for providing this!

Issue #209, opened 2019-09-13
Support more options for rollingWindow leading/trailing values


bentesha:
At the moment, `rollingWindow`s `leading` and `trailing` parameters supports values in the following format: `(-?\d+) (minute|hour|day|week|month|year)` which can be used to define the window size to the right and left of the `offset`.

While this works OK in many scenarios, there are circumstances in which the correct value for `leading`/`trailing` parameter cannot be determined at schema creation.

For example, let's say you want to calculate the total rolling revenue since the beginning of the last month to today. To achieve this, you will need to know the query's date range (which is only available at query creation) so that you can precisely calculate the number of days to set to the `trailing` parameter.

```
{
  offset: 'start',
  trailing: '? day' // how many days to the start of the previous month?
}
```

To overcome such limitations, additional options can be added to enable specifying beginning/ending of a period, e.g start/end of month, year, week, n months ago etc.

One way would be to add additional options such as:

```
{
  offset: 'start',
  trailing: 'start of 1 month ago'
}
```

Or something similar could be attained by introducing a modifier after a `pipe` symbol, to specify the start/end of a period as follow

```
{
  offset: 'start',
  trailing: '1 month | start of month'
}
```

The period part can optionally be omitted:

```
{
  offset: 'start',
  leading: '| end of year' // calculate running total until the period's year-end
}
```

My suggestions might not be the best way to implement this feature. But I do hope they serve well to explain how this improvement can be implemented.

I am open to your ideas and suggestions.

Thanks

paveltiunov:
@bentesha Hey Benedict! Thanks for posting this! It's a great use case! So I believe it'll be very hard to describe all the possible cases using just interval literals and logic. I'd suggest to go with arbitrary `sql` definitions for such rolling intervals. For example for start of the month it'll be:
```
{
  offset: 'start',
  trailing: {
    sql: `date_trunc('month', ${START_DATE})`
  }
}
```
What do you think?

bentesha:
@paveltiunov This would be a lot more powerful. I have never looked into CubeJs source code. How hard would it be for someone new to CubeJs to accomplish this? Would love to create a PR

paveltiunov:
@bentesha I believe not that hard for this particular change. There's only one place which controls SQL generated for trailing and leading intervals:
[...]

PR #223, opened 2019-10-03
feat(@cubejs-backend/schema-compiler): add context variables `START_DATE` & `END_DATE` to `rollingWindow`

bentesha:
This PR allows users to define arbitrary date ranges using `START_DATE` and `END_DATE` context variables in the `leading` and `trailing` properties of `rollingWindow`.

See issue #209 for more details

Example: To calculate rolling count since the start of the current month of the date range.

```javascript
monthToDateCount: {
  type: 'count',
  sql: 'id',
  rollingWindow: {
    trailing: {
      sql: `DATE_FORMAT(${START_DATE}, '%Y-%m-01')`
    },
    leading: {
      sql: `${END_DATE}`,
      inclusive: true
    }
  }
}
```

By default, the trailing start is inclusive, while the leading end is exclusive. This behavior can be changed by setting the `inclusive` property to `true` or `false`. In the example above, both the trailing start and the leading end are **inclusive**.

### DIFF
--- a/packages/cubejs-schema-compiler/adapter/BaseMeasure.js
+++ b/packages/cubejs-schema-compiler/adapter/BaseMeasure.js
@@ -88,7 +88,7 @@ class BaseMeasure {
     const rollingWindow = this.measureDefinition().rollingWindow;
     if (rollingWindow) {
       return this.query.rollingWindowDateJoinCondition(
-        rollingWindow.trailing, rollingWindow.leading, rollingWindow.offset
+        rollingWindow.trailing, rollingWindow.leading, rollingWindow.offset, this.cube().name
       );
     }
     return null;
@@ -110,7 +110,7 @@ class BaseMeasure {
   }
 
   granularityFromInterval(interval) {
-    if (!interval) {
+    if (!interval || typeof interval === 'object') {
       return undefined;
     }
     if (interval.match(/day/)) {
--- a/packages/cubejs-schema-compiler/adapter/BaseQuery.js
+++ b/packages/cubejs-schema-compiler/adapter/BaseQuery.js
@@ -207,19 +207,40 @@ class BaseQuery {
     );
   }
 
-  rollingWindowDateJoinCondition(trailingInterval, leadingInterval, offset) {
+  rollingWindowDateJoinCondition(trailingInterval, leadingInterval, offset, cubeName) {
     offset = offset || 'end';
     return this.timeDimensions.map(d =>
       [d, (dateFrom, dateTo, dateField, dimensionDateFrom, dimensionDateTo, isFromStartToEnd) => {
         // dateFrom based window
         const conditions = [];
-        if (trailingInterval !== 'unbounded') {
+        if (trailingInterval && trailingInterval.sql) {
+          const fn = () => this.evaluateSql(cubeName, trailingInterval.sql);
+          const context = {
+            startDate: dateFrom,
+            endDate: dateTo
+          }
+          const evaluated = this.evaluateSymbolSqlWithContext(fn, context);
+          // Unless explicitly defined, the trailing start is inclusive by default
+          const sign = trailingInterval.inclusive === undefined || trailingInterval.inclusive ? '>=' : '>';
+          conditions.push(`${dateField} ${sign} ${evaluated}`);
+        } else if (trailingInterval !== 'unbounded') {
           const startDate = isFromStartToEnd || offset === 'start' ? dateFrom : dateTo;
           const trailingStart = trailingInterval ? this.subtractInterval(startDate, trailingInterval) : startDate;
           const sign = offset === 'start' ? '>=' : '>';
           conditions.push(`${dateField} ${sign} ${trailingStart}`);
         }
-        if (leadingInterval !== 'unbounded') {
+
+        if (leadingInterval && leadingInterval.sql) {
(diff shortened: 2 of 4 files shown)


paveltiunov:
@bentesha Hey Benedict! Looks great! Thanks for contributing this! Would love if you can write test for it. 

Could you please also elaborate on your inclusive use case?

Will post some code comments shortly.

packages/cubejs-schema-compiler/adapter/BaseQuery.js
```diff
@@ -1167,6 +1188,9 @@ class BaseQuery {
   }
 
   evaluateFiltersArray(filtersArray, cubeName) {
+    console.log('evaluate sql', {
```
paveltiunov: I believe we can remove it.

packages/cubejs-schema-compiler/adapter/BaseMeasure.js
```diff
@@ -110,7 +110,7 @@ class BaseMeasure {
   }
 
   granularityFromInterval(interval) {
-    if (!interval) {
+    if (!interval || typeof interval === 'object') {
```
paveltiunov: This may lead to discrepancy problems. We either should provide a way to define it or to fallback to ungrouping here: https://github.com/cube-js/cube.js/blob/0744b132b15cd4c285bcc53b0d9f5916297f54c1/packages/cubejs-schema-compiler/adapter/BaseQuery.js#L417

packages/cubejs-schema-compiler/adapter/BaseQuery.js
```diff
@@ -1334,7 +1358,9 @@ class BaseQuery {
       sqlUtils: {
         convertTz: this.convertTz.bind(this)
       }
-    }, R.map(
+    },
+    this.safeEvaluateSymbolContext(),
```
paveltiunov: This one is too broad. Usually it contains just internal of SQL evaluation. It makes sense to use only specific `startDate` and `endDate` properties here.

packages/cubejs-schema-compiler/adapter/BaseQuery.js
bentesha: Definitely yes. That kind of slipped through the cracks

packages/cubejs-schema-compiler/adapter/BaseMeasure.js
bentesha: I will look through this and advise

bentesha:
Hi @paveltiunov 
Would love to write some tests for this PR. I wasn't able to run tests though. Seems like I need to have access to all databases that Cube.js support??

packages/cubejs-schema-compiler/adapter/BaseQuery.js
bentesha: Do you think this will require to write a separate method to handle `startDate` and `endDate`?

paveltiunov: @bentesha I think something like following will be sufficient:
```
return Object.assign({
      filterParams: this.filtersProxy(),
      sqlUtils: {
        convertTz: this.convertTz.bind(this)
      },
      startDate: this.safeEvaluateSymbolContext().startDate,
      endDate: this.safeEvaluateSymbolContext().endDate
    })
```

paveltiunov:
@bentesha You can just rebase your PR, start docker and run tests. Cube.js uses `testcontainers` to run tests so no need to install databases locally.

packages/cubejs-schema-compiler/adapter/BaseQuery.js
bentesha: @paveltiunov why not modify function signature and add a parameter for passing extra context variables? Having to modify function implementation to support new context variables does not seem sustainable in the long term. The function should be flexible enough to accommodate new context variables without modifying its implementation.

```javascript
return Object.assign({
      filterParams: this.filtersProxy(),
      sqlUtils: {
        convertTz: this.convertTz.bind(this)
      }
    }, extraContext )
```

bentesha:
> @bentesha You can just rebase your PR, start docker and run tests. Cube.js uses `testcontainers` to run tests so no need to install databases locally.

@paveltiunov Thanks. Will try this approach

packages/cubejs-schema-compiler/adapter/BaseQuery.js
paveltiunov: @bentesha Yep. Sounds good. We can go this route.

bentesha: @paveltiunov Will make changes and update later today

paveltiunov:
@bentesha Hey Benedict! Just wanted to check how things are going with this PR? Do you need any help with it?

bentesha:
Hello @paveltiunov 

I will resume this weekend. I had to work on something and didn't get the time to finalize the PR

paveltiunov:
@bentesha Sounds good! 🙌

hassankhan:
Hi @bentesha, just wanted to check-in with you about the status of the PR 😄

rpaik:
@bentesha I'll go ahead and close this PR for now. If you're ready to pick this up again, please feel free to reopen. Thanks...

tomsseisums:
@paveltiunov @rpaik I'm looking to pick this up. Any tips? Been two years since last update to this PR, probably a lot has changed in Cube's source by then.

PR #224, opened 2019-10-03
Fix continueWaitTimout is ignored

vieira:
**Check List**

I am not familiar with the internals of cube, however I am experiencing a problem where `continueWaitTimeout` is not being honored and upon inspecting the code it seems that the option is only used as a fallback to an argument that is never undefined.

Furthermore it seems thar a pair of parentheses is missing (?) otherwise it will wait for 5ms instead of 5s.

I might be missing something as I am really not familiar with the code, please review.

Thank you.

### DIFF
--- a/packages/cubejs-query-orchestrator/orchestrator/LocalQueueDriver.js
+++ b/packages/cubejs-query-orchestrator/orchestrator/LocalQueueDriver.js
@@ -34,7 +34,7 @@ class LocalQueueDriverConnection {
 
     const res = await Promise.race([
       this.getResultPromise(resultListKey),
-      timeoutPromise(continueWaitTimeout || this.continueWaitTimeout * 1000),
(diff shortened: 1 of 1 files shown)


paveltiunov:
@vieira Hey João! Thanks for contributing this! Nice catch! I believe `continueWaitTimeout` can be completely removed as argument. Seems It's just some code refactoring glitch.

vieira:
Thanks @paveltiunov, makes sense. Just pushed another commit to remove the argument.

paveltiunov:
@vieira Nice! Could you please rebase your branch to fix tests so we're good to merge? Thanks!

vieira:
@paveltiunov Sure, done!

paveltiunov:
@vieira Cool! Thanks!

PR #229, opened 2019-10-10
Feature: Add order, renewQuery, and reactivity to Vue component

TCBroad:
**Check List**
- [x] Docs have been added/updated if required

None, just an enhancement and brings the vue component more inline with the plain js version.

- Updated vue QueryBuilder component to enable reactivity on the query prop
- Added ordering support to QB
- Added renewQuery support to QB
- Fixed some spelling/grammar problems in the vue docs.

### DIFF
--- a/docs/Cube.js-Frontend/@cubejs-client-vue.md
+++ b/docs/Cube.js-Frontend/@cubejs-client-vue.md
@@ -15,7 +15,7 @@ into Vue.js app.
 
 ### Props
 
-- `query`: analytic query. [Learn more about it's format](query-format).
+- `query`: query parameters ([learn more about its format](query-format)).
 - `cubejsApi`: `CubejsApi` instance to use.
 
 ### Slots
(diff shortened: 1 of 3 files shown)


paveltiunov:
@TCBroad Hey Tom! Thanks for contributing this! This one like seems like good to merge. I'm curious how do you use it? I was thinking about what if we can introduce some sort of generic state handling for properties not managed by `QueryBuilder`. This way each time new property is added to query format it's already supported by `QueryBuilder` the same way as react package does. @TCBroad @richipargo would love to hear your thoughts here.

TCBroad:
Hi @paveltiunov, thanks for merging! 
We use the reactivity portion to update QueryBuilders in child components where the filter is controlled in the parent. It was difficult to hook into updateFilters without adding a load of watches and boilerplate in each component.
I haven't looked at the react package much but I think it would be a good idea to have some way to automatically pick up the properties. Could the same code be ported over to vue?

richipargo:
Hey @paveltiunov and @TCBroad,

While I did base off some of my initial component work from the React implementation I think the underlying reactivity systems differ so we cannot do the same, but also features are not the same yet, so it still requires a lot of work and I've been glad @TCBroad has improved on top of my initial work.

Issue #12, opened 2019-02-14
QueryRenderer: fix prop types warnings


keydunov:

paveltiunov:
Already fixed in 65e1380830f39a748a706debccceb1726b8bfe6d

PR #230, opened 2019-10-10
Read schema subfolders

lksilva:
**Check List**

[For example #12]

[Description goes here]

### DIFF
--- a/packages/cubejs-server-core/core/FileRepository.js
+++ b/packages/cubejs-server-core/core/FileRepository.js
@@ -12,9 +12,9 @@ class FileRepository {
   }
 
   async getFiles(dir, fileList = []) {
-    const files = await fs.readdir(dir);
+    const files = await fs.readdir(path.join(this.localPath(), dir));
     for (const file of files) {
-      const stat = await fs.stat(path.join(dir, file));
+      const stat = await fs.stat(path.join(this.localPath(), dir, file));
       if (stat.isDirectory())
         fileList = await this.getFiles(path.join(dir, file), fileList);
       else fileList.push(path.join(dir, file));
@@ -24,7 +24,7 @@ class FileRepository {
 
   async dataSchemaFiles(includeDependencies) {
     const self = this;
-    const files = await self.getFiles(self.localPath());
+    const files = await self.getFiles("");
     let result = await Promise.all(
       files
         .filter(file => R.endsWith(".js", file))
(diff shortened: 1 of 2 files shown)


paveltiunov:
@lksilva Hey Lucas! Thanks for contributing this! Looks great! The only thing: `fileName` is used to be local relative to `schemaPath`. We should save this behavior as there's some stuff that depends on it.

lksilva:
Do you have any suggestion to resolve this issue? That I would be very happy in implement.

I apply this suggestion because we need this solution here at work

packages/cubejs-server-core/core/FileRepository.js
```diff
@@ -11,13 +11,25 @@ class FileRepository {
     return path.join(process.cwd(), this.repositoryPath);
   }
 
+  async getFiles(dir, fileList = []) {
+    const files = await fs.readdir(dir);
```
paveltiunov: Here should be `path.join(this.localPath(), dir)` instead of `dir`

packages/cubejs-server-core/core/FileRepository.js
```diff
@@ -11,13 +11,25 @@ class FileRepository {
...
+    for (const file of files) {
+      const stat = await fs.stat(path.join(dir, file));
```
paveltiunov: Here should be `path.join(this.localPath(), dir, file)` instead of `path.join(dir, file)`.

packages/cubejs-server-core/core/FileRepository.js
```diff
@@ -11,13 +11,25 @@ class FileRepository {
...
+        fileList = await this.getFiles(path.join(dir, file), fileList);
+      else fileList.push(path.join(dir, file));
+    }
+    return fileList;
+  }
+
   async dataSchemaFiles(includeDependencies) {
     const self = this;
-    const files = await fs.readdir(this.localPath());
+    const files = await self.getFiles(self.localPath());
```
paveltiunov: Here should be `''` (empty string) instead of `self.localPath()`

lksilva:
I made the changes that you requested and I'm thinking about do test to cover FileRepository

paveltiunov:
@lksilva Looks great! Could you please rebase your branch and resolve conflicts and fix linter errors? After that we're good to merge. 

Test would be great! You can add it in this PR or create separate.

lksilva:
I prefer apply in another PR because I will take a little time for this, can it be like this?

paveltiunov:
@lksilva Yep. Sure! Sounds good!

PR #236, opened 2019-10-22
feat: dynamic case label

ferrants:
**Check List**

https://github.com/cube-js/cube.js/issues/235

label is run through the execute sql function

### DIFF
--- a/docs/Schema/dimensions.md
+++ b/docs/Schema/dimensions.md
@@ -100,6 +100,23 @@ size: {
 }
 ```
 
+`label` can be statically defined as a string or use a property from the cube as an object with a `sql` property.
+
+```javascript
+size: {
+  type: `string`,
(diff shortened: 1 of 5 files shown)


paveltiunov:
@ferrants Thanks for contributing this! So the implementation can be just to check if label is an object with `sql` property then we can evaluate it otherwise save old behavior. Need to add appropriate checks in `CubeValidator` as well. Test for this use case can be very helpful here.

ferrants:
> Linter has been run for changed code
I could add linting to this package, but it'll make the PR much bigger.

paveltiunov:
@ferrants schema compiler isn't ready for whole package linting so let's keep it as is. 

Overall PR looks great! Just remove this orphaned `res = ` and we're good to merge! Thanks again!

PR #240, opened 2019-10-25
improve: reduce repeat

Justin-ZS:


paveltiunov:
@Justin-ZS Hey Jin! This looks much cleaner! Thanks for contributing this!

PR #244, opened 2019-10-28
pretty default logger

Justin-ZS:
Before:
Screen Shot 2019-10-28 at 11 41 52 AM
After:
Screen Shot 2019-10-28 at 11 39 44 AM

### DIFF
--- a/packages/cubejs-server-core/core/index.js
+++ b/packages/cubejs-server-core/core/index.js
@@ -41,6 +41,29 @@ const checkEnvForPlaceholders = () => {
   }
 };
 
+const devLogger = (type, message, error) => {
+  const colors = {
+    red: '31', // ERROR
+    green: '32', // INFO
+    blue: '34', // TRACE
+  }
+  const withColor = (str, color = colors.green) => `\u001b[${color}m${str}\u001b[0m`;
+  const format = json => JSON.stringify(json, null, 2);
+
+  if ([ // INFO list
+    'Performing query',
+    'Performing query completed',
+  ].includes(type)) {
+    if (error) console.log(`${withColor(type, colors.blue)}: ${error}`);
+    else console.log(`${withColor(type, colors.blue)}: ${format(message)}`);
+    return;
+  }
+  // ERROR
+  if (error) return console.log(`${withColor(type, colors.red)}: ${error}`);
+  // TRACE
+  console.log(`${withColor(type, colors.green)}: ${format(message)}`);
+}
+
 class CubejsServerCore {
   constructor(options) {
     options = options || {};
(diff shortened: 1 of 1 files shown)


paveltiunov:
@Justin-ZS Hey Jin! Thanks for contributing this! Logging is definitely a place that requires some improvements. There're log levels that have been requested some time ago. Would you like to implement those? It would help not to overwhelm console output. Something like `INFO` (default), `TRACE` and `ERROR`. 
- `ERROR`: Output event data only for errors
- `INFO`: Output event data only for `Performing query`, `Performing query completed` and errors
- `TRACE`: Output event data for all events
By event data I mean `JSON.stringify(restParams, null, 2)`

What do you think?

Justin-ZS:
@paveltiunov Fine, I will try to implement it after work.

packages/cubejs-server-core/core/index.js
```diff
@@ -41,6 +41,29 @@ const checkEnvForPlaceholders = () => {
...
+const devLogger = (type, message, error) => {
+  const colors = {
+    red: '31', // ERROR
+    green: '32', // INFO
+    blue: '34', // TRACE
+  }
+  const withColor = (str, color = colors.green) => `\u001b[${color}m${str}\u001b[0m`;
+  const format = json => JSON.stringify(json, null, 2);
+
+  if ([ // INFO list
```
paveltiunov: Can we do it configurable? Using some env like `CUBEJS_LOG_LEVEL=TRACE`?

packages/cubejs-server-core/core/index.js
```diff
@@ -41,6 +41,29 @@ const checkEnvForPlaceholders = () => {
...
+    'Performing query',
+    'Performing query completed',
+  ].includes(type)) {
+    if (error) console.log(`${withColor(type, colors.blue)}: ${error}`);
```
paveltiunov: Would love if error still will show other log event props. They can contain additional info.

packages/cubejs-server-core/core/index.js
```diff
@@ -41,6 +41,29 @@ const checkEnvForPlaceholders = () => {
...
+    else console.log(`${withColor(type, colors.blue)}: ${format(message)}`);
+    return;
+  }
```
paveltiunov: Also let's output all other messages for `INFO` but just without log event props. This way we can keep console clean and still readable.

packages/cubejs-server-core/core/index.js
Justin-ZS: It looks like I misunderstand your words, log level should depend on env config rather than event itself.

packages/cubejs-server-core/core/index.js
Justin-ZS: OK, I get it.

packages/cubejs-server-core/core/index.js
paveltiunov: Yep. Right.

paveltiunov:
@Justin-ZS Hey Jin! Looks good! Thanks again for contributing this!

PR #245, opened 2019-10-28
refactor(@cubejs-backend/schema-compiler): improve readability

Justin-ZS:
Hard to read without documents, so I refactor the big function into pieces.

### DIFF
--- a/packages/cubejs-schema-compiler/adapter/PreAggregations.js
+++ b/packages/cubejs-schema-compiler/adapter/PreAggregations.js
@@ -195,57 +195,73 @@ class PreAggregations {
   }
 
   static canUsePreAggregationForTransformedQueryFn(transformedQuery, refs) {
+    // match helpers
+    // ([a], [b]) -> Bool
+    const contains = (arr, subArr) => R.all(item => arr.includes(item), subArr);
+    // match time dimensions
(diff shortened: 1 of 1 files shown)


Justin-ZS:
@paveltiunov 
And I think we can merge the `isAdditive` branch to `leafMeasureAdditive` branch.
In `isAdditive` branch, `transformedQuery.measures` always equals `transformedQuery.leafMeasures`.

paveltiunov:
@Justin-ZS Hey Jin! Thanks for refactoring this! Overall agree on that idea! However we'd need to test performance of this method very thoroughly. The reason for it there's high call throughput during `autoRollup` calculations. Like millions of calls. Current implementation is a base line: we should make it faster and not slower. So we want to be it both faster and more readable as a result of this change.

I'll write some performance tests which corresponds to real world load for this method and will get back to you on this.

Justin-ZS:
@paveltiunov  Can't agree more for the importance of performance!
As a fan of Ramda too, I find if you write some code and expect it to obtain a very high performance, the first choice is not to use Ramda at all.
Immutability and functional style benefits to composing high-level logic, but poor at performance due to unoptimized JS engine.

paveltiunov:
Closing as this code is out of date. Please feel free to open other PR to revisit the problem using a new codebase.

PR #247, opened 2019-10-31
refactor: compiler should not rely on query instance

Justin-ZS:

Based on the compiled schema, Query builder can convert the incoming queries to SQL.
Creating a query builder instance during compile schema seems strange and out of reason.
I add a static `aliasName` method to avoid creating instances for `CubeToMetaTransformer`.

### DIFF
--- a/packages/cubejs-schema-compiler/adapter/BaseQuery.js
+++ b/packages/cubejs-schema-compiler/adapter/BaseQuery.js
@@ -1239,6 +1239,10 @@ class BaseQuery {
   }
 
   aliasName(name) {
+    return BaseQuery.aliasName(name);
+  }
+
+  static aliasName(name) {
     return inflection.underscore(name).replace(/\./g, '__');
   }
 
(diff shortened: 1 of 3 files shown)


paveltiunov:
@Justin-ZS Hey Jin! This is interesting one. With recent changes we removed override of this method for BigQuery. This change was made to ensure external rollup compatibility. However there's significant chance we can return adapter specific aliases. So we can make it static but it should still depend on adapter class with an ability to override this method.

Justin-ZS:
@paveltiunov 
OK, I will close this PR.
I still think the `schema compiler` should keep some distance from `query builder`.
Create a query instance is a simple way but brings unexpected cyclic dependence.
For adapter specified aliases, maybe we can separate `CubeToMetaTransformer` from `schema compiler`

paveltiunov:
@Justin-ZS Yeah. Totally agree on that one. It's basically the hack to solve this alias variance problem. We still can get rid of instantiating actual query and using query class instead. There will be still dependence on actual DB adapter somewhere in compiler as we need to provide alias somehow though.

PR #248, opened 2019-11-03
feat: solve empty order query format

richipargo:
**Check List**

No referred issue

When order is empty `QueryBuilder` object passes it to the `QueryRenderer` as an empty object but `@cubejs-server/core` does not allow it so we get a error

### DIFF
--- a/packages/cubejs-vue/src/QueryBuilder.js
+++ b/packages/cubejs-vue/src/QueryBuilder.js
@@ -30,7 +30,7 @@ export default {
       limit: null,
       offset: null,
       renewQuery: false,
-      order: {}
+      order: null,
     };
 
     data.granularities = [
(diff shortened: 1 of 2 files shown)


ifokeev:
yeah, nice PR. I hacked `order` in my project, `defaultOrder` should be controlled behaviour

paveltiunov:
@richipargo Hey Ricardo! Looks good! Could you please rebase your branch so we can rid of these CI error messages? Thanks!

richipargo:
Ready @paveltiunov

ifokeev:
I noticed that this implementation only fix `vue` package, but really I meant https://github.com/cube-js/cube.js/blob/master/packages/cubejs-schema-compiler/adapter/BaseQuery.js#L77
=\

keydunov:
Hey @ifokeev, I agree that there are cases where you want to disable (or maybe have a different default values) of the order. Do you mind to create a separate issue with the detailed description of your current problem so we can kick off the process towards the desired changes.

PR #255, opened 2019-11-07
fix(server-core/DevServer): the schemaPath option does not work when …

ZhiXiao-Lin:
The schemaPath option does not work when generateing schema

**Check List**

[For example #12]

[Description goes here]


paveltiunov:
@ZhiXiao-Lin Hey Roy! Nice catch! Thanks for providing this!

PR #258, opened 2019-11-10
Feature/pre aggregation on time hierachy

Justin-ZS:
**Check List**
* [x] How to run the test locally?

[Support time dimension hieararchies](https://github.com/cube-js/cube.js/issues/246)

Screen Shot 2019-11-10 at 17 09 34

Pre-aggregation on `day` can be used for the query on `month` now

### DIFF
--- a/packages/cubejs-schema-compiler/adapter/BaseTimeDimension.js
+++ b/packages/cubejs-schema-compiler/adapter/BaseTimeDimension.js
@@ -73,10 +73,7 @@ class BaseTimeDimension extends BaseFilter {
       if (context.rollupGranularity === this.granularity) {
         return super.dimensionSql();
       }
-      return this.query.timeGroupedColumn(
-        granularity,
-        this.query.escapeColumnName(this.unescapedAliasName(context.rollupGranularity))
-      );
+      return this.query.timeGroupedColumn(granularity, this.query.dimensionSql(this));
     }
     if (context.ungrouped) {
       return this.convertedToTz();
--- a/packages/cubejs-schema-compiler/adapter/BaseTimeDimension.js
+++ b/packages/cubejs-schema-compiler/adapter/BaseTimeDimension.js
@@ -48,8 +48,10 @@ class BaseTimeDimension extends BaseFilter {
     return super.aliasName();
   }
 
-  unescapedAliasName() {
-    return `${this.query.aliasName(this.dimension)}_${this.granularity || 'date'}`; // TODO date here for rollups
+  unescapedAliasName(granularity) {
+    const actualGranularity = granularity || this.granularity || 'date';
+
+    return `${this.query.aliasName(this.dimension)}_${actualGranularity}`; // TODO date here for rollups
   }
 
   dateSeriesAliasName() {
@@ -64,16 +66,22 @@ class BaseTimeDimension extends BaseFilter {
   }
 
   dimensionSql() {
-    if (this.query.safeEvaluateSymbolContext().rollupQuery) {
-      return super.dimensionSql();
+    const context = this.query.safeEvaluateSymbolContext();
+    const granularity = context.granularityOverride || this.granularity;
+
+    if (context.rollupQuery) {
+      if (context.rollupGranularity === this.granularity) {
+        return super.dimensionSql();
+      }
+      return this.query.timeGroupedColumn(
+        granularity,
+        this.query.escapeColumnName(this.unescapedAliasName(context.rollupGranularity))
+      );
     }
-    if (this.query.safeEvaluateSymbolContext().ungrouped) {
+    if (context.ungrouped) {
       return this.convertedToTz();
     }
-    return this.query.timeGroupedColumn(
-      this.query.safeEvaluateSymbolContext().granularityOverride || this.granularity,
-      this.convertedToTz()
-    );
+    return this.query.timeGroupedColumn(granularity, this.convertedToTz());
   }
 
   convertedToTz() {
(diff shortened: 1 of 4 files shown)


packages/cubejs-schema-compiler/adapter/BaseTimeDimension.js
```diff
@@ -48,8 +48,10 @@ class BaseTimeDimension extends BaseFilter {
     return super.aliasName();
   }
 
-  unescapedAliasName() {
-    return `${this.query.aliasName(this.dimension)}_${this.granularity || 'date'}`; // TODO date here for rollups
+  unescapedAliasName(context = {}) {
```
paveltiunov: Context is a thread local so it's safe to access it right from query everywhere during query generation. It's thread local solely for the purpose of not polluting arguments of methods. So we shouldn't add context to arguments everywhere in the code. Otherwise every method should have context argument.

packages/cubejs-schema-compiler/adapter/PreAggregations.js
```diff
@@ -201,6 +201,25 @@ class PreAggregations {
         timeDimensions.map(d => [d.dimension, d.granularity || 'date'])
       ) || [];
     }
+    // [TimeDimension] -> [TimeDimension]
+    function spawnTimeDimensions(timeDimensions = []) {
```
paveltiunov: This is cool approach! It would be even better if we can pre-generate possible time dimensions based on `transformedQuery` rather than `references`. This way it can be saved once for query and used for all pre-aggregations which is very important from performance perspective.

packages/cubejs-schema-compiler/adapter/BaseTimeDimension.js
Justin-ZS: it's truly a bad practice to use `context` from the arguments.

My purpose is to distinguish `dimensionSQL()` from `selectColumns()` and  return the corresponding SQL.
If we access the `context` by `safeEvaluateSymbolContext()` rather than `arguments`, the SQL will become `date_trunc('month', ("orders.created_at_date)") "orders.created_at_date"`.

Maybe you have some elegant solutions for it?

packages/cubejs-schema-compiler/adapter/PreAggregations.js
Justin-ZS: Sharp eye! I will update accordingly.

Justin-ZS:
@paveltiunov I will try `rebase`. (I have never used before).
Could you give me some tips about how to run the test locally?
So that I can check the failed tests locally and add more tests for my commits.

paveltiunov:
@Justin-ZS Hey Jin! To run tests please start docker and just run `npm test` in schema compiler directory: https://github.com/cube-js/cube.js/blob/master/CONTRIBUTING.md#testing-schema-compiler

packages/cubejs-schema-compiler/adapter/BaseTimeDimension.js
paveltiunov: @Justin-ZS Ah. Gotcha. That's right. We'd need to go another route then. I believe we can apply following change:
```
    if (context.rollupQuery) {
      if (context.rollupGranularity === this.granularity) {
        return super.dimensionSql();
      } else {
        return this.query.timeGroupedColumn(
          this.query.safeEvaluateSymbolContext().granularityOverride || this.granularity,
          this.query.escapeColumnName(this.unescapedAliasName(context.rollupGranularity))
        );
      }
    }
```
And add explicit `rollupGranularity` argument to `unescapedAliasName()` rather than `context`.

packages/cubejs-schema-compiler/adapter/PreAggregations.js
Justin-ZS: @paveltiunov 
Before actual implementation, I considered it again and found it not an easy change.
If every `timeDimension` in query can also be found in a pre-aggregation, we can say they are matched!
So expanding the pre-aggregation time dimension to match more queries is a natural approach.
If we keep `references` unchanged and turn around to update `transformedQuery `, the logic will become obscure and hard to read.
https://github.com/cube-js/cube.js/blob/85208b755c1385f26c9855dd3c8f6e10b02ed588/packages/cubejs-schema-compiler/adapter/PreAggregations.js#L242

paveltiunov: @Justin-ZS Yeah. Agree. But in case of this particular method we need to prioritize performance over readability unfortunately.

packages/cubejs-schema-compiler/adapter/BaseTimeDimension.js
paveltiunov: @Justin-ZS Do you need any additional help on this one? Seems this one is only thing we need to fix and we're good to go.

Justin-ZS: @paveltiunov 
Thank you! I will ask for some help if I stuck somewhere.
I just have not to start to fix this yet.
This weekend, I will take some time to update it and complete the pull request.

paveltiunov:
@Justin-ZS Hey Jin! Everything looks great! The only thing pending: we need to fix this failing test. As you added more broad match rules for granularity matching another pre-aggregation takes precedence. We need just to swap definitions of `multiStage` and `partitioned` in `PreAggregationsTest.js` file. I think it should fix it.

Justin-ZS:
@paveltiunov 
Because the`ratio` and `googleRollup` can also be matched to `multi-Stage` query, I move down `ratio` to the end and change the `granularity` of `googleRollup`from `day` to `week`.

paveltiunov:
@Justin-ZS Hey Jin! Looks good to me! Thanks a lot for contributing this!

Issue #262, opened 2019-11-13
Dynamic Measures using FILTER_PARAMS ERROR when calculated from Pre-Aggregations


ferrants:
**Describe the bug**
Non-Additive measures that can be derived from a roll-up Additive or Leaf Additive measures should be derivable from a Pre-aggregate table query. `measures` that have `filters` that can be determined by the Pre-aggregate's Dimensions should be applied at query-time, not rolled into the preaggregate. Measure `filters` using `FILTER_PARAMS` that are in the Pre-aggregate's dimensions should use the Pre-aggregate table to calculate them.

**To Reproduce**

With this Cube Definition:
```javascript
cube(`SearchResults`, {

    sql: `SELECT * FROM search_result`,
    measures: {
        count: {
            sql: `${CUBE}.raw_count`,
            type: `sum`,
        },
        count_first_half: {
            sql: `${CUBE.count}`,
            type: `number`,
            filters: [
                {
                    sql: `${CUBE.search_time} < ${FILTER_PARAMS.SearchResults.search_time.filter((from, to) => {
                        return `${from}::timestamp + (${to}::timestamp - ${from}::timestamp) / 2`
                    })}`,
                },
            ],
        },
        count_second_half: {
            sql: `${CUBE.count}`,
            type: `number`,
            filters: [
                {
                    sql: `${CUBE.search_time} >= ${FILTER_PARAMS.SearchResults.search_time.filter((from, to) => {
                        return `${from}::timestamp + (${to}::timestamp - ${from}::timestamp) / 2`
                    })}`,
                },
            ],
        },
        count_change: {
            sql: `${CUBE.count_second_half} - ${CUBE.count_first_half}`,
            type: `number`,
        },
        count_pct_change: {
            sql: `
                (${CUBE.count_second_half} - ${CUBE.count_first_half}) * 1.0
                / ${CUBE.count_first_half}
            `,
            type: `number`,
        },
    },

    dimensions: {
        id: {
            sql: `id`,
            type: `string`,
            primaryKey: true,
        },

        asin: {
            sql: `asin`,
            type: `string`,
        },

        search_time: {
            sql: `search_time`,
            type: `time`,
        },

        keyword_id: {
            sql: `keyword_id`,
            type: `string`,
        },
    },

    preAggregations: {
        my_preagg: {
            type: `rollup`,
            measureReferences: [CUBE.count],
            dimensionReferences: [CUBE.keyword_id, CUBE.asin],
[...]

PR #273, opened 2019-11-21
fix for error when querying preaggregates with filters

ferrants:
**Check List**

#262 

Handling querying from preaggregates when using a filter and avoiding the nested aggregate functions

### DIFF
--- a/docs/Schema/cube.md
+++ b/docs/Schema/cube.md
@@ -121,7 +121,7 @@ cube(`OrderFacts`, {
 cube(`ExtendedOrderFacts`, {
   extends: OrderFacts,
 
-  measure: {
+  measures: {
     doubleCount: {
       type: `number`,
       sql: `${count} * 2`
--- a/docs/Schema/pre-aggregations.md
+++ b/docs/Schema/pre-aggregations.md
@@ -8,7 +8,7 @@ menuOrder: 8
 ---
 
 Pre-aggregations are materialized query results persisted as tables.
-In order to start using pre-aggregations Cube.js should have write access to `stb_pre_aggregations` schema, or whatever your [preAggregationsSchema]((@cubejs-backend-server-core#options-reference-pre-aggregations-schema)) setting is set to, where pre-aggregation tables will be stored.
+In order to start using pre-aggregations Cube.js should have write access to `stb_pre_aggregations` schema, or whatever your [preAggregationsSchema](@cubejs-backend-server-core#options-reference-pre-aggregations-schema) setting is set to, where pre-aggregation tables will be stored.
 Cube.js has an ability to analyze queries against defined set of pre-aggregation rules in order to choose optimal one that will be used to create pre-aggregation table.
 
 If Cube.js finds suitable pre-aggregation rule database querying becomes multi-stage.
--- a/packages/cubejs-schema-compiler/adapter/BaseQuery.js
+++ b/packages/cubejs-schema-compiler/adapter/BaseQuery.js
@@ -928,6 +928,14 @@ class BaseQuery {
   }
 
   evaluateSymbolSql(cubeName, name, symbol) {
+    const measurePath = `${cubeName}.${name}`;
+    if (this.safeEvaluateSymbolContext().usedInAggregationWithFilter){
+        return this.escapeColumnName(this.aliasCubeField(cubeName, name))
+    }
+    if (((this.evaluateSymbolContext || {}).renderedReference || {})[measurePath]) {
+      return this.evaluateSymbolContext.renderedReference[measurePath];
+    }
+
     this.pushCubeNameForCollectionIfNecessary(cubeName);
     if (this.cubeEvaluator.isMeasure([cubeName, name])) {
       let parentMeasure;
@@ -952,10 +960,6 @@ class BaseQuery {
           this.safeEvaluateSymbolContext().leafMeasures[this.safeEvaluateSymbolContext().currentMeasure] = true;
         }
       }
-      if (this.safeEvaluateSymbolContext().usedInAggregationWithFilter){
-        return this.escapeColumnName(this.aliasCubeField(cubeName, name))
-      }
-
       let evaluatedSql;
       if (this.safeEvaluateSymbolContext().rollupQuery && symbol.type === 'sum' && symbol.filters) {
         evaluatedSql = this.evaluateSymbolSqlWithContext(
@@ -1133,9 +1137,6 @@ class BaseQuery {
     }
     if (this.safeEvaluateSymbolContext().foundCompositeCubeMeasures && !parentMeasure) {
       this.safeEvaluateSymbolContext().rootMeasure = { multiplied: resultMultiplied, measure: measurePath };
-    }
-    if (((this.evaluateSymbolContext || {}).renderedReference || {})[measurePath]) {
-      return this.evaluateSymbolContext.renderedReference[measurePath];
     }
     if (
       this.safeEvaluateSymbolContext().ungrouped ||
--- a/packages/cubejs-schema-compiler/adapter/BaseQuery.js
+++ b/packages/cubejs-schema-compiler/adapter/BaseQuery.js
@@ -954,11 +954,7 @@ class BaseQuery {
       }
       let evaluatedSql;
       if (this.safeEvaluateSymbolContext().usedInAggregationWithFilter){
-        // TODO - check if this is a measure in the cube preaggregation? elsewhere?
-        // TODO - function for commonly setting this
-        // TODO - confirm context
-        // TODO - don't return here, return later
-        return this.escapeColumnName(`${this.aliasName(cubeName)}__${name}`)
+        return this.escapeColumnName(this.aliasCubeField(cubeName, name))
       } else if (this.safeEvaluateSymbolContext().rollupQuery && symbol.type === 'sum' && symbol.filters) {
 
         evaluatedSql = this.evaluateSymbolSqlWithContext(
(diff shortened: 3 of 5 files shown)


packages/cubejs-schema-compiler/adapter/BaseQuery.js
```diff
@@ -952,17 +952,35 @@ class BaseQuery {
...
+        // TODO - check if this is a measure in the cube
+        // TODO - function for commonly setting this
+        // TODO - confirm context
+        // TODO - don't return here, return later
+        return this.escapeColumnName(`${cubeName}__${name}`)
+      } else if (this.safeEvaluateSymbolContext().rollupQuery && symbol.type === 'sum' && symbol.filters) {
+
+        auto_prefix = this.evaluateSymbolSqlWithContext(
+          () => this.evaluateSql(cubeName, symbol.sql),
+          { usedInAggregationWithFilter: true }
```
ferrants: added this new context

packages/cubejs-schema-compiler/adapter/BaseQuery.js
```diff
@@ -952,17 +952,35 @@ class BaseQuery {
           this.safeEvaluateSymbolContext().leafMeasures[this.safeEvaluateSymbolContext().currentMeasure] = true;
         }
       }
+      let auto_prefix;
+      if (this.safeEvaluateSymbolContext().usedInAggregationWithFilter){
+        // TODO - check if this is a measure in the cube
+        // TODO - function for commonly setting this
+        // TODO - confirm context
+        // TODO - don't return here, return later
+        return this.escapeColumnName(`${cubeName}__${name}`)
```
ferrants: There's got to be an existing function for this. Unable to determine how. Should I make a function for this?

packages/cubejs-schema-compiler/adapter/BaseQuery.js
```diff
@@ -952,17 +952,35 @@ class BaseQuery {
...
+      } else if (this.safeEvaluateSymbolContext().rollupQuery && symbol.type === 'sum' && symbol.filters) {
+
+        auto_prefix = this.evaluateSymbolSqlWithContext(
```
ferrants: need this function here to set the context. not sure if I can consolidate this and the final else

packages/cubejs-schema-compiler/adapter/BaseQuery.js
ferrants: probably shouldn't return here

packages/cubejs-schema-compiler/adapter/BaseQuery.js
```diff
@@ -990,6 +1008,8 @@ class BaseQuery {
           "','",
           this.autoPrefixAndEvaluateSql(cubeName, symbol.longitude.sql)
         ])
+      } else if (this.safeEvaluateSymbolContext().rollupQuery) {
+        return this.escapeColumnName(`${cubeName}__${this.aliasName(name)}`)
```
ferrants: this fixes the dimension in the filter

packages/cubejs-schema-compiler/adapter/BaseQuery.js
```diff
@@ -1034,7 +1054,7 @@ class BaseQuery {
     }, {
       sqlResolveFn: (symbol, cube, n) => self.evaluateSymbolSql(cube, n, symbol),
       cubeAliasFn: self.cubeAlias.bind(self),
-      contextSymbols: this.parametrizedContextSymbols(),
+      contextSymbols: this.parameterizedContextSymbols(),
```
ferrants: fixed spelling

packages/cubejs-schema-compiler/test/PreAggregationsTest.js
```diff
@@ -300,6 +317,67 @@ describe('PreAggregations', function test() {
...
+        measures: [
+          'visitors.count',
+          'visitors.google_count',
+        ],
+        timeDimensions: [{
+          dimension: 'visitors.createdAt',
+          granularity: 'date',
+          dateRange: ['2017-01-01', '2017-01-30']
+        }],
+        dimensions: ['visitors.source'],
```
ferrants: add test case that doesn't use this dimension. should still use this preaggregation

docs/Schema/pre-aggregations.md
```diff
@@ -8,7 +8,7 @@ menuOrder: 8
 ---
 
 Pre-aggregations are materialized query results persisted as tables.
-In order to start using pre-aggregations Cube.js should have write access to `stb_pre_aggregations` schema, or whatever your [preAggregationsSchema]((@cubejs-backend-server-core#options-reference-pre-aggregations-schema)) setting is set to, where pre-aggregation tables will be stored.
+In order to start using pre-aggregations Cube.js should have write access to `stb_pre_aggregations` schema, or whatever your [preAggregationsSchema](@cubejs-backend-server-core#options-reference-pre-aggregations-schema) setting is set to, where pre-aggregation tables will be stored.
```
ferrants: fixed link I added last week

docs/Schema/cube.md
```diff
@@ -121,7 +121,7 @@ cube(`OrderFacts`, {
 cube(`ExtendedOrderFacts`, {
   extends: OrderFacts,
 
-  measure: {
```
ferrants: `measure` didn't work in my testing

packages/cubejs-schema-compiler/adapter/BaseQuery.js
```diff
@@ -952,17 +952,31 @@ class BaseQuery {
           this.safeEvaluateSymbolContext().leafMeasures[this.safeEvaluateSymbolContext().currentMeasure] = true;
         }
       }
+      if (this.safeEvaluateSymbolContext().usedInAggregationWithFilter){
```
ferrants: not sure if any other checks need to be made here

packages/cubejs-schema-compiler/adapter/BaseQuery.js
```diff
@@ -990,6 +1004,8 @@ class BaseQuery {
           "','",
           this.autoPrefixAndEvaluateSql(cubeName, symbol.longitude.sql)
         ])
+      } else if (this.safeEvaluateSymbolContext().rollupQuery) {
```
ferrants: not sure if any other checks are needed here

ferrants:
@paveltiunov , should we run linting on this package? We could do prettier/eslint on the whole thing in one different PR with no functional changes.

paveltiunov:
@ferrants Hey Matt! Just reviewed code and your solution is completely viable. I was thinking about old problem which would be really great to fix in scope of this change: generalization of `renderedReference` usage. In fact it's very powerful mechanism which is used only for measures but can be used for dimensions as well. It solves your issue and many other sophisticated cases for Leaf Measure Additive rollup usage. 

So the idea is pretty simple:
1. Move `renderedReference` usage from `renderSqlMeasure` to top of `evaluateSymbolSql`: https://github.com/cube-js/cube.js/blob/master/packages/cubejs-schema-compiler/adapter/BaseQuery.js#L1121
2. Pass `dimensions` as `renderedReference` together with measures: https://github.com/cube-js/cube.js/blob/6a91a8e2a51c279752548922b25b7e2fe7a838dd/packages/cubejs-schema-compiler/adapter/PreAggregations.js#L446

It should do all the job for your use case as well.

packages/cubejs-schema-compiler/adapter/BaseQuery.js
```diff
@@ -928,6 +928,14 @@ class BaseQuery {
   }
 
   evaluateSymbolSql(cubeName, name, symbol) {
+    const measurePath = `${cubeName}.${name}`;
+    if (this.safeEvaluateSymbolContext().usedInAggregationWithFilter){
+        return this.escapeColumnName(this.aliasCubeField(cubeName, name))
```
ferrants: this is specifically needed to avoid nested aggregation functions

packages/cubejs-schema-compiler/adapter/BaseQuery.js
```diff
@@ -952,17 +960,27 @@ class BaseQuery {
           this.safeEvaluateSymbolContext().leafMeasures[this.safeEvaluateSymbolContext().currentMeasure] = true;
         }
       }
+      let evaluatedSql;
+      if (this.safeEvaluateSymbolContext().rollupQuery && symbol.type === 'sum' && symbol.filters) {
```
paveltiunov: @ferrants Great! I think we're almost there! I believe we don't need these changes if we add timeDimensions to `renderedReference` as well. We should also get rid of these lines then:
https://github.com/cube-js/cube.js/blob/4705af080cd9549ff735f3ab6ea99dfd8de8b4ac/packages/cubejs-schema-compiler/adapter/BaseQuery.js#L894-L897

ferrants: This didn't work. I did add the `timeDimensions`, but the dimension, `source` I'm using for this measure filter in the tests isn't a timeDimension. Also, there is an issue with nested aggregate functions if I remove these lines. I end up with a:
```
sum(case when "visitors__source" = 'google' then sum("visitors__count") end)
```

paveltiunov:
To sum up all the discussions currently we have incorrect test in code base:
```
        google_count: {
          sql: `${CUBE.count}`,
          type: 'sum',
          filters: [
              {
                  sql: `${CUBE.source} = 'google'`
              }
          ],
        }
```
`google_count` can't reference `CUBE.count` in fact because it just can't work without pre-aggregation which is incorrect. Your case is pretty the same. It worked because of coincidence mostly not because it was designed to.

Grand solution to this issue is filtered measures should look after same measure definitions but without filters to be used as potential re-usable measures.

PR #284, opened 2019-12-03
docs: Update pre-aggregations on time hierarchy

Justin-ZS:
Update doc since we have already support pre-aggregation on time hierarchy.

I also think maybe we should consider the time priority when searching for matchable pre-aggregations.
e.g. If there exists both a pre-aggregation on `month` and a pre-aggregation on `day`,  query on `month` should be matched to the pre-aggregation on `month` firstly.


paveltiunov:
@Justin-ZS Hey Jin! Yeah. Right. Thanks for providing this!

PR #287, opened 2019-12-07
feat: hooks for dynamic schemas

willhausman:
**Check List**

Exposes option to customize the `RequestContext` and pass it to the `CompilerApi` and `FileRepository` per request.

Adds `lru-cache` to `CompilerApi` for managing many schemas per tenant.  Exposes options to optimize the cache size, and automatically prune the cache for expired items.

**Business Justification**
If schemas need to change based on grant permissions for row and column level security, the number of variations become a memory limitation.  A LRU Cache allows the most frequently used schemas to stay in memory until they are no longer needed, or a defined amount of time passes and grants should be checked again.
Exposing the creation of the context lets you send additional information to your cube server for use in dynamically generating schemas without requiring you to override the checkAuth hooks.

**New server create options**
`requestToContext` customizes the context. (Should consider enforcing `authInfo` is added?)
`schemaKey` for using more than 1 entry in the `CompilerApi` cache.
`compilerCacheSize` is the maxSize of the `lru-cache`.
`maxCompilerCacheKeepAlive` is the maxAge of the `lru-cache` in ms.

**Breaking changes**
`FileRepository.dataSchemaFiles(options)` instead of `FileRepository.dataSchemaFiles(includeDependencies)`.  `includeDependencies` is still supported through destructuring.  Packages in this repo are not using this option, so it affects adopters who customized the server but used the provided `FileRepository`.

**Example usage**
```javascript
CubejsServerCore.create({
  requestToContext: (req) => ({ authInfo: req.authInfo, schemaName: req.query.schemaName }),
  schemaKey: ({ authInfo, schemaName }) => `${schemaName}-${authInfo.sub}`,
  maxCompilerCacheKeepAlive: 30 * 60 * 1000, // 30 minutes
  repositoryFactory: () => ({
    dataSchemaFiles: async ({ authInfo, schemaName }) => {
      // ...
    }
  })
});
```

### DIFF
--- a/docs/Cube.js-Backend/@cubejs-backend-server-core.md
+++ b/docs/Cube.js-Backend/@cubejs-backend-server-core.md
@@ -48,7 +48,6 @@ Both [CubejsServerCore](@cubejs-backend-server-core) and [CubejsServer](@cubejs-
   logger: (msg: String, params: Object) => any,
   driverFactory: (context: DriverContext) => BaseDriver,
   externalDriverFactory: (context: RequestContext) => BaseDriver,
-  requestToContext: (req: ExpressRequest) => RequestContext,
   contextToAppId: (context: RequestContext) => String,
   contextToDataSourceId: (context: RequestContext) => String,
   repositoryFactory: (context: RequestContext) => String,
@@ -188,18 +187,6 @@ CubejsServerCore.create({
 CubejsServerCore.create({
   contextToAppId: ({ authInfo }) => `CUBEJS_APP_${authInfo.user_id}`,
   contextToDataSourceId: ({ authInfo }) => `CUBEJS_APP_${authInfo.tenantId}`
-});
-```
-
-### requestToContext
-
-`requestToContext` is a function to customize the context object passed to other hooks.  Default implementation sets authInfo from the payload.  Custom headers and query parameters could be separated into a useful context here.
-
-```javascript
-const independantInfoSource = {};
-
-CubejsServerCore.create({
-  requestToContext: (req) => ({ authInfo: req.authInfo, customProp: independantInfoSource.currentValue })
 });
 ```
 
--- a/docs/Cube.js-Backend/@cubejs-backend-server-core.md
+++ b/docs/Cube.js-Backend/@cubejs-backend-server-core.md
@@ -48,12 +48,16 @@ Both [CubejsServerCore](@cubejs-backend-server-core) and [CubejsServer](@cubejs-
   logger: (msg: String, params: Object) => any,
   driverFactory: (context: DriverContext) => BaseDriver,
   externalDriverFactory: (context: RequestContext) => BaseDriver,
+  requestToContext: (req: ExpressRequest) => RequestContext,
   contextToAppId: (context: RequestContext) => String,
   repositoryFactory: (context: RequestContext) => String,
   checkAuthMiddleware: (req: ExpressRequest, res: ExpressResponse, next: ExpressMiddleware) => any,
   queryTransformer: (query: Object, context: RequestContext) => Object,
   preAggregationsSchema: String | (context: RequestContext) => String,
   schemaVersion: (context: RequestContext) => String,
+  compilerCacheSize: Number,
+  maxCompilerCacheKeepAlive: Number,
+  updateCompilerCacheKeepAlive: Boolean,
   telemetry: Boolean,
   allowUngroupedWithoutPrimaryKey: Boolean,
   orchestratorOptions: {
@@ -175,6 +179,18 @@ CubejsServerCore.create({
 })
 ```
 
+### requestToContext
+
+`requestToContext` is a function to customize the context object passed to other hooks.  Default implementation sets authInfo from the payload.  Custom headers and query parameters could be separated into a useful context here.
+
+```javascript
+const independantInfoSource = {};
+
+CubejsServerCore.create({
+  requestToContext: (req) => ({ authInfo: req.authInfo, customProp: independantInfoSource.currentValue })
+});
+```
+
 ### repositoryFactory
 
 This option allows to customize the repository for Cube.js data schema files. It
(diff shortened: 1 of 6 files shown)


willhausman:
I ran the rollup to generate the UMD files as noted in the CONTRIBUTING doc, but it made a crap ton of changes.  Is that expected?

paveltiunov:
@willhausman Hey Will! Wow! This one is really huge! Really appreciate this contribution!

I was thinking about this grant based permission problem for a while. I'd suggest actually to introduce `metaTransformer` in the same way as `queryTransformer` here:
https://github.com/cube-js/cube.js/blob/6df008ed178ac4ad82688030cdd30060d26d292d/packages/cubejs-server-core/core/CompilerApi.js#L85

It can look like:
```
  async metaConfig(context) {
    return this.metaTransformer((await this.getCompilers()).metaTransformer.cubes, context);
  }
```

This way you can control it on per-user basis and also you don't need to recompile all of the stuff. Do you think it can work for you?

willhausman:
@paveltiunov Conceptually, I do think creating a `metaTransformer` could work for me, but I prefer this approach for two reasons.

1. The `compilers` object is more complex to try and customize than a `query`.  As a dev, I would have to learn a lot more about the innards of cubeJs to customize the meta correctly, vs. knowing how to edit the cube file I possibly wrote by hand.
2. It would be more way chatty. If I need to acquire my transformations from another server, I would need to duplicate that behavior for both `metaTransformer` and `queryTransformer` on *every request* to cubejs.  Collecting multiple compilers means I can just tell an implementation of FileRepository the request context and let it give me the exact schema to compile - the rest of cubeJs continues to work as if we were still in a single-compiler scenario.

Adding the LRU was  just for memory management on the number of compilers I would need. The real root of the capability I personally need is just passing custom query params to the call for `FileRepository`. I could handle the rest with existing multi-tenancy support.

As an aside, if we go with this approach, I think the `schemaVersion()` should be altered to pass the current request context instead of continuing to pass the first request context used to create the `CompilerApi`.  It wouldn't change anything for me, but it could certainly confuse someone if they were trying to debug a problem and expected the two contexts to be the same... though that also begs the question, can the `CompilerApi` just be directly used to handle multi-tenancy instead of `contextToAppId` now? I'm not sure that keeping both levels of caching buys anything anymore.

paveltiunov:
@willhausman Gotcha. Makes sense. We can stick with multiple schema approach as it provides additional flexibility. I briefly reviewed changes and overall it's a right direction. We need just to simplify it a little:
1. I believe `schemaKey` does the same job as `contextToAppId`. The only difference right now is how many orchestrators are allocated for specific context. To control it we can override it and get rid off all of `CompilerApi` changes then: https://github.com/cube-js/cube.js/blob/6df008ed178ac4ad82688030cdd30060d26d292d/packages/cubejs-server-core/core/index.js#L280.
2. RLU cache can be applied to `dataSourceIdToOrchestratorApi` after it: https://github.com/cube-js/cube.js/blob/6df008ed178ac4ad82688030cdd30060d26d292d/packages/cubejs-server-core/core/index.js#L288. We can also consider adding it for orchestrators cache.
3. `repositoryFactory` already receives `context` so it's just `RepositoryFactory` implementation responsibility on how to handle it. `FileRepository` class can be overridden by users in order to handle it.

willhausman:
@paveltiunov Yes, I came to the same conclusion about `schemaKey` and `contextToAppId`.  I'll move the LRU out of the `CompilerApi` and into `CubejsServerCore.getCompilerApi()` using the `contextToAppId` as the key.

Thanks for mentioning the `repositoryFactory`.  In my original implementation, I did not add a hook for customizing the `context`, so the repositoryFactory was insufficient for my needs as the `context` lacked additional per-request information.  With the LRU in the `CubejsServerCore` and the `requestToContext` customizing the `context`, my needs are completely met.

I need more information concerning the orchestrators ask, though. I never looked into them, so I don't know what role they play in these changes. Are you just asking me to apply the caching strategy to `getOrchestratorApi()` as well?

I'll go ahead and make the other changes in the mean time.

paveltiunov:
@willhausman `OrchestratorApi` is responsible for maintaining connections to database and query execution queues. Every instance has its own connection pool and queue. So it's important to maintain right quantitative mapping between orchestrators and compilers. 

LRU support for `OrchestratorApi` is not required for your case so you can skip it. I'm still not sure about exact use case for that either. Only infrequent database usage in case of large amount of tenants comes to mind.

willhausman:
@paveltiunov I've simplified the PR as we talked about.

Do orchestrators need to be kept in sync with compilers, or just relative number of connections?  Since I'm treating users as tenants, doesn't that mean I do need the `OrchestratorApi` gated behind the LRU cache now?

paveltiunov:
@willhausman Looks great! We're almost there!

For orchestrators you probably don't want to create `OrchestratorApi` for each new user. You just want it to manage per database probably. In order to do this we'd need to introduce `contextToDataSourceId` server option and use it here: https://github.com/cube-js/cube.js/blob/6df008ed178ac4ad82688030cdd30060d26d292d/packages/cubejs-server-core/core/index.js#L280
This way you can control when `OrchestratorApi` is instantiated.

I'm curious what type of info is provided to context using `requestToContext` that `authInfo` can't provide right now?

willhausman:
@paveltiunov 2 things for `requestToContext`.

1. Better separation of concerns.  If I'm adding a custom query param, it doesn't make sense to me that this parameter would exist as part of `authInfo` when it isn't auth related.
2. Simplifies customization. To do the same thing with `authInfo`, I have to override `checkAuth` or `checkAuthMiddleware` which I might not otherwise need to do at all. Otherwise, the express request is never exposed. I am currently doing this in my implementation to grab the request's `correlationId`, and it feels very out of place to me.

It sounds like I'll need `contextToDataSourceId` for my implementation.  I'll add that with the current implementation as the default fallback.

paveltiunov:
@willhausman I see. Could you please elaborate on what type of param you need to pass through `requestToContext`? There's another place where context is created: https://github.com/cube-js/cube.js/blob/master/packages/cubejs-api-gateway/SubscriptionServer.js#L29
So if transport switches to web sockets things will break.

willhausman:
@paveltiunov Is the `SubscriptionServer` even being used? It is only referenced by the `ApiGateway`, but the `/subscribe` endpoint just calls `load()`.

My intent is query parameters and custom headers on the REST endpoint itself.  I should be able to send an extra query param to the API, and have that extra param show up when trying to customize `contextToAppId` or `repositoryFactory`.  I don't need it, but I would think the same thing would be needed for a subscription.

How I want to use it:
```javascript
CubejsServerCore.create({
  requestToContext: (req) => ({ 
    authInfo: req.authInfo,
    schemaName: req.query.schemaName,
    correlationId: req.headers['my-custom-correlation-header'] 
  }),
  contextToAppId: ({ authInfo, schemaName }) => `${schemaName}-${authInfo.sub}`,
  repositoryFactory: () => ({
    dataSchemaFiles: async ({ authInfo, schemaName, correlationId }) => {
      // make request to microservice, passing the correlationId as a header
    }
  })
});
```

Because I can't load everything in advance, I **need** a way to tell the repositoryFactory *which* cubes to compile.

paveltiunov:
@willhausman Yep. It used by other packages. Could you please elaborate what `correlationId` does? Does it change schema?

willhausman:
@paveltiunov I've worked out another way to do this, so removing the hook from the PR.  I still think the option should exist, but it's not worth holding up the PR.

paveltiunov:
@willhausman Ok. Cool! Looks great! Thanks for contributing this!

PR #288, opened 2019-12-11
feat: type for returning specific http status codes

willhausman:
Simple type that can be thrown to return status codes other than 200, 400, or 500.

Specifically useful for communicating with another server within `repositoryFactory` and propagating a 401 for unauthorized access to a schema.

### DIFF
--- a/packages/cubejs-api-gateway/CubejsHandlerError.js
+++ b/packages/cubejs-api-gateway/CubejsHandlerError.js
@@ -1,4 +1,4 @@
-class HttpError extends Error {
+class CubejsHandlerError extends Error {
   constructor(status, type, message) {
     super(message || type);
     this.status = status;
@@ -6,4 +6,4 @@ class HttpError extends Error {
   }
 }
 
-module.exports = HttpError;
+module.exports = CubejsHandlerError;
--- a/packages/cubejs-api-gateway/HttpError.js
+++ b/packages/cubejs-api-gateway/HttpError.js
@@ -1,0 +1,9 @@
+class HttpError extends Error {
+  constructor(status, type, message) {
+    super(message || type);
+    this.status = status;
+    this.type = type;
+  }
+}
+
+module.exports = HttpError;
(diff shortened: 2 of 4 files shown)


paveltiunov:
@willhausman Hey Will! Thanks for submitting it! What if we add this optional status parameter to `User` error? Or add something like `UnauthorizedError` or `ForbiddenError`? `HttpError` implies to be protocol specific but still can be used for protocols other than http.

willhausman:
@paveltiunov Sure. I was going back and forth on the name. The intent was "returns an error with an http status code", and it wasn't necessarily user error. I didn't want to make a PR for every type of status to be returned, either.  Not when we can use inheritance.  How about a `CubejsHandlerError` and change `UserError` to extend from that base? 

Replace
```javascript
    if (e instanceof UserError) {
      this.log(context, {
        type: 'User Error',
        query,
        error: e.message,
        duration: this.duration(requestStarted)
      });
      res({ error: e.message }, { status: 400 });
    } else if (e instanceof HttpError) {
      this.log(context, {
        type: e.type,
        query,
        error: e.message,
        duration: this.duration(requestStarted)
      });
      console.log(`status: ${e.status}`);
      res({ error: e.message }, { status: e.status });
    } else if (e.error === 'Continue wait') {
```
with

```javascript
    if (e instanceof CubejsHandlerError) {
      this.log(context, {
        type: e.type,
        query,
        error: e.message,
        duration: this.duration(requestStarted)
      });
      console.log(`status: ${e.status}`);
      res({ error: e.message }, { status: e.status });
    } else if (e.error === 'Continue wait') {
```

willhausman:
Or just StatusError.

paveltiunov:
@willhausman Yeah. `CubejsHandlerError` or `StatusError` works.

willhausman:
@paveltiunov done.

paveltiunov:
@willhausman Looks great!

Issue #302, opened 2019-12-17
How to add a new healthcheck to the cube server?


hemanthsj:
I am trying to add a new /health endpoint to my implementation of the cube js backend and i was wondering if there is a preferred method of doing that?

paveltiunov:
@hemanthsj Hey Hemanth! You can do so using express: https://cube.dev/docs/deployment#express but I believe we should provide this route as part of API. Let's track it's implementation in this issue.

willhausman:
@hemanthsj if the healthcheck endpoint is the only thing you need to customize about the server itself, you don't have to implement a whole express server and attach `CubejsServerCore`.  `CubejsServer.listen()` resolves a reference to `app` which is the express app. You can use it to add an extra endpoint.

I'm doing effectively this:
```javascript
const CubejsServer = require('@cubejs-backend/server');

const server = new CubejsServer();

server.listen().then(({ app, port }) => {
  app.get('/ready', (req, res) => res.status(200).send('Healthy'));
  console.log(`🚀 Cube.js server is listening on ${port}`);
});
```

hemanthsj:
Thank you @willhausman

PR #308, opened 2019-12-23
feat: Health check methods

willhausman:
#302 

These are methods that can be used for checking health.  I did not explicitly add an endpoint because the format of returning health checks varies slightly.  These are methods that can be called with a custom implementation or using another package, e.g. terminus or lightship.


paveltiunov:
@willhausman Looks good to me! Also agree on endpoint. Would be nice to have it in the docs though as it may not seem obvious on how to add this health check route.

hemanthsj:
Thank you @willhausman

Issue #306, opened 2019-12-21
Automatically retrieve Redshift cluster credentials


hassankhan:
**Is your feature request related to a problem? Please describe.**
The AWS Redshift SDK allows retrieving credentials via an API:

```typescript
import Redshift, { ClusterCredentials } from 'aws-sdk/clients/redshift';

const getCredentials = async () => {
  const creds = await redshift.getClusterCredentials({
    ClusterIdentifier: process.env.CUBEJS_DB_HOST,
    DbUser: process.env.CUBEJS_DB_USER,
  }).promise();

  if (!creds.DbUser && !creds.DbPassword) {
    throw new Error('Unable to retrieve Redshift credentials');
  }

  return creds as Required<ClusterCredentials>;
};

getCredentials()
  .then((credentials) => {
    const config = {
      database : process.env.CUBEJS_DB_NAME,
      host: process.env.CUBEJS_DB_HOST,
      password : creds.DbPassword,
      port: process.env.CUBEJS_DB_PORT,
      user: creds.DbUser,
    };
  })
```

**Describe the solution you'd like**

It would be great if CubeJS would automatically attempt to do the above if the `CUBEJS_DB_PASS` environment variable is omitted for Redshift projects.

**Additional context**
A Serverless project using Redshift would require additional permissions to the IAM role:

```yaml
iamRoleStatements:
    - Effect: 'Allow'
      Action:
        - 'redshift:GetClusterCredentials'
      Resource:
        - '<REDSHIFT_CLUSTER_ARN>/<REDSHIFT_USER>'
```

paveltiunov:
@hassankhan Hey Hassan! Thanks for posting this! I think it would be great to support this in case credentials aren't passed in environment or options.

hassankhan:
I could take this on, would love a little bit of guidance though ... the code snippet I pasted above expects to run asynchronously ... however looking through the Postgres driver source, it seems that [the pool is instantiated in the constructor](https://github.com/cube-js/cube.js/blob/master/packages/cubejs-postgres-driver/driver/PostgresDriver.js#L20-L30). How would you recommend approaching this, @paveltiunov?

paveltiunov:
@hassankhan Sounds great! I believe you can instantiate it lazily when pool is accessed in `query()` or `testConnection()` methods. I also believe we should create separate package `@cubejs-backed/redshift-driver` for this and extend `PostgresDriver` there.

Issue #266, opened 2019-11-15
External Preaggregations in RDS for Cubes querying Redshift should pass through S3 using UNLOAD and COPY command 


ferrants:
**Is your feature request related to a problem? Please describe.**
I am trying to persist sizable preaggregate tables in an RDS Database, sourced from a Redshift Database while using Cube.js Serverless. Loading the external table through the running cube.js instance is inefficient for Redshift -> RDS data transfers.

**Describe the solution you'd like**
For persisting preaggregates in RDS from Redshift, instead of passing trough Cube.js, allow the Redshift Driver / RDS Driver (currently just Postgres driver, consider sub-classing) to run an [UNLOAD](https://docs.aws.amazon.com/redshift/latest/dg/r_UNLOAD.html) command on Redshift to land the results in S3, then run the COPY command from RDS to ingest to the database (https://aws.amazon.com/about-aws/whats-new/2019/04/amazon-rds-postgresql-supports-data-import-from-amazon-s3/). This will greatly speed up external preaggregations.

Logic seems to be in here:
https://github.com/cube-js/cube.js/blob/6c81a02776a092f8468332d979c6ac273eeca211/packages/cubejs-query-orchestrator/orchestrator/PreAggregations.js#L357
https://github.com/cube-js/cube.js/blob/feccdb5c40b1dec2b6ef9d39e6947e1c0c90c7db/packages/cubejs-query-orchestrator/driver/BaseDriver.js#L102-L104

**Describe alternatives you've considered**
Just using the slower cube.js method, but it's been slower than I'd like and with the Serverless approach, I feat we'll run into time and memory constraints.

**Additional context**
I could probably add support for this.

paveltiunov:
@ferrants Yeah. Right. Every driver should implement it's own download tailored for performance capabilities. It would be great if you can contribute it! We can consider option of subclassing as well as providing some options for Postgres driver.

narisso:
Hi, I recently had this issue and ended up using the dblink extension from postgres to redshift, it works like a charm with minor modifications to the drivers with a subclass.

paveltiunov:
@narisso Hey Nicolas! Interesting. Are you using dblink to download results from redshift to postgres? And what modifications did you make to driver?

ferrants:
@narisso , I'm glad that worked for you. I was using dblink to generate materialized views of  preaggregates (not for cube.js), but it was taking too long to generate the views for the large tables I was materializing. This method should definitely work for smaller data though.

[...]

PR #320, opened 2019-12-30
feat(@cubejs-server/redshift-driver): create driver for Redshift

hassankhan:
**Check List**

#306. Also creates a foundation for #266.

Added a new package `@cubejs-backend/redshift-driver` according to the `CONTRIBUTING.md` guidelines, tested using `@cubejs-backend/serverless-aws`. If the `CUBEJS_DB_PASS` environment variable is set, then it is used, otherwise the driver attempts to retrieve credentials from AWS.


paveltiunov:
@hassankhan Hey Hassan! Looks great! The only thing: can we fallback to existing behavior if `CUBEJS_DB_USER` or `CUBEJS_DB_PASS` are defined?

hassankhan:
Hi @paveltiunov, thanks for checking out the PR 😄 

I've added the fallback, as well as a really basic test suite that covers the Redshift-specific additions to the `PostgresDriver`. Just wondering if my tests will automatically get run by the CI or do I need to set that up somewhere?

packages/cubejs-redshift-driver/driver/RedshiftDriver.js
```diff
@@ -0,0 +1,82 @@
...
+  getCredentialsFromEnv() {
+    return {
+      user: process.env.CUBEJS_DB_USER,
+      password: process.env.CUBEJS_DB_PASS,
+    };
+  }
+
+  async createPool() {
+    if (this.pool && this.pool instanceof Pool) {
+      return;
```
paveltiunov: Just realized we need mutex here that will ensure only one pool is creating at the same time in case multiple queries arrived at the same time.

hassankhan: Is there an example somewhere else in the codebase?

paveltiunov: Yep. https://github.com/cube-js/cube.js/blob/26e5843de6a844613ef6e10f0a061c84200c57b3/packages/cubejs-server-core/core/index.js#L321

igorlukanin:
Hey @hassankhan, is this branch something we'd like to keep open? It just appeared in my GitHub notifications, that's why :)

PR #367, opened 2020-02-10
fix: Wrong interval functions for BigQuery

lvauvillier:
Fix BigQuery error:

```
Dropping Cache: {
  "cacheKey": [
    "SELECT CASE\n    WHEN CURRENT_TIMESTAMP() < DATETIME_ADD(TIMESTAMP(?), INTERVAL 7 day) THEN FLOOR(UNIX_SECONDS(CURRENT_TIMESTAMP()) / 86400) END",
    [
      "2020-01-12T23:59:59Z"
    ]
  ]
} 
Error: No matching signature for function DATETIME_ADD for argument types: TIMESTAMP, INTERVAL INT64 DATE_TIME_PART. Supported signature: DATETIME_ADD(DATETIME, INTERVAL INT64 DATE_TIME_PART) at [2:32]
```

### DIFF
--- a/packages/cubejs-schema-compiler/adapter/BaseQuery.js
+++ b/packages/cubejs-schema-compiler/adapter/BaseQuery.js
@@ -312,6 +312,14 @@ class BaseQuery {
     return `${date} + interval '${interval}'`;
   }
 
+  addTimestampInterval(timestamp, interval) {
+    return this.addInterval(timestamp, interval);
+  }
+
+  subtractTimestampInterval(timestamp, interval) {
+    return this.subtractInterval(timestamp, interval);
+  }
+
   cumulativeMeasures() {
     return this.measures.filter(m => m.isCumulative());
   }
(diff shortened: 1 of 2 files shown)


paveltiunov:
@lvauvillier Thanks for contributing this one! `DATETIME` here may seem weird but it's correct implementation but probably wrong name right now. Bug is on this line: https://github.com/cube-js/cube.js/blob/c72ab079d53458b7377bc12f498667110af18689/packages/cubejs-schema-compiler/adapter/BaseQuery.js#L1533. So the solution is to introduce `addTimestampInterval` and `subtractTimestampInterval` methods which should use `addInterval` and `subtractInterval` in base implementation and should be overridden in `BigqueryQuery`.

lvauvillier:
Thanks @paveltiunov. I have just pushed an update.

paveltiunov:
@lvauvillier Looks great! Thanks again!

PR #376, opened 2020-02-13
fix(@cubejs-client/core): improve types

hassankhan:
**Check List**

Loosen the types enough so that common issues go away. Long term, it might be better to replace the default generic type for `ResultSet` from `any` to `unknown`.


paveltiunov:
Hey @hassankhan ! Thanks for providing it!

Issue #239, opened 2019-10-24
Support for External Pre-aggregation without writing to DB (MongoDB case)


MuhammadMattiullah:
I am trying to use pre-aggregation feature of cube js which is indeed very helpful. I am using mongodb and connecting it with mongodb connectorBI i.e mongosqld. 
I have created a pre-aggregation query but it's not working and not creating any pre-aggregation tables as well.  
Adding a support of pre-aggregation on mongosqld or anyother mongodb connectorBI.

paveltiunov:
@MuhammadMattiullah Thanks for creating this one! Here we should support download of pre-aggregations data without creating tables in main DB because MongoDB BI connector doesn't allow to create tables.

MuhammadMattiullah:
@paveltiunov can you tell me how much time will it take to implement this feature ? 
Any rough estimation ? 
So i can integrate this feature in future project.

paveltiunov:
@MuhammadMattiullah Great question! So to implement it we'd need just to download results here: https://github.com/cube-js/cube.js/blob/master/packages/cubejs-query-orchestrator/orchestrator/PreAggregations.js#L325 
instead of creating a table and pass here:
https://github.com/cube-js/cube.js/blob/master/packages/cubejs-query-orchestrator/orchestrator/PreAggregations.js#L336
This should should be fallback behavior so I think driver can provide some boolean method like `supportsPreAggregationTables()` to check if it has this capability.

So overall not a lot of changes but it's not on our road map though. It'd great if someone else can pickup it.

PR #395, opened 2020-02-17
feat: Support external rollups from readonly source

jcw-:
**Check List**

#239 

- Refactors external rollups into three strategies (the third being new):
  - Store In Source
  - Temp Table External
  - Stream External
- Removes ability to cancel temp table creation (not sure if this is actually used or not, can add it back in if needed, it was just getting complicated with the refactor)
- Adds a `readOnly` config to drivers (defaults to `false`) that controls which external strategy is used for rollups
- Adds capability for both row and type data to be returned at query time, called `downloadQueryResults` and implemented so far for Postgres only, as it already had the info it just needed to be surfaced

Example config for specifying a read-only source driver:
```
const CubejsServer = require('@cubejs-backend/server');
const PostgresDriver = require('@cubejs-backend/postgres-driver');

const options = {
  driverFactory: () => new PostgresDriver({
    readOnly: true
  }),
  externalDbType: 'postgres',
  externalDriverFactory: () => new PostgresDriver({
    host: 'my_host',
    database: 'my_db',
    user: 'my_user',
    password: 'my_pw'
  })
};
```

### DIFF
--- a/packages/cubejs-query-orchestrator/driver/BaseDriver.js
+++ b/packages/cubejs-query-orchestrator/driver/BaseDriver.js
@@ -38,15 +38,15 @@ class BaseDriver {
   }
 
   testConnection() {
-    throw 'Not implemented';
+    throw new Error('Not implemented');
   }
 
   query() {
-    throw 'Not implemented';
+    throw new Error('Not implemented');
   }
 
   downloadQueryResults() {
-    throw 'Not implemented';
+    throw new Error('Not implemented');
   }
 
   readOnly() {
--- a/packages/cubejs-query-orchestrator/driver/BaseDriver.js
+++ b/packages/cubejs-query-orchestrator/driver/BaseDriver.js
@@ -45,6 +45,14 @@ class BaseDriver {
     throw 'Not implemented';
   }
 
+  downloadQueryResults() {
+    throw 'Not implemented';
+  }
+
+  readOnly() {
+    return false;
+  }
+
   tablesSchema() {
     const query = this.informationSchemaQuery();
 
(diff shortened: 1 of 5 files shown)


packages/cubejs-query-orchestrator/orchestrator/PreAggregations.js
```diff
@@ -387,42 +387,71 @@ class PreAggregationLoader {
...
+    }
+
+    this.logger('Downloading external pre-aggregation via query', {
+      preAggregation: this.preAggregation,
+      requestId: this.requestId
+    });
+    const tableData = await client.downloadQueryResults(sql, params);
+    await this.uploadExternalPreAggregation(tableData, newVersionEntry);
+    await this.loadCache.reset(this.preAggregation);
+    await this.dropOrphanedTables(client, this.targetTableName(newVersionEntry));
```
paveltiunov: This seems not required as tables aren't created while downloading

packages/cubejs-query-orchestrator/driver/BaseDriver.js
```diff
@@ -45,6 +45,14 @@ class BaseDriver {
     throw 'Not implemented';
   }
 
+  downloadQueryResults() {
+    throw 'Not implemented';
```
paveltiunov: It would be great to have `new Error('Not implemented')` instead of `'Not implemented'` here as it'll be easier to find this line when it's thrown.

paveltiunov:
Hey @jcw- ! Looks great! Thanks for contributing this one!

PR #421, opened 2020-02-20
cast double to decimal for postgres

sandeepravi:
**Check List**

While pre-aggregating data from Aws Athena to postgres, there was an issue because Athena supports "double" whereas postgres does not.

Added a mapping to convert "double" fields to "decimal"


paveltiunov:
@sandeepravi Hey Sandeep! Thanks for contributing this one! Do you think it should map to `decimal` or to `double precision`?

sandeepravi:
yeah, I thought about this but I thought decimal should cover all cases as there's no limit on the storage.

But yeah, since postgres supports "double precision", it can be double precision as well.

paveltiunov:
@sandeepravi Yep. Agree! Let's stick with decimal for now and see how it goes. Thanks for contributing this one!

PR #433, opened 2020-02-24
feat: Redis connection pooling

### LINKED ISSUE #104: Too many Redis connections: ERR max number of clients reached on Heroku's free tier
A clear and concise description of what the problem is. Ex. I'm always frustrated when 

I use Heroku for Cube.js deployments. If I make 3+ concurrent queries to the server, the app instance will crash with the error ERR max number of clients reached. Heroku's free tier allows for 20 connections at once.

As mentioned by Pavel on the Cube.js Slack channel, a connection poll would probably solve this issue to prevent Cube from using more than is available.

The only alternative I can think of is to pay more for Heroku. I've had to go up to Premium 2 to not crash, but I'm worried about what happens when more users are generating reports at the same time.

**Additional context**

jcw-:
**Check List**

Fixes #104 

Adds a connection pool class based on `generic-pool`. By default, uses (min=2, max=10). If max is set to zero (e.g. through the env var), then the pool is disabled and it falls back to prior behavior (non-pooled).

A single instance of the connection pool is created in the `QueryOrchestrator` and is wired up to two final destinations:
- `RedisQueueDriverConnection`
- `RedisCacheDriver`

### DIFF
--- a/packages/cubejs-query-orchestrator/orchestrator/RedisCacheDriver.js
+++ b/packages/cubejs-query-orchestrator/orchestrator/RedisCacheDriver.js
@@ -1,25 +1,47 @@
-const createRedisClient = require('./RedisFactory');
-
 class RedisCacheDriver {
-  constructor() {
-    this.redisClient = createRedisClient(process.env.REDIS_URL);
+  constructor(pool) {
+    this.redisPool = pool;
+  }
+
+  async getClient() {
+    return await this.redisPool.getClient();
   }
 
   async get(key) {
-    const res = await this.redisClient.getAsync(key);
-    return res && JSON.parse(res);
+    const client = await this.getClient();
+    try {
+      const res = await client.getAsync(key);
+      return res && JSON.parse(res);
+    } finally {
+      this.redisPool.release(client);
+    }
   }
 
-  set(key, value, expiration) {
-    return this.redisClient.setAsync(key, JSON.stringify(value), 'EX', expiration);
+  async set(key, value, expiration) {
+    const client = await this.getClient();
+    try {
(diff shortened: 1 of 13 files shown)


jcw-:
@paveltiunov I could use some extra eyes on this in terms of testing, it is working locally but I haven't exercised it as extensively as I wanted - one thing I want to do is to cause redis connections to ramp up and then watch that they are detected as idle and ramp back down.

I haven't gone out of my way to write extra tests for this - but I did convert to jest while I was troubleshooting something, hopefully that is ok, it seems to be in use in other areas of this project.

I also haven't updated any docs.

paveltiunov:
@jcw- Hey Jc! Looks really great! And thanks for migrating tests to jest! Really appreciate that!

packages/cubejs-query-orchestrator/orchestrator/RedisCacheDriver.js
```diff
@@ -1,25 +1,47 @@
-const createRedisClient = require('./RedisFactory');
-
 class RedisCacheDriver {
-  constructor() {
-    this.redisClient = createRedisClient(process.env.REDIS_URL);
+  constructor(pool) {
```
hassankhan: Would it be possible for the constructor options to be an object, with a property `pool` instead? This would be helpful in the future, in case we need to add any more parameters to instantiate a Redis driver

paveltiunov: @hassankhan Yeah. Great idea! I think we should provide options object to `RedisPool` instead of arguments.

jcw-: https://github.com/cube-js/cube.js/pull/468

PR #444, opened 2020-02-25
refactor: replace `joi` with `ajv`

cemremengu:
**Check List**

This PR intends to replace Joi with AJV since it is a much faster and (imo) a more solid alternative. I intend to work on this as I have more available time. Please let me know if this replacement is something you are not very eager to do.


paveltiunov:
@cemremengu Hey Cemre! Thanks for providing this one! This is huge! Could you please elaborate a little bit on a use case which brought you to these changes? AJV seems really fast to me. Also on AJV capabilities side: do you know if it supports custom functions? Cube.js schema isn't necessarily a JSON but rather a complex JavaScript object which contains functions that still need to be validated.

cemremengu:
Hi @paveltiunov, I am not an expert on the capability side but as far as I know you can add custom keywords as shown in the link which should provide what you are asking:

https://ajv.js.org/custom.html

The schema is very raw at the moment but I think we can use `"$ref":` to reuse parts and make regex variables as it was in the Joi case.

ifokeev:
why do we need this? It's a big codebase change and `Joi` is more readable and writable than `Ajv`. If you prefer speed then it's better to disable schema validation if user wants or use websockets.

cemremengu:
I agree that Joi is more readable/writable but I don't think it is a priority as it doesn't change often. 

Also unlike Joi, JSON schema is universal so you can find a lot of tools for editing/validating/viewing or even [generating forms](https://github.com/rjsf-team/react-jsonschema-form) using the schema

ifokeev:
@cemremengu you can use `describe()` if you want to pass JSON to other systems https://github.com/hapijs/joi/blob/master/API.md#anydescribe 

`editing/validating/viewing or even generating forms using the schema`
not the case of `cubejs`

cemremengu:
I guess it is up to the maintainer at this point. I can continue to try or abort. @paveltiunov if all tests pass, do you think this will be nice to have or it would rather not be pleasant.

paveltiunov:
@cemremengu Hey Cemre! Really appreciate you taking a time for implementing this! I think before we move on with this one we need to weigh in rationale for switching. So we need to address specific questions:
1. What exact problem this PR will solve?
2. What's the cost of such implementation and maintenance?

Dozens of thousands servers run Cube.js right now and we're committed for backward compatibility as much as possible here.

Right now I see only one problem with Joi: error messages. It isn't really easy to read them.
In terms of performance I never saw it to be the bottleneck either on compile stage or even for API serving. AFAIK biggest time consumers are transpiler on compile stage and SQL generation on serving stage. It doesn't make sense to fix Joi while 90% of time is spent on something else. However I never looked at exact Joi times and it would be helpful if you can provide some benchmarks here.

Given that I would love to hear your thoughts on how you envision this problem?

rpaik:
@cemremengu are you still working on this PR? Should we keep this open?

cemremengu:
@rpaik unfortunately I never had the time to come back :( 
The intention here was to use the fastest schema validation as possible. However, given the comments I think it is fair to close it for the time being.

PR #475, opened 2020-03-04
feat: Add API gateway request logging support

jcw-:
**Check List**

Uses middleware technique to add logging of API Gateway requests using generic logger.

- Default prod log level changed from info to warn (there was nothing logged at info, so no apparent change here)
- Put 1-line per-request logging at info level, this means you'll need to set `CUBEJS_LOG_LEVEL=info` to see it

image

### DIFF
--- a/packages/cubejs-api-gateway/requestParser.js
+++ b/packages/cubejs-api-gateway/requestParser.js
@@ -1,0 +1,19 @@
+module.exports = (req, res) => {
+  const path = req.originalUrl || req.path || req.url;
+  const httpHeader = req.header && req.header('x-forwarded-for');
+  const ip = req.ip || httpHeader || req.connection.remoteAddress;
+  const requestData = {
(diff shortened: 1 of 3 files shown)


packages/cubejs-api-gateway/requestParser.js
```diff
@@ -0,0 +1,19 @@
+module.exports = (req, res) => {
+  const path = req.originalUrl || req.path || req.url;
+  const httpHeader = req.header && req.header('x-forwarded-for');
+  const ip = req.ip || httpHeader || req.connection.remoteAddress;
+  const requestData = {
+    path,
+    method: req.method,
+    status: res.statusCode,
+    ip,
+    time: (new Date()).toISOString(),
```
paveltiunov: There're also some pending changes for logging so it should be expected that timestamps will be handled in one place.

PR #521, opened 2020-03-18
fix(types): Fix index.d.ts errors in cubejs-server.

jwalton:
**Check List**

So... when I did [this](https://github.com/cube-js/cube.js/pull/518) I didn't realize there were some problems in the index.d.ts file.  :P

* index.d.ts imports a `CreateOptions` from server-core, but also defined a new conflicting CreateOptions.
* index.d.ts has an `export =`, which is forbidden because it also has an `export interface CreateOptions`, and you can't use`export =` if you have other exports in the same module.

This fixes both of these problems.


paveltiunov:
@jwalton Hey Jason! Thanks for fixing this one!

Issue #493, opened 2020-03-06
Max usage with dates


richipargo:
**Describe the bug**
When using the `type: max` in a measure over a date type field `@cubejs-server/core` generates the correct response but treats the field as a number, so when `@cubejs-client/core` parses the information it parses the date as a number cutting of most data provided

Running `chartPivot` function from `@cubejs-client/core` generates this data output
```
[{"category":"","x":"","ExecutionLogs.latestRun":2020}]
```

**Expected behavior**
I would like to have the type come correctly from API so that `chartPivot` renders proper data

**Version:**
^0.18.0 on both `@cubejs-client/core` & `@cubejs-server/core`

**Additional context**
Schema
```
cube(`ExecutionLogs`, {
  sql: `SELECT * FROM v_execution_logs`,
  measures: {
    latestRun: {
      sql: `startDate`,
      type: `max`,
    },
  }
});
```
Query
```
{
  "measures": ["ExecutionLogs.latestRun"],
}
```
Response
```
{
  "query": {
    "measures": [
      "ExecutionLogs.latestRun"
    ],
    "filters": [],
    "timezone": "UTC",
    "dimensions": [],
    "timeDimensions": []
  },
  "data": [
    {
      "ExecutionLogs.latestRun": "2020-03-03T10:05:00.257Z"
    }
  ],
  "lastRefreshTime": "2020-03-06T14:44:47.795Z",
  "refreshKeyValues": [
    [
      {
        "max(execution_logs.startDate)": "2020-03-04T17:19:49.142Z"
      }
    ]
  ],
  "usedPreAggregations": {},
  "annotation": {
    "measures": {
      "ExecutionLogs.latestRun": {
        "title": "Execution Logs Latest Run",
        "shortTitle": "Latest Run",
        "type": "number"
      }
    },
    "dimensions": {},
    "segments": {},
    "timeDimensions": {}
  }
}
```

paveltiunov:
@richipargo Hey Ricardo! Thanks for posting this one! Yep. We should fallback to providing a string if measure value isn't a numeric.

paveltiunov:
For those who want to pickup and fix it here's the line of interest:
https://github.com/cube-js/cube.js/blob/80e74916b4dd4a0ca389ea87180cd89b5834616d/packages/cubejs-client-core/src/ResultSet.js#L295

PR #522, opened 2020-03-19
date parsing

richipargo:
**Check List**

Addresses #493 validating date fields before parsing


paveltiunov:
@richipargo Hey Ricardo! Looks great! Can we also check if value doesn't match any of regex and even not a string and return it as-is? So it won't fail with errors. Thanks!

richipargo:
@paveltiunov not the prettiest of solutions but it works welcome to any feedback

paveltiunov:
@richipargo Hey Ricardo! Looks good! Thanks for contributing this!

PR #531, opened 2020-03-23
fix(postgres-driver): Clean-up deprecation warning

jcw-:
**Check List**

When the pg driver is used with ssl set to true (e.g. `CUBEJS_DB_SSL=true`), the following deprecation warning is logged:
```
(node:23) DeprecationWarning: Implicit disabling of certificate verification is deprecated and will be removed in pg 8. Specify `rejectUnauthorized: true` to require a valid CA or `rejectUnauthorized: false` to explicitly opt out of MITM protection.
```

This PR updates how the ssl configuration options for the pg driver are set to address the deprecation.

Reference: 
https://github.com/GSG-G8/Books-store/issues/26
https://github.com/brianc/node-postgres/issues/2089


paveltiunov:
@jcw- Hey Jc! Thanks for providing this! Change looks good to me.  👍

jcw-:
Oh great, I haven't had a chance to verify fix but if you're able to do so then I shall submit it, thanks!

Issue #174, opened 2019-07-30
Support for ElasticSearch


AnandChowdhary:
**Is your feature request related to a problem? Please describe.**
Thanks for already great support with multiple DBMSes! I'd love to get support for ElasticSearch since it's perfect for analytics-related data and aggregation.

**Describe the solution you'd like**
Support for ElasticSearch as a database

paveltiunov:
@AnandChowdhary Hey Anand! Thanks for posting this! This driver actually was requested for 3 times already in https://slack.cube.dev. There's a good chance that someone is working on it already. Makes sense to double check that. 

Here's guide on how to implement driver: https://github.com/cube-js/cube.js/blob/master/CONTRIBUTING.md#implementing-driver

chad-codeworkshop:
@AnandChowdhary did PR #551 solve your issue?

AnandChowdhary:
Looks great, thanks!

Dineshagrawal85:
Hi, elasticsearch support added cubejs?

PR #551, opened 2020-03-30
Native ElasticSearch Driver

chad-codeworkshop:
**Check List**
- [ x] Tests has been run in packages where changes made if available
- [ x] Linter has been run for changed code

#174 

Updated Typings and QueryBuilder to include elasticsearchcloud
Implemented ElasticSearchCloudQuery.js for Native ElasticSearch dialect
Updated ElasticSearchDriver to include cloud config for elasticsearch cloud

Bug Fix on TS typings, renamed `member` to `dimension`

### DIFF
--- a/packages/cubejs-server-core/core/index.d.ts
+++ b/packages/cubejs-server-core/core/index.d.ts
@@ -100,7 +100,7 @@ declare module "@cubejs-backend/server-core" {
     | "sqlite";
 
   export interface QueryFilter {
-    dimension: string;
+    member: string;
     operator:
       | "equals"
       | "notEquals"
--- a/packages/cubejs-server-core/core/index.d.ts
+++ b/packages/cubejs-server-core/core/index.d.ts
@@ -92,7 +92,6 @@ declare module "@cubejs-backend/server-core" {
     | "mssql"
     | "mysql"
     | "odelasticsearch"
-    | "elasticsearchcloud"
     | "oracle"
     | "postgres"
     | "prestodb"
--- a/packages/cubejs-server-core/core/index.d.ts
+++ b/packages/cubejs-server-core/core/index.d.ts
@@ -92,6 +92,7 @@ declare module "@cubejs-backend/server-core" {
     | "mssql"
     | "mysql"
     | "odelasticsearch"
+    | "elasticsearchcloud"
     | "oracle"
     | "postgres"
     | "prestodb"
--- a/packages/cubejs-server-core/core/index.d.ts
+++ b/packages/cubejs-server-core/core/index.d.ts
@@ -101,7 +101,7 @@ declare module "@cubejs-backend/server-core" {
     | "sqlite";
 
   export interface QueryFilter {
-    member: string;
+    dimension: string;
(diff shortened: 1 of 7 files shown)


ifokeev:
@paveltiunov no styleguide = everybody style his code as he wants

chad-codeworkshop:
@ifokeev To be honest I was struggling with this as prettier attempts to reformat everything, I eventually just switched it off

ifokeev:
@chad-codeworkshop cubejs have to integrate eslint rules and maintainers need follow this rules too. That's the reason of my message

paveltiunov:
@chad-codeworkshop Hey Chad! Thanks for contributing it! Overall looks good to me! What's elastic search cloud driver? Is it X-Pack SQL or something else?

Also could you please rebase and revert all auto formatting changes? Those really hard to digest and make impossible to use git blame over time. Formatting of blocks you changed is totally ok as soon as it passes linter.

@ifokeev Hey Ivan! Thanks for heads up! We actually have style guide in place: https://github.com/cube-js/cube.js/blob/master/CONTRIBUTING.md#style-guides. Linter isn't part of CI build for all the packages right now however I just activated it for schema-compiler and it's enabled for elastic driver as well.

chad-codeworkshop:
@paveltiunov Hey Pavel, I currently work with the cloud hosted version of ElasticSearch provided by elastic themselves, there seems to be 3 different dialects, AWS-ES, Native ES and, OpenDistro

As you said in slack, the current implementation only works with AWS ES, I implemented the Dialect thats documented by ElasticSearch and the same version they run in their cloud offering

I also was not sure of the naming conventions here as they are going to get confusing in the future especially since the current one is `odelasticsearch` but only works with AWS-ES

I'll sort out the formatting changes today and update the pull request

paveltiunov:
@chad-codeworkshop Hey Chad! Cool! Looks great! I think it makes sense to rename dialect just to ElasticSearch as X-Pack SQL is default and native to elastic unless cloud dialect is different from it. `odelasticsearch` meant to be the single dialect shared between OpenDistro and AWS ES but turns out it isn't the case and seems we'd need to support those separately.

packages/cubejs-server-core/core/index.d.ts
```diff
@@ -100,7 +100,7 @@ declare module "@cubejs-backend/server-core" {
     | "sqlite";
 
   export interface QueryFilter {
-    member: string;
+    dimension: string;
```
paveltiunov, line 103: `member` and `dimension` are both allowed however `dimension` is deprecated here.

chad-codeworkshop:
@paveltiunov I have renamed everything to elasticsearch, If I get a bit of time I might look at the OpenDistro dialect, any recommendations for the naming there, since we already have odelasticsearch?

paveltiunov:
@chad-codeworkshop Hey Chad! Looks good to me! I believe we can move `odelasticsearch` to AWS one and create new `odelasticsearch`. It isn't backward compatible but currently driver itself just in preview so it makes sense to do this swap.

PR #557, opened 2020-03-31
add await when invoking schemaVersion

barakcoh:
**Check List**
- [X ] Tests has been run in packages where changes made if available
- [ X] Linter has been run for changed code

Add support for cases where `schemaVersion` needs to be `async`, e.g. fetching the version from remote storage

### DIFF
--- a/packages/cubejs-server-core/core/CompilerApi.js
+++ b/packages/cubejs-server-core/core/CompilerApi.js
@@ -18,7 +18,7 @@ class CompilerApi {
   async getCompilers(options) {
     const { requestId } = options || {};
     let compilerVersion = (
-      this.schemaVersion && this.schemaVersion() ||
+      this.schemaVersion && await this.schemaVersion() ||
       'default_schema_version'
     ).toString();
     if (this.options.devServer) {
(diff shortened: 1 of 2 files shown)


paveltiunov:
@barakcoh Hey Barak! Just briefly checked code usage and it seems it probably not so easy as I thought before. `schemaVersion` property set in `CubejsServerCore.getCompilerApi()` and it isn't async. Good new is probably we can make it async.

barakcoh:
hi @paveltiunov ! should I just add another `await` at https://github.com/cube-js/cube.js/blob/af07865ae2d3c11836ab5e01f17e97c3b2d5d662/packages/cubejs-server-core/core/index.js#L383?

paveltiunov:
@barakcoh Yeah. Right. And make it `async`.

paveltiunov:
@barakcoh Just checked usages more thoroughly and it seems we're safe to revert all changes to index.js as only `await` is in fact required. So your initial change was correct.

barakcoh:
thanks @paveltiunov i'll push a reverse commit

PR #582, opened 2020-04-09
Allow usage of CUBEJS_DB_SSL_CA parameter in postgres Driver.

### LINKED ISSUE #579: Support usage of CUBEJS_SSL_CA parameter for postgres Driver
Amazon RDS has released its 2019-2024 CA’s. The users must upgrade their CA’s to the new one before February 5th 2020. This is easily possilbe by setting the CA property in the SSL options of the database connection. Unfortunately the CUBEJS_SSL_CA parameter is not utilized in the postgres Driver.

All the Environment variable regarding SSL connections should be utilized in the Driver.

Currently I am using a local modified copy of the driver passed in to the DriverFactory which works well.

gauravlanjekar:
**Check List**

[fixes #579]

Copied over SSL code from MongoBI Driver. Assumed CUBEJS_DB_SSL_CA will be a filename. Updated the docs with this info.

### DIFF
--- a/docs/Cube.js-Backend/Connecting-to-the-Database.md
+++ b/docs/Cube.js-Backend/Connecting-to-the-Database.md
@@ -63,9 +63,9 @@ here.](https://cube.dev/blog/building-mongodb-dashboard-using-node.js)
 
 Use `CUBEJS_DB_SSL=true` to enable SSL as MongoDB Atlas requires it. `CUBEJS_DB_SSL_CA`, `CUBEJS_DB_SSL_CERT`, `CUBEJS_DB_SSL_CIPHERS`, `CUBEJS_DB_SSL_PASSPHRASE` can be left blank.
 
-### AWS RDS 
+### AWS RDS Postgres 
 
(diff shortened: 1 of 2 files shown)


paveltiunov:
@gauravlanjekar Hey Gaurav! Overall looks good to me. Can we remove `readFileSync()` and just stick with same approach used in `MongoBIDriver`? Seems to be more generic to me. Also could you please fix linter errors? Thanks!

paveltiunov:
@gauravlanjekar Hey Gaurav! Looks great! Thanks for providing this!

jcw-:
@gauravlanjekar 
> Assumed CUBEJS_DB_SSL_CA will be a filename

how is the new CUBEJS_DB_SSL_CA intended to be used? it doesn’t seem to work if I put in the relative path to the .pem file, is it intended to contain the actual file contents?

gauravlanjekar:
Hi @jcw- The idea before was to use this parameter as a file name. but to keep the consistency with the Mongo driver we decided to have it as actual contents of the file. 
@jcw- may be you could create a separate issue to support file based certificate parameter
@paveltiunov what do you think ?

paveltiunov:
@gauravlanjekar @jcw- Yeah. I think it makes sense to extract SSL code to BaseDriver and support file env variables as well in centralized manner.

jcw-:
Sounds good, I'll make a separate issue - Heroku limits the total bytes of env var key-value-pairs and putting the pem in this env var exceeds the total size allowed, so the filename approach will provide Heroku support.

PR #615, opened 2020-04-24
feat: Include version in startup message

jcw-:
**Check List**

Includes the Cube.js version in the startup message:
image

The version is available in three places:

1. `CubejsServer.version()`
1. `CubejsServerCore.version()`
1. `const { version, port } = await cubejsServer.listen();`

### DIFF
--- a/docs/Cube.js-Backend/@cubejs-backend-server-core.md
+++ b/docs/Cube.js-Backend/@cubejs-backend-server-core.md
@@ -195,7 +195,6 @@ CubejsServerCore.create({
 })
 ```
 
-
 ### contextToAppId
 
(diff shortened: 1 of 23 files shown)


jcw-:
@paveltiunov curious what you think about doing something like this so the version is more discoverable (and also logged)? If this makes sense then I'll proceed to update the docs, examples, and guide content with the same change.

paveltiunov:
@jcw- Hey Jc! I think it's good idea! It would be great to move this version retrieval to method of server instance.

paveltiunov:
@jcw- Hey Jc! Looks good! We'd need just to resolve conflicts before merge.

jcw-:
@paveltiunov conflicts resolved!

PR #369, opened 2020-02-10
Add more Typings

lvauvillier:
* Improve core typings
* Add biquery-driver typings
* Add sqlite-driver typings
* Add postgres-driver typings

PR #619, opened 2020-04-27
feat: Update type defs for query transformer

jcw-:
**Check List**

This PR builds on #369 and moves the Query type to be owned by the package that is responsible for it - `api-gateway`. Query now means the incoming query - nothing is actually using this at this point, although methods like `load` could be declared to use it.

Internally, `api-gateway` "normalizes" the incoming query and adds a few things to the structure - this is represented by the new `NormalizedQuery` and `NormalizedQueryFilter` types. Since it is actually the `NormalizedQuery` that is made available to query transformers, I was running into a problem where TypeScript was complaining about the use of `dimension`, which it thought wasn't available but it actually is - this will now work properly.

### DIFF
--- a/packages/cubejs-api-gateway/index.d.ts
+++ b/packages/cubejs-api-gateway/index.d.ts
@@ -1,0 +1,57 @@
+declare module "@cubejs-backend/api-gateway" {
+  export interface QueryFilter {
+    member: string;
+    operator:
+      | "equals"
+      | "notEquals"
+      | "contains"
+      | "notContains"
+      | "gt"
+      | "gte"
+      | "lt"
+      | "lte"
+      | "set"
+      | "notSet"
+      | "inDateRange"
+      | "notInDateRange"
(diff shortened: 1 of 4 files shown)


packages/cubejs-api-gateway/index.d.ts
```diff
@@ -0,0 +1,57 @@
...
+    ungrouped?: boolean;
+  }
+
+  export interface NormalizedQuery extends Query {
+    filters?: NormalizedQueryFilter[];
+    rowLimit?: number;
+  }
+
+  export interface NormalizedQueryFilter extends QueryFilter {
+    dimension: string;
```
paveltiunov, line 55: `dimension` still accepted here but `member` more recent name for this field.

jcw-: @paveltiunov oh good to know - i've now marked it as optional

PR #623, opened 2020-04-28
feat: Update server type definitions

willhausman:
Core server create options should be optional.  Added missing options parameter for CubejsServer.


paveltiunov:
@willhausman Hey Will! Looks great! Thanks for contributing this!

PR #631, opened 2020-05-04
feat(@cubejs-backend/schema-compiler): Add index clustering, compression

ack210:
**Check List**

Currently, with respect to `mssql`, pre-aggregations are stored in [heap tables](https://docs.microsoft.com/en-us/sql/relational-databases/indexes/heaps-tables-without-clustered-indexes?view=sql-server-ver15) (`SELECT [...] INTO`), which for most cases will have a negative impact on the performance of queries referencing those tables, particularly those involving filtering by a range of values or grouping.  In these cases, it is recommended to add a `CLUSTERED` index to the table, e.g. `CREATE CLUSTERED INDEX ON orders(id)`.

`mssql` also offers some additional features that may help query performance:
* [`row`- and `page`-level compression](https://docs.microsoft.com/en-us/sql/relational-databases/data-compression/data-compression?view=sql-server-ver15) for rowstore indexes, which reduces the size of the index on disk, and thus the disk io operations that must be performed when accessing data within the index.
* The `unique` keyword, which indicates that the index key comprises unique values, [helping the query planner come up with better estimates](https://www.brentozar.com/archive/2015/08/performance-benefits-of-unique-indexes/)
* `columnstore` compression, which can greatly improve query performance (up to 100x according to MS) on large tables for [a variety of reasons](https://docs.microsoft.com/en-us/sql/relational-databases/indexes/columnstore-indexes-query-performance?view=sql-server-ver15)
* [included columns](https://docs.microsoft.com/en-us/sql/relational-databases/indexes/create-indexes-with-included-columns?view=sql-server-ver15) via the `INCLUDE` keyword, which can improve query performance by increasing the chances that the index covers the query, thus eliminating the need for including non-key columns in the index key or requiring an additional lookup on the underlying table to retrieve values from columns missing in the index.

This PR allows these index options to be specified via properties of the `indexes` object in a pre-aggregation:

```javascript
{
    include: ['column_a', 'column_b', ...],
    clustered: boolean,
    unique: boolean,
    compression: 'row' | 'page' | 'columnstore'
}
```

For example:
```javascript
cube(`Orders`, {
  sql: `select * from orders where completed = true`,

  preAggregations: {
    main: {
      type: `originalSql`,
      indexes: {
        clustered: {
          clustered: true,
          compression: 'columnstore',
        },
        orders: {
          columns: ['id'],
          include: ['timestamp']
          unique: true,
          compression: 'row'
        }
      }
    }
  }
});
```
which would (approximately) result in the following SQL statements:
```tsql
CREATE CLUSTERED COLUMNSTORE INDEX orders_clustered ON orders;
CREATE UNIQUE INDEX orders_orders ON orders (id) INCLUDE(timestamp) WITH(DATA_COMPRESSION = ROW);
```

**postgres**
Postgres supports the `UNIQUE` [keyword](https://www.postgresql.org/docs/9.6/indexes-unique.html) for indexes.

As of version 11, Postgres also supports the [INCLUDE clause](https://www.postgresql.org/docs/11/sql-createindex.html) for indexes, however I did not implement this since the postgres containers in the package's tests are currently using 9.6.

### DIFF
--- a/docs/Schema/pre-aggregations.md
+++ b/docs/Schema/pre-aggregations.md
@@ -489,3 +489,39 @@ cube(`Orders`, {
   }
 });
 ```
+
+Depending on your underlying database, you can specify additional index options for improved performance, where appropriate.
+Currently, supported options are:
+ * **include** (mssql) - Provide additional columns to be included in the index that are not part of the index's key.  This can improve performance by allowing the index to
+cover more queries, thus reducing the need to retrieve values from missing columns from the underlying table or clustered index.  Not permitted when `clustered` is specified, or when
+`columnstore` compression is used.
+ * **clustered** (mssql) - When `true`, indicates that the index should be used as the clustering index for the table.  Only one clustering index may be specified per table.
+ * **unique** (mssql) - When `true`, indicates that values in the key columns of the index are unique, which can help the query compiler come up with better estimates when the index is used.
+ * **compression** (mssql) - Supported values are `page`, `row`, or `columnstore`.  `row` and `page` compression levels can improve performance by reducing the amount of disk io.
+In addition to reducing the size of the table, `columnstore` compression unlocks [additional query performance improvements](https://docs.microsoft.com/en-us/sql/relational-databases/indexes/columnstore-indexes-query-performance?view=sql-server-ver15) such as column and rowgroup elimination, batch mode execution, and aggregate and predicate pushdown.
+
+Note:  `clustered` indexes using `columnstore` compression may not be `unique`, and may not specify any `columns` or `include` columns.
+
+```javascript
+cube(`Orders`, {
+  sql: `select * from orders where completed = true`,
+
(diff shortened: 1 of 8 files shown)


paveltiunov:
@ack210 Hey Andrew! Wow! Thanks for contributing this! Looks really great! The only thing I think `compression` and `clustered` cases should go to `sql` instead of dedicated fields. Cube.js data schema approach is to support everything portable across most of databases and leave all the rest for `sql` fields. While `compression` in general is supported by most of databases in some way it's still unclear to me how can we make this property portable. What do you think?

Also for `sql` field to reuse columns definitions and aliases we can introduce `COLUMNS` context variable.

rpaik:
👋 @ack210, I'll go ahead and close this since this has been inactive for almost a year. When you're ready to pick this up again, please feel free to re-open. 🙏

Issue #651, opened 2020-05-15
PostgresDriver execution timeout is hardcoded to 10 minutes


vjfuenzalida:
**Describe the bug**
The `executionTimeout` parameter exposed in `queueOptions` (to create a `CubejsServer`) doesn't work when using an external Driver (for the external rollouts feature) such as Postgresql.

This parameter is currently hardcoded to `600000` milliseconds (10 minutes) in the `PostgresDriver` class definition, line 85:

```javascript
async queryResponse(query, values) {
    const client = await this.pool.connect();
    try {
      await client.query(`SET TIME ZONE '${this.config.storeTimezone || 'UTC'}'`);
      await client.query("set statement_timeout to 600000");
      const res = await client.query({
        text: query,
        values: values || [],
        types: {
          getTypeParser: (dataType, format) => {
            const isTimestamp = timestampDataTypes.indexOf(dataType) > -1;
            let parser = types.getTypeParser(dataType, format);

            if (isTimestamp) {
              parser = timestampTypeParser;
            }

            return val => parser(val);
          },
        },
      });
      return res;
    } finally {
      await client.release();
    }
  }
```

**To Reproduce**
1. Setup Cubejs server using PostgresDriver.
2. Pass executionTimeout setting in queueOptions with any value T.
3. Submit a long duration queue and see that it will be terminated exactly at 10 minutes, and not at time T.

**Expected behavior**
The executionTimeout should be used in the PostgresDriver class definition when calling `set statement_timeout` in query submission.

**Version:**
node: 12.16.1
npm: 6.13.4
@cubejs-backend/postgres-driver: 0.19.17
@cubejs-backend/server: 0.19.19

ovr:
I am going to close it because It was resolved by https://github.com/cube-js/cube.js/pull/652

PR #652, opened 2020-05-16
fix:PostgresDriver execution timeout is hardcoded to 10 minutes #651

cd-romulus:
**Check List**

#651 

[Description goes here]
executionTimeout (in seconds) can be added to the driver config argument and will be used as statetment_timeout for postgres. config.executionTimeout=0 will, set statemtent_timeout to infinite. if executionTimeout is undefined, the original default value of 10 min will be used.


paveltiunov:
Hey @cd-romulus ! Looks great! I think we can merge it as workaround for now however right fix is to get rid of it and implement `cancel()` logic. cc @vasilev-alex

jmadan:
is this released? I still see the postgres driver terminates the connection around 7.5 minutes consistently.

PR #659, opened 2020-05-23
Fix sqlite name and type extraction

### LINKED ISSUE #626: Sqlite Driver cannot create schema
I'm trying to connect a sqlite3 database but the driver was returning an error:

```
TypeError: Cannot read property '2' of null
    at ./node_modules/@cubejs-backend/sqlite-driver/driver/SqliteDriver.js:63:33
```

1. Create a new project based on the sqlite driver.
2. Create new table in sqlite `CREATE TABLE test (id TEXT PRIMARY KEY, name TEXT NOT NULL);`
3. Launch Cube with `node index.js`.
4. Try to build a new schema for the test table.

I expect that it should be able to create the schema.

**Screenshots**

**Version:**
0.19.14

I was able to make it work by changing the regex matching in the sqlite driver to `      .match(/(\w+)\s(\w+)/);` but I know that's not going to work for every case.

avin3sh:
Resolves #626 

* Fixed regex to extract field names and types
* Added condition to ignore `sqlite_sequence` table created for maintaining autoincrement index in `informationSchemaQuery`

### DIFF
--- a/packages/cubejs-sqlite-driver/driver/SqliteDriver.js
+++ b/packages/cubejs-sqlite-driver/driver/SqliteDriver.js
@@ -59,7 +59,7 @@ class SqliteDriver extends BaseDriver {
           .map((nameAndType) => {
             const match = nameAndType
               .trim()
-              // replace \t with whitespace 
+              // replace \t with whitespace
               .replace(/\t/g, ' ')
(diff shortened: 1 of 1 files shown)


paveltiunov:
@avin3sh Hey Avinesh! Looks good! As it getting more complex I believe it would be great to add a test for this case.

PR #868, opened 2020-07-19
fix: allow renewQuery in dev mode with warning

jcw-:
**Check List**

The combination of `renewQuery` on a query against a server running in `externalRefresh` mode results in a validation error (and correctly so - you can't renew queries against a server running in that mode), but it is breaking the dev playground's Cache tab.

This PR improves the guards for missing data in the Cache section, and also makes a change to allow `renewQuery` to be specified in dev mode, except it will issue a warning and force it back to false.


paveltiunov:
@jcw- Hey Jc! Looks great! Thanks for fixing it!

PR #913, opened 2020-07-29
fix: CubeContext ts type missing

jd2rogers2:
**Check List**
  -- well an example comment anyways

fixes https://github.com/cube-js/cube.js/issues/912

add export of CubeContext type to index.d.ts
usage example comment

### DIFF
--- a/packages/cubejs-client-react/index.d.ts
+++ b/packages/cubejs-client-react/index.d.ts
@@ -44,6 +44,8 @@ declare module '@cubejs-client/react' {
    */
   export const CubeProvider: React.FC<CubeProviderVariables>;
 
+  export const CubeContext: Context<{ cubejsApi: CubejsApi }>;
+
   type TLoadingState = {
     isLoading: boolean;
   };
--- a/packages/cubejs-client-react/index.d.ts
+++ b/packages/cubejs-client-react/index.d.ts
@@ -44,22 +44,34 @@ declare module '@cubejs-client/react' {
    */
   export const CubeProvider: React.FC<CubeProviderVariables>;
 
+  type CubeContextVariables = {
+    cubejsApi: CubejsApi;
+  };
+
   /**
    * Cube.js context consumer
+   * In case when you need access to cubejsApi directly you can use CubeContext anywhere in your app
+   *
    * ```js
    * import React from 'react';
    * import { CubeContext } from '@cubejs-client/react';
    *
    *
    * export default function DisplayComponent() {
-   * const { cubejsApi } = React.useContext(CubeContext);
-   * const query = {
-   *   ...
-   * };
-   * const resultSet = await cubejsApi.load(query);
+   *   const { cubejsApi } = React.useContext(CubeContext);
+   *   const [rawResults, setRawResults] = React.useState([]);
+   *   const query = {
+   *     ...
+   *   };
+   *
+   *   React.useEffect(() => {
+   *     const resultSet = await cubejsApi.load(query);
+   *     setRawResults(resultSet.rawData());
+   *   }, [query]);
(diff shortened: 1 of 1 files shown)


packages/cubejs-client-react/index.d.ts
```diff
@@ -44,6 +44,31 @@ declare module '@cubejs-client/react' {
    */
   export const CubeProvider: React.FC<CubeProviderVariables>;
 
+  /**
+   * Cube.js context consumer
```
vasilev-alex, line 48: Would be better to add a bit more descriptive comment (eg "In case when you need access to `cubejsApi` directly you can use `CubeContext` anywhere in your app")

packages/cubejs-client-react/index.d.ts
```diff
@@ -44,6 +44,31 @@ declare module '@cubejs-client/react' {
...
+   * import React from 'react';
+   * import { CubeContext } from '@cubejs-client/react';
+   *
+   *
+   * export default function DisplayComponent() {
+   * const { cubejsApi } = React.useContext(CubeContext);
+   * const query = {
+   *   ...
+   * };
+   * const resultSet = await cubejsApi.load(query);
```
vasilev-alex, line 59: You can't really use `await` directly in components. You should use `useEffect`

packages/cubejs-client-react/index.d.ts
```diff
@@ -44,6 +44,31 @@ declare module '@cubejs-client/react' {
...
+   *     <>
+   *       {resultSet.rawData().map(row => (
+   *         ...
+   *       ))}
+   *     </>
+   *   )
+   * }
+   * ```
+   */
+  export const CubeContext: Context<{ cubejsApi: CubejsApi }>;
```
vasilev-alex, line 70: Better to factor the `type` out and refer to `Context` using `React.Context` (as all these types will go directly to our docs)

e.g.

```js
export const CubeContext: React.Context<CubeContextProps>
```

packages/cubejs-client-react/index.d.ts
jd2rogers2: added

packages/cubejs-client-react/index.d.ts
jd2rogers2: ah good catch, had over simplified
added useEffect

packages/cubejs-client-react/index.d.ts
jd2rogers2: followed `CubeProviderVariables` naming convention from above
so added `CubeContextVariables`

packages/cubejs-client-react/index.d.ts
```diff
@@ -44,6 +44,43 @@ declare module '@cubejs-client/react' {
...
+   *
+   *
+   * export default function DisplayComponent() {
+   *   const { cubejsApi } = React.useContext(CubeContext);
+   *   const [rawResults, setRawResults] = React.useState([]);
+   *   const query = {
+   *     ...
+   *   };
+   *
+   *   React.useEffect(() => {
```
vasilev-alex, line 67: Almost here :) 

```js
React.useEffect(async () => {
```

packages/cubejs-client-react/index.d.ts
```diff
@@ -44,6 +44,43 @@ declare module '@cubejs-client/react' {
...
+   *     const resultSet = await cubejsApi.load(query);
```
vasilev-alex, line 68: Almost there :) 

But `useEffect` can't be async on it own. I guess simple `.then` would work here

```js
cubejsApi.load(query).then((resultSet) => ... )
```

PR #915, opened 2020-07-29
feat: Make the Filter type more specific.

ylixir:
**Check List**
- [ x ] Tests has been run in packages where changes made if available
- [ x ] Linter has been run for changed code
- [ x ] Tests for the changes have been added if not covered yet
- [ x ] Docs have been added / updated if required

none

Now the type more accurately represents which Filter objects are valid.
This will provide better code completion, documentation and allow the
typechecker to better help in writing more robust new code and in more
easily refactoring existing code.

This is just encoding existing docs in the type system, so no docs are
needed. This also does not add or change any existing behavior, so
tests are not needed.

This has the potential to cause new errors at compile time, but can only
erase runtime errors. I'm not sure what the tolerance for that is with
this project. Presumably if one is using typescript in the first place, then
one wants better data models?

### DIFF
--- a/packages/cubejs-client-core/index.d.ts
+++ b/packages/cubejs-client-core/index.d.ts
@@ -563,12 +563,35 @@ declare module '@cubejs-client/core' {
     annotation(): QueryAnnotations;
   }
 
-  export type Filter = {
+  export type Filter = BinaryFilter | UnaryFilter;
+  type BinaryFilter = {
     dimension?: string;
     member?: string;
-    operator: string;
-    values?: string[];
+    operator: BinaryOperator;
+    values: string[];
   };
+  type UnaryFilter = {
+    dimension?: string;
+    member?: string;
+    operator: UnaryOperator;
+    values?: never;
+  };
+  type UnaryOperator = 'set' | 'notSet';
+  type BinaryOperator =
+    | 'equals'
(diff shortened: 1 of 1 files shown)


packages/cubejs-client-core/index.d.ts
```diff
@@ -563,12 +563,36 @@ declare module '@cubejs-client/core' {
...
+    operator: BinaryOperator;
+    values: string[];
   };
+  type UnaryFilter = {
+    dimension?: string;
+    member?: string;
+    operator: UnaryOperator;
+    values?: never;
+  };
+  type FilterOperator = UnaryOperator | BinaryOperator;
```
vasilev-alex, line 579: Looks like this type is redundant

ylixir: Oops. I have removed it now.

ylixir:
I'm not sure what happened to the tests. They were working and now they aren't. An external issue with circle-ci maybe?

vasilev-alex:
@ylixir yeah, we had some trouble with it. Could you please pull from master? It should be fixed there.

ylixir:
> @ylixir yeah, we had some trouble with it. Could you please pull from master? It should be fixed there.

@vasilev-alex done, sorry for the delay

PR #920, opened 2020-07-30
feat(mssql-driver): add readonly aggregation for mssql sources

JoshMentzer:

Updating MSSqlDriver to implement downloadQueryResults function, enabling ability to have source systems use pre-aggregation where readOnly is true and externalDb utilized.

### DIFF
--- a/packages/cubejs-mssql-driver/driver/MSSqlDriver.js
+++ b/packages/cubejs-mssql-driver/driver/MSSqlDriver.js
@@ -14,7 +14,8 @@ class MSSqlDriver extends BaseDriver {
         process.env.CUBEJS_DB_DOMAIN : undefined,
       requestTimeout: 10 * 60 * 1000, // 10 minutes
       options: {
-        encrypt: !!process.env.CUBEJS_DB_SSL || false
+        encrypt: !!process.env.CUBEJS_DB_SSL || false,
+        useUTC: false
       },
       pool: {
         max: 8,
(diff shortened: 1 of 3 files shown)


jcw-:
@JoshMentzer Nice! I think you'll also need to implement `readOnly` on the driver, otherwise the base implementation that always returns `false` will be used - you can see #395 for an example of when this functionality was added to the Postgres driver.

JoshMentzer:
@jcw- Got it, thx!

jcw-:
@JoshMentzer what if you define your time dimension columns with explict SQL that does a conversion? https://cube.dev/docs/working-with-string-time-dimensions#top

JoshMentzer:
> @JoshMentzer what if you define your time dimension columns with explict SQL that does a conversion? https://cube.dev/docs/working-with-string-time-dimensions#top

Sorry, I should have been more explicit - it's not the timeDimensions themselves, it's the timeDimension with granularity; the timeDimensions are datetime; so let's say there is one called closedAt - in the query for preaggregation it's coming back as closedAt_day.  That is the column that is being typed as nvarchar

jcw-:
@JoshMentzer Oh I see - looks like that driver uses FORMAT for time dimension granularity, which turns it into a string: 

https://github.com/cube-js/cube.js/blob/846a519c99167459e8d3513150ebaaf0e11eb929/packages/cubejs-schema-compiler/adapter/MssqlQuery.js#L31-L39

While postgres uses `date_trunc` - looks like the `GRANULARITY_TO_INTERVAL` for mssql would be a good place to start!

JoshMentzer:
@jcw- I think this is now good - thanks for the assist!

JoshMentzer:
Is there something special to do to get the test-mssql to run locally?  I get 'ConnectionError: Login failed for user 'sa'' when attempting to run.  I have docker, and am able to run 'test' in the schema-compilor just not test-mssql

jcw-:
@JoshMentzer I'm not sure - but when I run tests for say, the query-orchestrator, I use this command, maybe it will help:
```
yarn lerna run --concurrency 1 --ignore @cubejs-backend/jdbc-driver --scope @cubejs-backend/query-orchestrator test
```

paveltiunov:
@JoshMentzer Hey Joshua! Looks good to me! Thanks for contributing it!

PR #929, opened 2020-07-31
fix: avoid querying source database when caching tables from external rollup db

jcw-:
**Check List**

Currently,  table caching will cause a test query to be made to BOTH the source and the rollup databases. This PR prevents a connection and query from being ran against the source database when caching the tables in the external rollup database. This is particularly important when `externalRefresh` is being used, as in that architecture the cubejs API instance likely isn't configured with the capability to connect to the source database.

The reason this happens is because `QueryCache.createQueue` will always setup a driver instance for the source database:
https://github.com/cube-js/cube.js/blob/c30a44c26e76e83c3d26234850d5c6b0e76cd924/packages/cubejs-query-orchestrator/orchestrator/QueryCache.js#L190

This normally wouldn't create a connection or run a query, except that there is additional logic in place when creating a driver instance that when it is created for the first time, it will run a query in order to validate that it is configured correctly (`testConnection`):
https://github.com/cube-js/cube.js/blob/c30a44c26e76e83c3d26234850d5c6b0e76cd924/packages/cubejs-server-core/core/index.js#L462-L473

### DIFF
--- a/packages/cubejs-query-orchestrator/orchestrator/PreAggregations.js
+++ b/packages/cubejs-query-orchestrator/orchestrator/PreAggregations.js
@@ -714,7 +714,7 @@ class PreAggregations {
 
   getLoadCacheQueue() {
     if (!this.loadCacheQueue) {
-      this.loadCacheQueue = QueryCache.createQueue(`SQL_PRE_AGGREGATIONS_CACHE_${this.redisPrefix}`, this.driverFactory, (client, q) => {
+      this.loadCacheQueue = QueryCache.createQueue(`SQL_PRE_AGGREGATIONS_CACHE_${this.redisPrefix}`, () => {}, (_, q) => {
         const {
           preAggregation,
           requestId
--- a/packages/cubejs-query-orchestrator/test/PreAggregations.test.js
+++ b/packages/cubejs-query-orchestrator/test/PreAggregations.test.js
@@ -1,5 +1,5 @@
 /* globals describe, beforeAll, afterAll, beforeEach, test, expect */
-const crypto = require('crypto');
+const R = require('ramda');
 
 class MockDriver {
   constructor() {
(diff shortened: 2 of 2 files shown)


paveltiunov:
@jcw- Hey Jc! I'm open to merge this however I believe it doesn't solve original issue: selection of incorrect DB to fetch tables. Your case should be probably already covered here: https://github.com/cube-js/cube.js/blob/dbdbb5ff259ee4d4e6c229eeb28bc7016e01ecbb/packages/cubejs-query-orchestrator/orchestrator/PreAggregations.js#L83.

This is not really obvious one so I'd suggest to add test that fails with a current code base and fixed by a code change to make sure we're fixing the right use case.

jcw-:
@paveltiunov Sure I can add tests - the place you referenced is using the correct driver, but it is only determining the SQL to use at that point. Later it executes that SQL via the query queue which does not do that same check, resulting in the right query, but against the wrong database.

paveltiunov:
@jcw- Sounds great! I think test will help to resolve this a lot. Also just checked this line again: the next line executes query: https://github.com/cube-js/cube.js/blob/dbdbb5ff259ee4d4e6c229eeb28bc7016e01ecbb/packages/cubejs-query-orchestrator/orchestrator/PreAggregations.js#L86. Naming a little bit misleading here or probably you're referring to some other place.

jcw-:
@paveltiunov ah yes - sorry, yes I am talking about a different code path:

loadPreAggregation -> getVersionEntries -> getTablesQuery:
https://github.com/cube-js/cube.js/blob/fb57bda0a2ac4c75d840fe3e29ddb21caba7eb3e/packages/cubejs-query-orchestrator/orchestrator/PreAggregations.js#L110

In that flow, the problem I describe surfaces - I'll see if I can capture it in a test momentarily...

jcw-:
@paveltiunov You were right, I was missing the root cause of what I was seeing - I dug further and it as you pointed out, the right driver IS being used for caching tables.

However, that code path is ALSO establishing a connection to the source database and executing `testConnection` which runs a test query, even though that driver isn't used or needed in all cases:
https://github.com/cube-js/cube.js/blob/c30a44c26e76e83c3d26234850d5c6b0e76cd924/packages/cubejs-query-orchestrator/orchestrator/QueryCache.js#L190

Since the table caching code doesn't actually leverage the driver instance that `createQueue` is setting up for convenience, this PR gives `createQueue` an empty driver factory method in that flow.

I have added a test that not only covers this case (and fail on master) but also some other tests for aggregation flows I have been wanting to get some test coverage on. Let me know if this PR makes more sense now!

paveltiunov:
@jcw- Gotcha. Now it's clear what's happening. Thanks for tracking this down and fixing it!

PR #972, opened 2020-08-09
Resolve #962: fix diff date format between filters and timeDimension

### LINKED ISSUE #962: Specifying `dateRange` in time dimension should produce same result as `inDateRange` in filter
## Describe the bug

When querying a cube with a time dimension, the behavior is different when specifying a filter on the time dimension via the `dateRange` option of the time dimension, vs specifying the same constraint as a `filter` with an `inDateRange` operator.

In the case of a `filter` with an `inDateRange` operator, the upper endpoint of the date range is directly cast to a `timestamptz` (with a PostgreSQL backend); e.g.,

```
{"query":{..., "filters":[{"dimension":"...","operator":"inDateRange","values":["2020-08-01","2020-08-10"]}], ...}
```

Results in:

```
{"query":"... WHERE (... >= $1::timestamptz AND ... <= $2::timestamptz)","params":["2020-08-01","2020-08-10"],...}
```

On the other hand, when using a time dimension with a `dateRange` specification, the date value provided in the API request is converted to the final second of the given day of the upper end of the date range; e.g.,

```
{"query":{"measures":["..."],"order":{},"dimensions":[],"filters":[],"timeDimensions":[{"dimension":"...","granularity":"day","dateRange":["2020-08-01","2020-08-10"]}]}
```

Results in:

```
{"query":"... WHERE (... >= $1::timestamptz AND ... <= $2::timestamptz)","params":["2020-08-01T00:00:00Z","2020-08-10T23:59:59Z"], ...}
```

## Expected behavior

Both `inDateRange` on `filter` and `dateRange` on `timeDimension` would behave the same for time dimensions.

## Version:

0.19.55

RusovDmitriy:
**Check List**

Specifying `dateRange` in time dimension should produce same result as `inDateRange` in filter #962

I found 2 problems:

1. **BaseFilter** created with operator in **CamelCase**, but array DATE_OPERATORS includes **snake_case** operators only
Move to https://github.com/cube-js/cube.js/pull/999

2. In **BaseQuery** first allocated to paramAllocator with date formated for timeSeries array, but is different date format
Resolved in this PR

### DIFF
--- a/packages/cubejs-schema-compiler/adapter/BaseFilter.js
+++ b/packages/cubejs-schema-compiler/adapter/BaseFilter.js
@@ -79,8 +79,15 @@ class BaseFilter extends BaseDimension {
     return this.operator === 'contains' || this.operator === 'not_contains';
   }
 
+  isDateOperator() {
+    // Fix camelCase => snake_case
+    const operator = this.operator.replace(/[A-Z]/g, letter => `_${letter.toLowerCase()}`);
+
+    return contains(operator, DATE_OPERATORS);
+  }
+
   filterParams() {
-    if (contains(this.operator, DATE_OPERATORS)) {
+    if (this.isDateOperator()) {
       return [this.inDbTimeZoneDateFrom(this.values[0]), this.inDbTimeZoneDateTo(this.values[1])];
     }
     if (this.operator === 'set' || this.operator === 'not_set' || this.operator === 'expressionEquals') {
--- a/packages/cubejs-schema-compiler/adapter/BaseFilter.js
+++ b/packages/cubejs-schema-compiler/adapter/BaseFilter.js
@@ -79,15 +79,8 @@ class BaseFilter extends BaseDimension {
     return this.operator === 'contains' || this.operator === 'not_contains';
   }
 
-  isDateOperator() {
-    // Fix camelCase => snake_case
-    const operator = this.operator.replace(/[A-Z]/g, letter => `_${letter.toLowerCase()}`);
-
-    return contains(operator, DATE_OPERATORS);
-  }
-
   filterParams() {
-    if (this.isDateOperator()) {
+    if (contains(this.operator, DATE_OPERATORS)) {
       return [this.inDbTimeZoneDateFrom(this.values[0]), this.inDbTimeZoneDateTo(this.values[1])];
     }
     if (this.operator === 'set' || this.operator === 'not_set' || this.operator === 'expressionEquals') {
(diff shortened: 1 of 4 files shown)


packages/cubejs-schema-compiler/adapter/BaseFilter.js
```diff
@@ -79,8 +79,15 @@ class BaseFilter extends BaseDimension {
...
+  isDateOperator() {
+    // Fix camelCase => snake_case
+    const operator = this.operator.replace(/[A-Z]/g, letter => `_${letter.toLowerCase()}`);
+
+    return contains(operator, DATE_OPERATORS);
+  }
+
   filterParams() {
-    if (contains(this.operator, DATE_OPERATORS)) {
+    if (this.isDateOperator()) {
```
paveltiunov, line 90: This place is handled in api-gateway. To make this change I'd suggest separate PR and update all the tests so we can get rid of the snake case.

packages/cubejs-schema-compiler/adapter/BaseQuery.js
```diff
@@ -592,10 +592,10 @@ class BaseQuery {
             const timeSeries = d.timeSeries();
             return f(
               isFromStartToEnd ?
-                this.dateTimeCast(this.paramAllocator.allocateParam(timeSeries[0][0])) :
```
paveltiunov, line 595: It seems isn't equivalent to `d.dateFrom()`.

packages/cubejs-schema-compiler/adapter/BaseFilter.js
RusovDmitriy: It's really better to separate, please review: https://github.com/cube-js/cube.js/pull/999

packages/cubejs-schema-compiler/adapter/BaseQuery.js
RusovDmitriy: Deepened into the code and found the reason for the differences values for date filter: 
- queries with timeSeries generated with timezone in SQL
- queries with filter by inDateRage generated date values with TZ offset 

Now I understand that the code works correctly, i revert changes and added some tests, maybe they'll be useful ))

paveltiunov: @RusovDmitriy Makes sense. Do you want to use this PR to fix the original issue or are you thinking about starting the new one?

RusovDmitriy: @paveltiunov In original issue was about different date format between dateRange and timeDimension, but i think is over anymore:

1. For filter with date operators, after fix https://github.com/cube-js/cube.js/pull/999 we'll get the date format like `2017-01-01T08:00:00Z`, with timezone offset

2. For timeDimension don't need any fix... yes, values have diff format like `YYYY-MM-DDT00:00:00.000`, but timeDimension build SQL with diff code and have special logic for `granularity`.

So... if you have any idea - i accept advice))

PS: I think deep down the problem is that there are no contracts in your code, DRY principles are being violated... api-gateway is a good candidate for TypeScript, it is small for fast rewrite and very important because it is public interface. Tell me if this contribution would be helpful.

PR #990, opened 2020-08-13
feat(@cubejs-client/react): Expose progress in useCubeQuery hook

anton164:
**Check List**

**Tests:**
There are no tests for this package at the moment, so I wasn't able to add any for this extension.

**Docs:**
I updated the `index.d.ts` file, but have not updated the other docs/examples. Are they automatically generated somehow or shall I do it manually?

When loading a Cube Query, it might take some due to waiting in the queue or because pre-aggregations are being built.

By exposing the ProgressResponse from the CubeJS API, the UI can share more information with the user when the query is loading.

This allows us to show the following in our UI (big win for our users):
Screenshot 2020-08-13 at 13 41 48

___

I hope you find this to be a useful addition to Cube.JS :)

### DIFF
--- a/packages/cubejs-client-react/src/useCubeQuery.js
+++ b/packages/cubejs-client-react/src/useCubeQuery.js
@@ -17,14 +17,7 @@ export default (query, options = {}) => {
 
   let subscribeRequest = null;
 
-  const progressCallback = ({ progressResponse }) => {
-    setProgress({
-      rawResponse: progressResponse,
-      message: progressResponse.stage
-        ? progressResponse.stage.stage
-        : 'Loading...'
-    });
-  };
+  const progressCallback = ({ progressResponse }) => setProgress(progressResponse);
 
   useEffect(() => {
     const { skip = false, resetResultSetOnChange } = options;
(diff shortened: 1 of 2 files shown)


packages/cubejs-client-react/src/useCubeQuery.js
```diff
@@ -11,11 +11,21 @@ export default (query, options = {}) => {
...
   const [isLoading, setLoading] = useState(false);
   const [resultSet, setResultSet] = useState(null);
+  const [progress, setProgress] = useState(null);
   const [error, setError] = useState(null);
   const context = useContext(CubeContext);
 
   let subscribeRequest = null;
 
+  const progressCallback = ({ progressResponse }) => {
+    setProgress({
```
vasilev-alex, line 21: Let's just save the `ProgressResponse` directly and let the user customize it on their side

vasilev-alex (changes requested):
Hey @anton164! Thanks for the PR

packages/cubejs-client-react/src/useCubeQuery.js
anton164: I can do that, but then the code for consuming the progress response will be much more convoluted?

I.e. the caller needs to access `.stage.stage` to get the message 🤷‍♂️ 

```
 export default function App() {
   *   const { resultSet, isLoading, error } = useCubeQuery({
   *   const { resultSet, isLoading, error, progress } = useCubeQuery({
   *     measures: ['Orders.count'],
   *     dimensions: ['Orders.createdAt.month'],
   *   });
   *
   *   if (isLoading) {
   *     return <div>Loading...</div>;
   *     return <div>{progress && progress.stage && progress.stage.stage || 'Loading...'}</div>;
   *   }
```

anton164: Updated now 👍

PR #999, opened 2020-08-16
refactor(@cubejs-backend/schema-compiler): use snake_case for all date operators in `BaseFilter`

RusovDmitriy:
**Check List**

https://github.com/cube-js/cube.js/pull/972#discussion_r471025382

What changes:

- All date operators replaced to camelCase style, including all the tests
- Add isDateOperator() method in BaseFilter

PS: 
I think good point for refactoring:
- Too bad the constants are not used and the problem of shared between modules base interfaces is not solved. Turn your attention to such DDD techniques as "ubiquitous language" and as an option for implement it is protobuf. 

- Test excluded from eslint checker, i recommend that you treat the tests same way you treat the work code

### DIFF
--- a/packages/cubejs-schema-compiler/adapter/BaseFilter.js
+++ b/packages/cubejs-schema-compiler/adapter/BaseFilter.js
@@ -7,7 +7,8 @@ const {
 
 const BaseDimension = require('./BaseDimension');
 
-const DATE_OPERATORS = ['in_date_range', 'not_in_date_range', 'on_the_date', 'before_date', 'after_date'];
+// Todo: `onTheDate` is missing in the documentation, possibly deprecated?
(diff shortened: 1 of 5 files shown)


paveltiunov:
Hey @RusovDmitriy ! Looks good! There should be also corresponding changes in api-gateway I believe.

RusovDmitriy:
> looks good! There should be also corresponding changes in api-gateway I believe.

Hmm, but i didn't find some snake_case operators in api-gateway.

In api-gateway:

1. Define operators array, used in joi validation schema https://github.com/cube-js/cube.js/blob/master/packages/cubejs-api-gateway/index.js#L112

2. normalize query, i don't see any transform operators to snake_case  https://github.com/cube-js/cube.js/blob/master/packages/cubejs-api-gateway/index.js#L162

If there are any moments that I haven't seen, please write about it

PR #1177, opened 2020-10-05
misc: Use yarn workspaces to reduce disk usage

ovr:
Hello!

@vasilev-alex Ask a great question:

If we  install `typescript` locally for each package inside our monorepo, how much disk space we need?

32 * 50mb (typescript = 1.6 gb

In this PR I am going to check the ability to use `workspaces` from yarn. It's not working now, but we will see when Yarn release a bug fix for it.

Current usage:

```
 ovr@MBP-Dmitry-2  ~/projects/cube/cube.js   master ?  du -sh                                                                                                                  64 ↵
3,7G	.
```

vs

```
du -sh                                                                                    ✔  10.17.0 ⬢  11.26 Dur
1,6G	.
```

image

Awaiting: https://github.com/yarnpkg/yarn/issues/7807


merceyz:
> Awaiting: yarnpkg/yarn#7807

Most likely fixed in V2, try using `yarn set version 2 && yarn config nodeLinker node-modules`, if you got questions you can get help on our Discord https://discord.com/invite/yarnpkg

ovr:
@merceyz 

Thank you for your recommendations related Yarn 2 & PnP.

hassankhan:
@ovr @merceyz I think we need to get the repo in shape with Yarn v1 Workspaces first. Yarn v2 and PnP add a lot of new variables into the module resolution problem and it would be nice if that didn't block efforts to get a basic Workspace set up.

merceyz:
> Yarn v2 and PnP add a lot of new variables into the module resolution problem and it would be nice if that didn't block efforts to get a basic Workspace set up.

Yarn v2 and the node_modules linker is better and more stable than the one in V1, but that's up to you :)

PR #1200, opened 2020-10-13
Playground look-n-feel

tenphi:
New design for the playground.


vasilev-alex (commented):
I'm not quite sure if we need to mix the `styled components` approach and `less` (considering it's used in few places). I guess we could use https://styled-components.com/docs/api#createglobalstyle 

LGTM otherwise

tenphi:
Discussed & fixed. Ready to merge.

PR #1222, opened 2020-10-19
short token

Levhav:
**Check List**


ovr (approved):
LGTM

packages/cubejs-cli/src/config.js
```diff
@@ -77,6 +89,22 @@ export class Config {
...
 
+    const answer = await this.cloudTokenReq({
+      url: `${process.env.CUBE_CLOUD_HOST || 'https://cubecloud.dev'}/v1/token`,
+      method: 'POST',
+      body: {
+        token: authToken
+      }
+    });
+    
+    if (answer.error) {
```
ovr, line 100: If `cloudTokenReq` is a general method for calling all available API, I think we should move this check for error to it.

PR #1272, opened 2020-10-30
fix(@cubejs-backend/schema-compiler): deprecation warning

### LINKED ISSUE #1268: Warning: Attribute filter.dimension is deprecated - appears at every query
Hi! What do you think about turning this warning: 'Warning: Attribute filter.dimension is deprecated. Please use 'member' instead of 'dimension', into a console.info?. Since it appears on each request, even though in the query filter the new operator name 'member' is used, I consider it to be more of an informative message than a warning. It's helpful also for logging, where you want to filter out only warnings, and this message appears too much. Thanks!

vasilev-alex:
Closes #1268

### DIFF
--- a/packages/cubejs-schema-compiler/adapter/BaseQuery.js
+++ b/packages/cubejs-schema-compiler/adapter/BaseQuery.js
@@ -113,7 +113,7 @@ class BaseQuery {
         throw new UserError(`You cannot use dimension and measure in same condition: ${JSON.stringify(f)}`);
       }
 
-      if (!f.dimension && !f.member) {
+      if (!f.member && !f.dimension) {
         throw new UserError(`member attribute is required for filter ${JSON.stringify(f)}`);
       }
 


packages/cubejs-schema-compiler/adapter/BaseQuery.js
```diff
@@ -113,7 +113,7 @@ class BaseQuery {
         throw new UserError(`You cannot use dimension and measure in same condition: ${JSON.stringify(f)}`);
       }
 
-      if (!f.dimension && !f.member) {
+      if (!f.member && !f.dimension) {
```
ovr, line 116: A & B == B & A (associative operation), Why did you change it?)

ovr: Similar case https://github.com/cube-js/cube.js/blob/master/packages/cubejs-schema-compiler/adapter/BaseQuery.js#L63

vasilev-alex: @ovr the idea behind the change is this warning https://github.com/cube-js/cube.js/blob/054c3357ad2f1ad7d6ffb2312c43299e738774d1/packages/cubejs-api-gateway/index.js#L305

If we check for `.member` first and it's there we skip `.dimension` call and therefore avoid printing the warning to the console

PR #1281, opened 2020-11-02
feat(cubejs-cli): Fully move CLI to TS

ovr:
Hello!

Thanks


packages/cubejs-cli/src/command/token.ts
```diff
@@ -0,0 +1,86 @@
...
+  expiresIn?: string
+  payload: string[]
+  userContext: string[]
+};
+
+export const token = async (options: TokenOptions) => {
+  event('Generate Token');
+
+  const CubejsServer = await requireFromPackage('@cubejs-backend/server');
+  const { expiry = defaultExpiry, secret = CubejsServer.apiSecret() } = options;
```
vasilev-alex, line 32: doesn't it mean that `expiry` and `secret` are optional?

packages/cubejs-cli/src/config.ts
```diff
@@ -23,28 +33,30 @@ export class Config {
...
     return path.join(os.homedir(), '.env');
   }
 
-  cubeCloudConfigPath() {
+  protected cubeCloudConfigPath() {
     return path.join(os.homedir(), '.cubecloud');
   }
 
-  async deployAuth(url) {
+  public async deployAuth(url?: string) {
```
vasilev-alex, line 59: all methods are public by default

packages/cubejs-cli/src/command/token.ts
ovr: Good catch, fixed!

packages/cubejs-cli/src/config.ts
ovr: You are right, but when it's not written explicitly we can skip some places where methods, properties must be protected/private.

I enabled rule that will help us to protect this moment.

https://github.com/cube-js/cube.js/pull/1281/commits/f96c6e16312c8dc5471b447d3aeccf63d77d108f#diff-44ead173231c78d119ff39f7c132458a4cd7dde766706f8af36ae7b856ed149aR52

It's useful, example of detected problems.

https://github.com/cube-js/cube.js/pull/1281/commits/cf78849d061632057cbd660f8e100ff9b4752bd3#diff-e6ac146ee7d51883c4831298bc0c00a879996706e99b90b2cfc5139d2705440dR30

PR #1293, opened 2020-11-03
feat(@cubejs-client/ws-transport): Move to TypeScript

ovr:
Hello!

refs https://github.com/cube-js/cube.js/issues/1091

Thanks


vasilev-alex (approved):

packages/cubejs-client-ws-transport/src/index.ts
```diff
@@ -11,18 +24,48 @@ class WebSocketTransportResult {
   }
 }
 
+type WebSocketTransportOptions = {
+  authorization: string,
+  apiUrl: string,
+  hearBeatInterval?: number,
```
hassankhan, line 30: I think this should be `heartBeatInterval?`

vasilev-alex: @hassankhan nice catch! I guess we've had it like this since the very beginning. 

I think we should add the correct one but keep the `hearBeatInterval` option (and deprecate it) for backward compatibility

hassankhan: Ahh good point, I'd forgotten it had always been like that. Totally agree.

ovr: Done ✅

PR #1359, opened 2020-11-11
docs: minor fixes to env vars page

keydunov:
Minor things to improve the readability of the page.


keydunov:
@hassankhan I removed the category of the env var from the table to free some space. It seems it doesn't provide much value to users anyway.

hassankhan (approved):
@keydunov Makes sense 👍

PR #1378, opened 2020-11-12
docs: update 'Caching' page

keydunov:


docs/Cube.js-Backend/Caching.md
```diff
@@ -21,24 +20,24 @@ key is generated SQL with any existing query dependent pre-aggregations.
 
 Upon incoming request Cube.js first checks the cache using this key.
 If nothing is found it executes the query in database, returns result back alongside writing to the cache.
-Since 0.15.0 in case there's an existing value in cache it will be returned only if `refreshKey` value for query isn't changed.
+In case there's an existing value in cache it will be returned only if `refreshKey` value for query isn't changed.
```
igorlukanin, line 23: I guess it's quite hard to understand this sentence because it starts with a conditional part and ends with another conditional part. What if we put it like this?

```suggestion
If an existing value is present in the cache and the `refreshKey` value for the query hasn't changed, the cached value will be returned.
```

igorlukanin (approved):

docs/Cube.js-Backend/Caching.md
hassankhan: 👍

hassankhan (approved):

docs/Cube.js-Backend/Caching.md
```diff
@@ -77,25 +79,25 @@ If you need to force a specific query to load fresh data from the database (if i
...
 
 For situations like real-time analytics or responding to live user changes to underlying data, the `refreshKey` query cache can prevent fresh data from showing up immediately.
-For these situations, you can mostly disable the `refreshKey` cache by setting the [refreshKey every](cube#parameters-refresh-key) parameter to something very low, like `1 second`.
+For these situations, the `refreshKey` cache can effectively be disabled by setting the [refreshKey every](cube#parameters-refresh-key) parameter to something very low, like `1 second`.
```
keydunov, line 82: @paveltiunov @hassankhan I think it is ok to leave the `1 second` suggestion for now, but it is potentially a dangerous road and we need to replace it with streaming techniques.

PR #1396, opened 2020-11-17
feat(docs-build): add `gatsby-redirect-from` to allow redirects with SEO

### LINKED ISSUE #1395: Add `gatsby-redirect-from` plugin
We'd like to be able to reorder and change URLs where documentation is hosted, but without losing SEO. We need a solution that can generate 301 redirects for pages that don't exist any more.

We should add [`gatsby-redirect-from`](https://www.gatsbyjs.com/plugins/gatsby-redirect-from/). This plugin will give us the flexibility we need for changing URLs and redirecting the old ones, all while keeping our SEO scores up.

hassankhan:
**Check List**

Fixes #1395


igorlukanin (approved):
If it works (at least on staging), then it LGTM 😄

PR #1400, opened 2020-11-17
docs: Remove refs of the NODE_ENV

keydunov:


docs/Deployment/Production-Checklist.md
```diff
@@ -8,15 +8,17 @@ menuOrder: 2
 This is a checklist for configuring and securing Cube.js for a production
 deployment.
 
-## Enable Production Mode
+## Disable Dev Mode
 
-When running Cube.js in production make sure `NODE_ENV` is set to `production` and `CUBEJS_DEV_MODE` is not set to `true`.
-Some platforms, such as Heroku, do it by default.
+When running Cube.js in production make sure dev mode is disabled. Running dev
+mode in production environment can lead to security vulnerabilities.
```
igorlukanin, line 14: I doubt that using "security vulnerabilities" here is appropriate because it really depends on the software using Cube.js and the nature of the data. IMO, it's quite misleading right now.

I think we can be more specific and add a relevant link here:

```suggestion
When running Cube.js in production environment, make sure the development mode is disabled. Running Cube.js in development
mode in a production environment allows accessing Cube.js API [without authorization](https://cube.dev/docs/security).
```

docs/Deployment/Production-Checklist.md
```diff
@@ -8,15 +8,17 @@ menuOrder: 2
...
 
-In this mode, the insecure development server and Playground will be disabled
-by default because there's a security risk serving those in production
-environments. Production Cube.js servers can only be accessed through the
-[REST API](rest-api) and Cube.js frontend libraries.
+The dev mode is disabled by default.
```
igorlukanin, line 16:
```suggestion
The development mode is disabled by default.
```

docs/Deployment/Production-Checklist.md
```diff
@@ -8,15 +8,17 @@ menuOrder: 2
 This is a checklist for configuring and securing Cube.js for a production
 deployment.
 
-## Enable Production Mode
+## Disable Dev Mode
```
igorlukanin, line 11: I think that, generally, we should refrain from using contractions because they decrease the accessibility of our docs. It's okay to have `DEV` in the code though.
```suggestion
## Disable Development Mode
```

igorlukanin (approved):
I've added a number of suggestions that I deem good enough to be incorporated.

@hassankhan Would also love to know your take on this.

hassankhan (approved):
Updated as per @igorlukanin's great feedback 👍

keydunov:
@igorlukanin @hassankhan Disagreed on security note. Disabling dev mode has multiple implications, including, but not limited to disabling authentication. Log level and exceptions can lead to security vulnerabilities either.

igorlukanin:
@keydunov You're right, no doubt. My point is that if we say one should take care to disable the development mode, it's better to describe what are the implications specifically rather than leaving a (quite vague!) statement about security vulnerabilities.

Should we have a (sub)section in the docs that describes in detail what the development mode is? Here in production checklist or just linked from here.

keydunov:
@igorlukanin we have this section already. I'm ok to revert the text back to the original and link to the https://cube.dev/docs/configuration/overview#development-vs-production cc @hassankhan

hassankhan:
Reverted the text, also added a callout

/cc @keydunov @igorlukanin

PR #1406, opened 2020-11-18
fix(@cubejs-client/core): propagate segments to drillDown queries

### LINKED ISSUE #1394: Add support for `segments` in `drillDown` query
When we are doing a `drillDown()` on `resultSet`, it doesn't accommodate `segments:[]`, There fore, chart data and drilldown result data are a mismatch. however, if we add it manually to the drilldown query it works. 

Ideally, `segements:[]` should be handled by `drillDown()`

For now I am adding it manually to the drillDown query, if its available in the `resultSet`

vasilev-alex:
Closes #1394


SandeepVattapparambil:
Nice, thank you so much for this fast response !
Kudos

PR #1438, opened 2020-11-24
misc(@cubejs-backend/server): change `x-powered-by` header to Cube.js

ovr:
Hello!

Thanks

### DIFF
--- a/packages/cubejs-server/src/server.ts
+++ b/packages/cubejs-server/src/server.ts
@@ -74,6 +74,7 @@ export class CubejsServer {
 
       const app = express();
 
+      app.set('x-powered-by', 'Cube.js');
       app.use(cors(this.config.http.cors));
       app.use(bodyParser.json({ limit: '50mb' }));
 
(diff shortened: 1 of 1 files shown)


igorlukanin:
Looks good to me 👍 

However, are header names case-insensitive? Should we make it `X-Powered-By` instead?

ovr:
> However, are header names case-insensitive? Should we make it X-Powered-By instead?

It this case, it's not a header, it's an option.

https://github.com/expressjs/express/blob/master/lib/middleware/init.js#L30

hassankhan (approved):

packages/cubejs-server/src/server.ts
```diff
@@ -74,6 +74,7 @@ export class CubejsServer {
 
       const app = express();
 
+      app.set('x-powered-by', 'Cube.js');
```
vasilev-alex, line 77: Use `Cube` please

vasilev-alex (changes requested):

packages/cubejs-server/src/server.ts
ovr, line 77:
```suggestion
      app.set('x-powered-by', 'Cube');
```

ovr: Nice catch!

PR #1442, opened 2020-11-24
Fix a typo in @cubejs-client-ws-transport docs

igorlukanin:
Hey @hassankhan and @ovr, I think I've found a typo. Would you please verify and fix it? 🙏


hassankhan:
Hi @igorlukanin, I spoke to @vasilev-alex and @ovr about this - basically it used to be `hearBeatInterval` but has now been deprecated by `heartBeatInterval`. This might be worth fixing in the Typescript definition files so that it automatically updates the file in this PR 👍

ovr:
@hassankhan is right, this file is auto-generated from d.ts file. It's not needed to update anything here.

igorlukanin:
Okay, closing this PR then. Thanks for the investigation!

PR #1456, opened 2020-11-26
docs: merge `docs/` and `docs-build/`

hassankhan:
**Check List**

Merging the `docs/` and `docs-build/` folders. Also updated paths in CI where appropriate.


igorlukanin (approved):

igorlukanin:
@hassankhan, I'm good with the changes 👍 

However, I wonder if it would make more sense to move `/docs-build` under `/docs/build` rather than moving `/docs` to `/docs/content`. It's extremely subjective, but IMO the actual content is more important than the docs build code, so we might try to have the content as close to the repo root as possible. WDYT?

hassankhan:
@igorlukanin Thanks for the feedback. There are a few reasons why we did it this way:

1. Moving to `docs/build/` would create complications for #1177
2. The 'Edit Page' links in the static docs will point to the right place, so no loss of access there
3. We want `docs/` to be a Node package so we can add tooling like Markdownlint etc., without having to make the textual content a separate Node package.

igorlukanin:
@hassankhan, thanks for the explanation! Totally makes so much sense!

ovr (approved):
LGTM

PR #1459, opened 2020-11-26
docs: add documentation for `CUBEJS_SSL_*` environment variables

hassankhan:
**Check List**

Update the 'Environment Variables' reference with SSL variables.


igorlukanin (approved):

igorlukanin:
@hassankhan, we definitely should have these `SSL_` variables to the docs. Good job 👍 

However, I wonder if it would make sense to extract them to a separate table and put that table to a subsection of *Database Connection*. We already have quite a list in this section, so organizing information in any way can improve its accessibility. Also, it would give us an option to put a laser-focused context link here: https://cube.dev/docs/connecting-to-the-database#notes-ssl. WDYT?

hassankhan:
@igorlukanin Makes sense, would need to get @keydunov's thoughts.

Separately, I think this PR should perhaps include some "explanation" documentation in [Connecting to the Database](https://cube.dev/docs/connecting-to-the-database). Would love to get your thoughts, @igorlukanin @keydunov

igorlukanin:
@hassankhan, I fully agree! I think that...

* we might have some explanation of a "generic" database connection flow around here (https://cube.dev/docs/connecting-to-the-database#source-database) and also lift the SSL part (https://cube.dev/docs/connecting-to-the-database#notes-ssl) up here
* we might remove the table with the database-to-env-vars mapping because it doesn't make a lot of sense without explaining what each env var, if non-trivial, is for
* we might add a subsection for every database where the content from the table mentioned above will be put

keydunov:
@igorlukanin @hassankhan Agreed, we need to have an "explanation" piece in **Connecting to the Databases** section and replace links to the Node's docs with links to this new section.

keydunov:
Additionally, can we remove links to Node doc? They have a specific node version and I'm afraid could be outdated if we change something. I'd prefer to keep the interface and hide the implementation details.

hassankhan:
@keydunov Did you mean the links in the 'Environment Variables' page?

keydunov:
@hassankhan Yes

keydunov:
:shipit:

PR #1518, opened 2020-12-04
revert: bootstrap dependencies via `lerna`

ovr:
Hello!

Thanks


hassankhan:
After discussing with @vasilev-alex, let's close this for now. I'm happy to keep an eye on the CI and help out with issues (if any) that might arise from keeping it as-is 👍

PR #1540, opened 2020-12-08
fix(@cubejs-backend/api-gateway): Add missing exports

willhausman:
#1474 hid two classes I was relying on.


hassankhan (approved):

ovr (approved):
LGTM, because https://github.com/cube-js/cube.js/pull/288

PR #1570, opened 2020-12-10
fix: use `decimal(38,10)` for external MySQL and Postgres pre-aggregations

ovr:
Hello!

refs https://github.com/cube-js/cube.js/issues/1563

Thanks

### DIFF
--- a/packages/cubejs-mysql-driver/driver/MySqlDriver.js
+++ b/packages/cubejs-mysql-driver/driver/MySqlDriver.js
@@ -3,11 +3,11 @@ const genericPool = require('generic-pool');
 const { promisify } = require('util');
 const { BaseDriver } = require('@cubejs-backend/query-orchestrator');
 const crypto = require('crypto');
-const fs = require('fs');
 
 const GenericTypeToMySql = {
   string: 'varchar(255) CHARACTER SET utf8mb4',
-  text: 'varchar(255) CHARACTER SET utf8mb4'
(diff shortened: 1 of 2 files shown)


benswinburne:
Hi @ovr - any reason in particular for choosing two decimal places for this?

Does it not make sense to extend it slightly further perhaps so that users have the option to format/round it as they please as opposed to being limited by the column setup?

ovr:
> Does it not make sense to extend it slightly further perhaps so that users have the option to format/round it as they please as opposed to being limited by the column setup?

It's a quick solution to fix problem with MySQL only for now. I think that It will be cool to improve conversion from DriverType -> GenericType -> DriverType.

benswinburne:
Thanks very much for your work on this. I wondered if you might be able to suggest how I best use this fix until such time it's been reviewed and merged here.

Typically I would just refer to the git hash in my package.json but as this is a monorepo I can't get it to work.

    "@cubejs-backend/mysql-driver": "cube-js/cube.js#4d783f39b684366fbfdf7df0df88883514a1180b",

Any suggestions on how i can use this fix temporarily would be much appreciated.

benswinburne:
I have taken a temporary copy of the mysql-driver into its own repo and referenced that in case it's useful to anyone else trying to work this out. Then I just referred to it as described above in package json

    "@cubejs-backend/mysql-driver": "benswinburne/cubejs-mysql-driver#50236e386fc056199b4ca6e18febd4a5a921ba5e",

Issue #1573, opened 2020-12-11
chartPivot() fails when data blending on the same measure 


MattGson:
**Describe the bug**

When doing cohort analysis, often we want to compare the same measure for multiple different groups.
The definitions of these groups can be complex (think audiences)

i.e. Compare:
-         average purchase price for users in group A who have made less than 10 purchases
-         average purchase price for users in group B who have made at least 10 purchases

As far as I can see, doing this with one query is difficult due to the way dimensions group the data. The resulting groups will be wrong because it will group by the actual value of Users.purchase_count, not the binary filters <10 or >=10. There is also no way to tell cube to combine the groupings correctly as far as I can tell?

A easy and obvious solution is to create one query for each cohort. This works well and is easy to construct (and probably more efficient to execute for a small number of cohorts).

The problem is that using `chartPivot()` on the client fails. Only one of the queries results end up in the data set. This is likely because `chartPivot()` just sets the key in the object based on the measure name (which is the same for both series) and ends up overwriting each query with the previous next one.

**Describe the solution you'd like**
A potential solution would be to give every query a name/title. This would be used to prefix the measures for the resulting series. 
i.e.
Query:
```
 { 
     name: `Group A < 10`
     measures: [...],
    ....
 }
```

ChartPivot
```
{ x: '2020-10-15T00:00:00', 'GroupA<10.Users.AveragePurchase': 6, ... }
```

**Describe alternatives you've considered**
Bypassing or extending the chartPivot functionality manually.

vasilev-alex:
Hey @MattGson! Thanks for the detailed description. Yes, we need to fix it.

PR #1594, opened 2020-12-14
Added pivotConfig option to alias series with a prefix

MattGson:
Allows for multiple series using the same measure in
blending queries.

Currently works in seriesNames and chartPivot only

**Check List**
- [☑️ ] Tests has been run in packages where changes made if available
- [☑️ ] Linter has been run for changed code
- [☑️ ] Tests for the changes have been added if not covered yet
- [☑️ ] Docs have been added / updated if required

#1573

Added an option to `pivotConfig` called `aliasSeries ` with type `string[]`.
This takes a list of series aliases which are prefixed to the corresponding series in the pivot.

This has only been implemented for the `chartPivot` and `seriesNames` methods currently.

**Note**: There seems to be a lot of line changes due to the existing code not following the `prettier` guidelines.
Actual changes at 435 and 592


packages/cubejs-client-core/index.d.ts
```diff
@@ -178,6 +178,10 @@ declare module '@cubejs-client/core' {
      * If `true` missing dates on the time dimensions will be filled with `0` for all measures.Note: the `fillMissingDates` option set to `true` will override any **order** applied to the query
      */
     fillMissingDates?: boolean | null;
+    /**
+     * Give each series a prefix alias. Should have one entry for each query:measure.
+     */
+    aliasSeries: string[];
```
vasilev-alex, line 184: This should probably be optional

vasilev-alex (changes requested):
Hey! Thanks for your time submitting this PR.

I think we could even make the aliasing thing automatic when no `pivotConfig` is provided, so it doesn't come as a surprise resulting in missing all duplicate measures. What do you think?

Also, could you please only submit the actual changes without any formatting applied by prettier as it currently breaks our linter rules and makes history difficult to follow.

packages/cubejs-client-core/index.d.ts
MattGson: Fixed

MattGson:
@vasilev-alex I have made the suggested changes including an automatic alias for duplicate measures.

Currently the alias will just use the index i.e. 0,Users.Count 1,Users.count ...

There is potential for some abstractions/re-use around the aliasing code but I'm not super familiar with the code-base and the ResultSet code isn't the easiest to follow (Typescript would be a big improvement!).

Issue #1558, opened 2020-12-09
Query with many filter values ends up in 413 URL too large


Lozzi:
**Problem**

I ran into an issue when querying with a filter that had a lot of long string values that the request URL was too large [error 413](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/413#:~:text=The%20HTTP%20413%20Payload%20Too,a%20Retry%2DAfter%20header%20field).

The request is 4534 characters in legnth.

In a previous project we implemented a check for the length of the query and if it exceeded 2048 characters we did a POST with body instead of the usual GET request.

Is there any work around for this currently in Cube or has anyone ran into the same issue?

hassankhan:
Hi @Lozzi, you can [use the `POST` method](https://cube.dev/docs/rest-api#api-reference-v-1-load) to send queries to Cube.js.

ovr:
@hassankhan Our `@cubejs-client/core` is doing everything by GET method inside Request, I think we should prepare a fix for that.

Lozzi:
@ovr Yes, forgot to mention that I'm using the @cubejs-client/core to make the requests. I could try to take a look at it and submit a Pull Request over the holidays.

hassankhan:
@Lozzi That would be amazing 😄

telunc:
We ran into this issue with our Angular app. Our angular app initialize client in app.module with forRoot meaning that client is initialized once at start. In HttpTransport, `this.method = this.method || (url.length < 2000 ? 'GET' : 'POST');` the later url length logic will only get run once if method is undefined. Any request afterwards will be taken from `this.method`.

We can't pass in `POST` directly in client option as well since Meta is a `GET` endpoint.

PR #1608, opened 2020-12-15
feat(@cubejs-client/core): allow making HTTP POST requests to Cube.js API

mnifakram:
**Check List**

 #1558

### DIFF
--- a/packages/cubejs-client-core/src/HttpTransport.js
+++ b/packages/cubejs-client-core/src/HttpTransport.js
@@ -2,34 +2,42 @@ import fetch from 'cross-fetch';
 import 'url-search-params-polyfill';
 
 class HttpTransport {
-  constructor({ authorization, apiUrl, headers = {}, credentials }) {
+  constructor({ authorization, apiUrl, method, headers = {}, credentials }) {
     this.authorization = authorization;
     this.apiUrl = apiUrl;
+    this.method = method;
     this.headers = headers;
     this.credentials = credentials;
   }
 
   request(method, { baseRequestId, ...params }) {
+    let spanCounter = 1;
     const searchParams = new URLSearchParams(
       params && Object.keys(params)
         .map(k => ({ [k]: typeof params[k] === 'object' ? JSON.stringify(params[k]) : params[k] }))
         .reduce((a, b) => ({ ...a, ...b }), {})
     );
 
-    let spanCounter = 1;
+    let url = `${this.apiUrl}/${method}${searchParams.toString().length ? `?${searchParams}` : ''}`;
+    
+    this.method = this.method || (url.length < 2000 ? 'GET' : 'POST');
+    if (this.method === 'POST') {
+      url = `${this.apiUrl}/${method}`;
+      this.headers['Content-Type'] = 'application/json';
+    }
 
     // Currently, all methods make GET requests. If a method makes a request with a body payload,
-    // remember to add a 'Content-Type' header.
-    const runRequest = () => fetch(
-      `${this.apiUrl}/${method}${searchParams.toString().length ? `?${searchParams}` : ''}`, {
-        headers: {
-          Authorization: this.authorization,
-          'x-request-id': baseRequestId && `${baseRequestId}-span-${spanCounter++}`,
-          ...this.headers
-        },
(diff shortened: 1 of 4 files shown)


mnifakram:
@hassankhan @ovr can you please guys review? Thanks.

packages/cubejs-client-core/src/HttpTransport.js
```diff
@@ -2,34 +2,40 @@ import fetch from 'cross-fetch';
 import 'url-search-params-polyfill';
 
 class HttpTransport {
-  constructor({ authorization, apiUrl, headers = {}, credentials }) {
+  constructor({ authorization, apiUrl, method = 'GET', headers = {}, credentials }) {
```
hassankhan, line 5: @ovr Should we also be changing the default to `POST` too?

vasilev-alex: We should prefer the `GET` method whenever possible

vasilev-alex:
So, currently, I can see one limitation. It's not possible to change the `method` once we've created the `cubejsApi` instance.

```js
const transport = new HttpTransport({
  // ...
  method: 'POST'
});

const cubejsApi = cubejs(CUBEJS_TOKEN, {
  apiUrl: `${API_URL}/cubejs-api/v1`,
  transport
});

// Can't change the `method` during runtime
// I want to send this one using `GET`
cubejsApi.load(query);
// and this one using `POST`
cubejsApi.load(query);
```

What if we would pass the HTTP `method` to the load?

```js
cubejsApi.load(query, { method: 'GET' });
cubejsApi.load(query, { method: 'POST' });
```

Or, maybe, we could allow users to override the method by passing it to the `load` method? Do we even need it?

@ovr @hassankhan @mnifakram what do you think?

mnifakram:
@vasilev-alex I don't think that users will switch between `GET` or `POST`, that's an edge case. In real world as a user if I didn't encounter the large URL problem and I only have `POST`, I will not even complain about not having `GET` as an option. 

I don't see any advantages of using `GET` over `POST`. Not to mention that, it is way easier to inspect the query in the request body vs trying to decoded from the request URL.

packages/cubejs-client-core/src/HttpTransport.js
ovr: 100% agree with @vasilev-alex, it's better to use GET whenever possible.

ovr:
> I don't see any advantages of using GET over POST.

`GET` requests support cache inside browser, and possible Cube.js will return specific headers to support it.

packages/cubejs-client-core/src/HttpTransport.js
ovr: @vasilev-alex @hassankhan Maybe it's better to set up it to `undefined` by default and detects automatically. 

```
method = method || (url.length < 2000 ? 'GET' : 'POST')
```

mnifakram:
Is there anything that I can help with to get this merged? I'm really in need to get this released asap. Thanks

himessa:
Any update on this? I need the ability to handle large queries as well.

packages/cubejs-client-core/src/HttpTransport.js
vasilev-alex: @ovr yep, let's do like this. @mnifakram can you please introduce this check?

mnifakram: yes sure thing

mnifakram: @ovr @vasilev-alex I've just updated it. Please kindly check it. thanks

packages/cubejs-client-core/src/index.js
```diff
@@ -25,11 +25,13 @@ class CubejsApi {
     options = options || {};
     this.apiToken = apiToken;
     this.apiUrl = options.apiUrl || API_URL;
+    this.method = options.method || 'GET';
```
vasilev-alex, line 28: In this case, it will always be `GET` by default

mnifakram: oops forgot to change this file, I've updated it again. Thanks

PR #1623, opened 2020-12-16
fix(@cubejs-core): Add type definition for compareDateRange, fix #1621

### LINKED ISSUE #1621: `compareDateRange` needs to be added to the type definitions for
**Describe the bug**

The article on https://cube.dev/blog/comparing-data-over-different-time-periods/ describes a new `compareDateRange` property available on TimeDimensions. Based on what I could see this is available in the repository however it doesn't look like the type definitions were added so it is unavailable for Typescript projects. 

**To Reproduce**
1. npm install the latest cubejs-core (0.24.8 as of this writing)
2. Include cubejs-core
3. Try to instantiate a query with a compareDateRange property in a Typescript file.

**Expected behavior**
Typescript projects can use the `compareDateRange` property

**Version:**
 * `@cubejs-client/core": "^0.24.8"`
 * ` @cubejs-client/react": "^0.24.8"`

cbroome:
**Check List**

#1621 

Added compareDateRanges as a type to TimeDimension

### DIFF
--- a/packages/cubejs-client-core/index.d.ts
+++ b/packages/cubejs-client-core/index.d.ts
@@ -691,7 +691,7 @@ declare module '@cubejs-client/core' {
   }
 
   export interface TimeDimensionRanged extends TimeDimensionBase {
-    dateRange: string | string[];
+    dateRange?: string | string[];
   }
 
   export type TimeDimension = TimeDimensionComparison | TimeDimensionRanged;
--- a/packages/cubejs-client-core/index.d.ts
+++ b/packages/cubejs-client-core/index.d.ts
@@ -680,19 +680,19 @@ declare module '@cubejs-client/core' {
 
   export type TimeDimensionGranularity = 'second' | 'minute' | 'hour' | 'day' | 'week' | 'month' | 'year';
 
-  export interface TimeDimensionBase {
+  export type TimeDimensionBase = {
     dimension: string;
     granularity?: TimeDimensionGranularity;
-  }
+  };
 
-  export interface TimeDimensionComparison extends TimeDimensionBase {
+  export type TimeDimensionComparison = TimeDimensionBase & {
     compareDateRange: Array<string | Array<string>>;
     dateRange?: never;
-  }
+  };
 
-  export interface TimeDimensionRanged extends TimeDimensionBase {
+  export type TimeDimensionRanged = TimeDimensionBase & {
     dateRange?: string | string[];
-  }
+  };
 
   export type TimeDimension = TimeDimensionComparison | TimeDimensionRanged;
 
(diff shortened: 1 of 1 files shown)


vasilev-alex (changes requested):
Hi @cbroome! Thanks for the PR.

The update you're suggesting does not cover all possible cases for `compareDateRange` https://cube.dev/docs/query-format#time-dimensions-format

On top of that, the `compareDateRange` and `dateRange` props are mutually exclusive, so we should consider it.

cbroome:
@vasilev-alex updated, let me know if that works.

packages/cubejs-client-core/index.d.ts
```diff
@@ -680,11 +680,21 @@ declare module '@cubejs-client/core' {
...
-  };
+  }
+
+  export interface TimeDimensionComparison extends TimeDimensionBase {
+    compareDateRange: Array<string | Array<string>>;
+    dateRange?: never;
+  }
+
+  export interface TimeDimensionRanged extends TimeDimensionBase {
+    dateRange?: string | string[];
```
vasilev-alex, line 694: I think the second option should be `[string, string]`

packages/cubejs-client-core/index.d.ts
```diff
@@ -680,11 +680,21 @@ declare module '@cubejs-client/core' {
 
   export type TimeDimensionGranularity = 'second' | 'minute' | 'hour' | 'day' | 'week' | 'month' | 'year';
 
-  export type TimeDimension = {
+  export interface TimeDimensionBase {
```
vasilev-alex, line 683: Could you please change to be `type`s? While interfaces also work here, our docs-gen (which generates the docs' md files) works better with type definitions rather than with interfaces.

vasilev-alex (changes requested):

packages/cubejs-client-core/index.d.ts
cbroome: Should it? It was `dateRange?: string | string[];` in the original code.

packages/cubejs-client-core/index.d.ts
cbroome: Done.

packages/cubejs-client-core/index.d.ts
vasilev-alex: It was incorrect then. `dateRange` is either a string or an array of two dates.

packages/cubejs-client-core/index.d.ts
```diff
@@ -680,12 +680,22 @@ declare module '@cubejs-client/core' {
...
 
-  export type TimeDimension = {
+  export type TimeDimensionBase = {
     dimension: string;
-    dateRange?: string | string[];
     granularity?: TimeDimensionGranularity;
   };
 
+  export type TimeDimensionComparison = TimeDimensionBase & {
+    compareDateRange: Array<string | Array<string>>;
```
vasilev-alex, line 689: `Array<string | [string, string]>`

cbroome:
@vasilev-alex Just to follow-up, making the DateRange check more explicit did necessitate some updates in my code. It could be a rude surprise to other users if they were playing loose like I was.

PR #1650, opened 2020-12-18
feat(api-gateway): Support schema inside Authorization header, fix #1297

### LINKED ISSUE #1297: Support "Authorization: Bearer xxx" token passing format in cubejs-api-gateway

It is idiomatic for JWTs to be accepted using a header format of `Authorization: Bearer <JWT>` (see [jwt.io introduction](https://jwt.io/introduction/).) In general, in history, the RFCs surrounding the authorization header have taken care to specify the mode of Authorization as the first part of the header value (e.g. Basic, Digest et al.) See [RFC 2617](https://tools.ietf.org/html/rfc2617) and [RFC 6750](https://tools.ietf.org/html/rfc6750).

However, according to Cube.js [docs](https://cube.dev/docs/rest-api), you in fact must pass the token as a bare string in the Authorization header. Since you do use JWTs idiomatically, people will naturally be drawn to specifying the Authorization header as a Bearer token.

In cubejs-api-gateway, please support the use of the idiomatic `Authorization: Bearer xxx` method of passing JWTs into a webservice.

If you had a reason to want the key passed as a bare string, you could come up with an unofficial header key and value, such as X-Cubejs-JWT. Then there would be no expectations around what format that header would have and people would double check before using the API.

ovr:
Hello!

Thanks

### DIFF
--- a/packages/cubejs-api-gateway/src/gateway.ts
+++ b/packages/cubejs-api-gateway/src/gateway.ts
@@ -677,7 +677,7 @@ export class ApiGateway {
     }
   }
 
-  protected async defaultCheckAuth(req, auth) {
+  protected async defaultCheckAuth(req: Request, auth?: string) {
     if (auth) {
       const secret = this.apiSecret;
       try {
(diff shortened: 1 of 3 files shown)


packages/cubejs-api-gateway/src/gateway.ts
```diff
@@ -673,7 +673,11 @@ export class ApiGateway {
     }
   }
 
-  protected async defaultCheckAuth(req, auth) {
+  protected async defaultCheckAuth(req: Request, auth: string|undefined, schema: string) {
```
vasilev-alex, line 676: should we default to the `bearer` schema?

vasilev-alex (approved):

packages/cubejs-api-gateway/src/gateway.ts
ovr: It's a default implementation for `checkAuth` middleware. I pass `bearer` as default value inside `extractAuthorizationHeaderWithSchema`.

igorlukanin (approved):

PR #1659, opened 2020-12-21
feat(server-core): Initial support for TS

ovr:
Hello!

Thanks


vasilev-alex (approved):

vasilev-alex (approved):

igorlukanin (approved):
LGTM!

@hassankhan and @ovr, what do you think about specifying in which use cases developers (most likely) should take care about a feature being deprecated? For example, I believe that `BaseDriver` is only imported by driver developers but not by Cube.js users so we might specify that in the description, and so on. WDYT?

ovr:
@igorlukanin 

>  I believe that BaseDriver is only imported by driver developers but not by Cube.js users so we might specify that in the description, and so on. WDYT

It's an example of a problem that be in user's code. In this context, all absolute imports are deprecated, for `Query`, `BaseDriver`, etc inside `query-ochestrator`, similar thing to `server-core`.

vasilev-alex (approved):

PR #1660, opened 2020-12-21
feat!: Drop support for CUBEJS_ENABLE_TLS

ovr:
Hello!

Thanks


packages/cubejs-backend-shared/src/env.ts
```diff
@@ -8,10 +8,6 @@ const variables = {
     .default(4000)
     .required()
     .asPortNumber(),
-  tlsPort: () => get('TLS_PORT')
-    .default(4433)
-    .required()
-    .asPortNumber(),
   tls: () => get('CUBEJS_ENABLE_TLS')
```
hassankhan, line 11: Should this property also be removed?

hassankhan (changes requested):
PR needs to be rebased to include latest changes to `DEPRECATION.md`

packages/cubejs-backend-shared/src/env.ts
ovr: @hassankhan, I am still using it to detect that CUBEJS_ENABLE_TLS is used 😄

hassankhan: Ahh yes, good point 😅

vasilev-alex (approved):

PR #1671, opened 2020-12-22
feat(@cubejs-client/playground): Database connection wizard

vasilev-alex:


packages/cubejs-server-core/src/core/DevServer.ts
```diff
@@ -25,13 +30,17 @@ export class DevServer {
   public constructor(
     protected readonly cubejsServer: CubejsServerCore,
   ) {
+    console.log(process.env);
```
ovr, line 33: Should be removed before merge

packages/cubejs-server/src/server/container.ts
```diff
@@ -33,6 +33,7 @@ export class ServerContainer {
   public constructor(
     protected readonly configuration: { debug: boolean }
   ) {
+    console.log('ServerContainer .>>>');
```
ovr, line 36: Should be removed before merge

packages/cubejs-server/src/index.ts
```diff
@@ -1,9 +1,11 @@
...
 
 export * from './server';
 
 // Internal staff, don't show it as a public api
 // export * from './command/server';
 // export * from './command/dev-server';
 
-export default CubejsServer;
+export default ServerContainer;
+// export default CubejsServer;
```
ovr, line 11: It's a breaking change because some users continue to use legacy express way
I think We should not do it for now.

ovr (approved):
LGTM, but this PR contains BC with exports for cubejs-server, IMHO: We dont need to do it now.

packages/cubejs-server-core/src/core/server.ts
```diff
@@ -173,15 +173,13 @@ export class CubejsServerCore {
       schemaPath: process.env.CUBEJS_SCHEMA_PATH || 'schema',
       ...opts,
     };
-
-    if (
```
ovr, line 177: Maybe it's better to pass an additional flag in options to disable this checks when the wizard is in progress? Instead of commenting this code

packages/cubejs-api-gateway/src/gateway.ts
```diff
@@ -826,11 +826,13 @@ export class ApiGateway {
...
   protected readiness: RequestHandler = async (req, res) => {
     let health: 'HEALTH' | 'DOWN' = 'HEALTH';
-
+    
     if (this.standalone) {
       const orchestratorApi = await this.adapterApi({});
 
       try {
+        // todo: test other data sources
+        orchestratorApi.seenDatasources = { default: true };
```
ovr, line 835: It's better to use `orchestratorApi .addDataSeenSource('default');`

Weird about the moment when user is using multiple datasources without default.

packages/cubejs-server-core/src/core/OrchestratorApi.ts
```diff
@@ -1,20 +1,25 @@
...
-const pt = require('promise-timeout');
-const { QueryOrchestrator, ContinueWaitError } = require('@cubejs-backend/query-orchestrator');
+import pt from 'promise-timeout';
+import { QueryOrchestrator, ContinueWaitError } from '@cubejs-backend/query-orchestrator';
 
 export class OrchestratorApi {
-  constructor(driverFactory, logger, options) {
-    options = options || {};
-    this.options = options;
+  public seenDataSources: { [dataSource: string]: boolean };
```
ovr, line 6: Let's make it private, because we can modify it by public method - `addDataSeenSource `

packages/cubejs-server-core/src/core/OrchestratorApi.ts
```diff
@@ -1,20 +1,25 @@
...
+  
+  protected orchestrator: QueryOrchestrator;
```
ovr, line 8: protected readonly

packages/cubejs-server-core/src/core/OrchestratorApi.ts
```diff
@@ -1,20 +1,25 @@
...
+  
+  protected externalDriverFactory: any;
```
ovr, line 10: protected readonly

PR #1715, opened 2021-01-01
docs: add 'Dynamic Schema Creation' page

hassankhan:
**Check List**

Create a single page to collect all dynamic schema creation documentation.

**Note to reviewers:** I think we could probably add some examples of dynamic schema creation using the security context too; are there any other examples of interesting dynamic schema creation problems that are worth documenting?

### DIFF
--- a/docs/content/Schema/Dynamic-Schema-Creation.md
+++ b/docs/content/Schema/Dynamic-Schema-Creation.md
@@ -1,0 +1,268 @@
+---
+title: Dynamic Schema Creation
+permalink: /schema/dynamic-schema-creation
+category: Data Schema
+menuOrder: 2
+---
+
+Cube.js allows schemas to be created on-the-fly using a special
+[`asyncModule()`][ref-async-module] function only available in the [schema
+execution environment][ref-schema-env]. `asyncModule()` allows registering an
+async function to be executed at the end of the data schema compile phase so
+additional definitions can be added. This is often useful in situations where
+schema properties can be dynamically updated through an API, for example.
+
+<!-- prettier-ignore-start -->
+[[warning | Note]]
+| Each `asyncModule` call will be invoked only once per schema compilation.
+<!-- prettier-ignore-end -->
+
+[ref-schema-env]: /schema-execution-environment
+[ref-async-module]: /schema-execution-environment#asyncmodule
+
+When creating schemas via `asyncModule()`, it is important to be aware of the
+following differences compared to statically defining schemas with `cube()`:
+
+- The `sql` and `drillMembers` properties for both dimensions and measures must
+  be of type `() => string`
+
+Cube.js supports importing JavaScript logic from other files in a schema, so it
+is useful to declare utility functions for handling the above differences in a
+separate file:
+
+[ref-import-export]: /export-import
+
+```javascript
+// schema/utils.js
+export const convertStringPropToFunction = (propNames, dimensionDefinition) => {
+  let newResult = { ...dimensionDefinition };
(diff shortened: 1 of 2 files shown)


rpaik (approved):

docs/content/Schema/Dynamic-Schema-Creation.md
```diff
@@ -0,0 +1,122 @@
...
+  https://cube.dev/docs/schema-execution-environment#cube-js-globals-cube-and-others
+
+```javascript
+// schema/DynamicSchema.js
+const fetch = require('node-fetch');
+
+asyncModule(async () => {
+  const dynamicCubes = await (
+    await fetch('http://your-api-endpoint/dynamicCubes')
+  ).json();
```
keydunov, line 43: Can we show an example of how this JSON should look like? So users would know how to format `/dynamicCubes`API.

docs/content/Schema/Dynamic-Schema-Creation.md
```diff
@@ -0,0 +1,122 @@
...
+  const {
+    authInfo: { deploymentId },
+  } = COMPILE_CONTEXT;
+
+  const dynamicCubes = await (
+    await fetch(`http://your-api-endpoint/dynamicCubes`)
+  ).json();
+
+  const allowedDimensions = await (
+    await fetch(`http://your-api-endpoint/dynamicDimensions/${deploymentId}`)
```
keydunov, line 102: Should we use `tenantId` in the example here as well, same as before?

docs/content/Schema/Dynamic-Schema-Creation.md
```diff
@@ -0,0 +1,122 @@
+---
+title: Dynamic Schema Creation
+permalink: /schema/dynamic-schema-creation
+category: Data Schema
+menuOrder: 2
+---
+
+Cube.js allows schemas to be created on-the-fly using a special
+[`asyncModule()`][ref-async-module] function only available in the [schema
```
igorlukanin, line 9: It's really a very small detail but I'd like to ask whether we should add parentheses in the end of function names. I understand why we can do that but the drawback is that one might think that they are argument-less functions which is rarely true. WDYT, @hassankhan? (Anyway, regardless of the side you take :), it's worth documenting how we spell function names.)

hassankhan: @igorlukanin Great point, I've add parentheses mainly to distinguish functions from variables/properties. I usually try to link out to the function's documentation too where possible. Happy to go either way though 😀

keydunov: I believe it is a quite common notation for functions to display name with parentheses.

docs/content/Schema/Dynamic-Schema-Creation.md
```diff
@@ -0,0 +1,268 @@
...
+// schema/DynamicSchema.js
+const fetch = require('node-fetch');
+import {
+  convertStringPropToFunction,
+  transformDimensions,
+  transformMeasures,
+} from './utils';
+
+asyncModule(async () => {
+  const dynamicCubes = await (
```
ovr, line 97: @hassankhan Should we write somewhere pretty simple notice that JSON doesn't support serialization for functions?

hassankhan: @ovr I tried to explain that on lines 23-27, perhaps I could be clearer?

keydunov:
:shipit:

PR #1727, opened 2021-01-05
feat: Allow native booleans in filters, fix #1723

### LINKED ISSUE #1723: Boolean filter throws error
Hello guys,

Requesting the following filter throws: 
`Invalid query format: child "filters" fails because ["filters" at position 0 does not match any of the allowed types]`

Filter: 
```
{
  "member:" "MyCube.myBooleanDimension",
  "operator": "equals",
  "values": [true]
}
```

Found a workaround to use `'1' `and `'0' ` as values but it took me time to guess where was the issue from!

**Expected behavior**
Booleans in filters should work :)

**Version:**
[0.25.3]

Thanks guys!

ovr:
Hello!

refs https://github.com/cube-js/cube.js/issues/1723

Thanks

### DIFF
--- a/packages/cubejs-api-gateway/src/query.js
+++ b/packages/cubejs-api-gateway/src/query.js
@@ -69,7 +69,7 @@ const oneFilter = Joi.object().keys({
   dimension: id,
   member: id,
   operator: Joi.valid(...operators).required(),
-  values: Joi.array().items(Joi.string().allow('', null), Joi.link('...'))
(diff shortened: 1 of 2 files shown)


benswinburne:
Hi @ovr, do you know when this is planned to be merged? I came across #1723 and we've had to implement the numbers as strings to represent booleans "fix" to get around it i.e. `'1' = true '0' = false`

Cheers

PR #1791, opened 2021-01-15
fix(@cubejs-backend/api-gateway): readiness fix

vasilev-alex:


ovr (approved):
I continue to think that it's not a good solution to add default data source, but LGTM for now because We don't have another solutions

PR #1818, opened 2021-01-19
Allow to use HyperLogLog on top of precomputed sketches in CubeStore and CubeJS

ilya-biryukov:
See individual commits for more information.

This PR depends on the following  [arrow commit](https://github.com/cube-js/arrow/pull/5).
We add `merge()` and `cardinality()` UDFs to CubeStore and allow to use them from CubeJS on pre-aggregated data for `countDistinctApprox`.

Tested on Web Analystics Demo and various random sketches generated from Presto, both implementations produce exactly the same results on all datasets I tested.

More to come: HLL++ implementation and custom column type for sketches to allow early checks.


ilya-biryukov:
merged as part of https://github.com/cube-js/cube.js/pull/1827

PR #1877, opened 2021-01-25
docs: add warning to `orchestratorOptions` in 'Configuration' page

hassankhan:

Add a prominent warning to [`orchestratorOptions`](https://cube.dev/docs/config#options-reference-orchestrator-options) in the [Configuration page](https://cube.dev/docs/config).

### DIFF
--- a/docs/content/Configuration/Config.md
+++ b/docs/content/Configuration/Config.md
@@ -213,17 +213,21 @@ module.exports = {
 
 ### contextToOrchestratorId
 
-`contextToOrchestratorId` is a function to determine a caching key for Query Orchestrator instance.
-Query Orchestrator instance holds database connections, execution queues, pre-aggregation table caches.
-By default, returns the same value as `contextToAppId`.
+`contextToOrchestratorId` is a function to determine a caching key for Query
+Orchestrator instance. Query Orchestrator instance holds database connections,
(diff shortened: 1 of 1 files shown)


rpaik (approved):

keydunov:
:shipit:

docs/content/Configuration/Config.md
```diff
@@ -434,9 +434,12 @@ with options [from here][link-express-cors-opts].
 
 ### orchestratorOptions
 
-You can pass this object to set advanced options for Cube.js Query Orchestrator.
+<!-- prettier-ignore-start -->
+[[warning | ]]
+| We **strongly** recommend leaving these options set to the defaults. Only **advanced** users should tweak these.
```
igorlukanin, line 439: IDK if this is an actionable and effective advice. How does one decide if they're *advanced enough*? What are the implications of setting these options to unsound values?

My suggestion is that we might replace these with the answer to the second question, and remove the "advanced" thing entirely.

hassankhan: That's a great point, @igorlukanin 

```suggestion
<!-- prettier-ignore-start -->
[[warning | ]]
| We **strongly** recommend leaving these options set to the defaults. Changing these values can result in application instability and/or downtime.
```

PR #1906, opened 2021-01-27
feat: Ability to specify dataSource from request

ovr:
Hello!

Thanks

### DIFF
--- a/packages/cubejs-server-core/src/core/server.ts
+++ b/packages/cubejs-server-core/src/core/server.ts
@@ -25,6 +25,7 @@ import type {
   OrchestratorOptionsFn,
   PreAggregationsSchemaFn,
   RequestContext,
+  DriverContext,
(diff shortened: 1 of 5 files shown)


packages/cubejs-server-core/src/core/server.ts
```diff
@@ -549,9 +550,9 @@ export class CubejsServerCore {
     return scheduler.runScheduledRefresh(context, queryingOptions);
   }
 
-  public async getDriver() {
+  public async getDriver(ctx: DriverContext) {
     if (!this.driver) {
-      const driver = this.options.driverFactory(<any>{});
+      const driver = this.options.driverFactory(ctx);
```
ovr, line 555: Now We will not cast `{}` as `DriverContext` which is awful

vasilev-alex (approved):

PR #1911, opened 2021-01-27
fix(@cubejs-client/core): propagate time dimension to the drill down query

### LINKED ISSUE #1737: Parent timeDimensions not automatically propagating to drillDown
**Describe the bug**
resultSet.drillDown method does not automatically apply timeDimensions from the parent query to the drill down query

**To Reproduce**
Exactly the same issue as what we saw in (https://github.com/cube-js/cube.js/issues/1138) but in this case timeDimensions.

**Expected behavior**
resultSet.drillDown method applies the timeDimensions from the parent query to the drill down query as well.

vasilev-alex:
Closes #1737

### DIFF
--- a/packages/cubejs-client-core/src/ResultSet.js
+++ b/packages/cubejs-client-core/src/ResultSet.js
@@ -154,14 +154,23 @@ class ResultSet {
           });
         }
       });
-
+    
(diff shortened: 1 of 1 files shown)


ovr (approved):
LGTM

PR #1920, opened 2021-01-28
docs(configuration): Describe scheduledRefreshTimeZones, scheduledRef…

ovr:
Hello!

Thanks

### DIFF
--- a/docs/content/Configuration/Config.md
+++ b/docs/content/Configuration/Config.md
@@ -390,13 +390,17 @@ module.exports = {
 
 ### scheduledRefreshContexts
 
-If you are using `authInfo` inside `contextToAppId` or `contextToOrchestratorId` without `scheduledRefreshContexts`, 
-you will get `undefined`, because it's impossible to automatically generate context for it.
+If you are using `securityContext` inside `contextToAppId` or
+`contextToOrchestratorId`, you must also set up  `scheduledRefreshContexts`.
+This allows Cube.js to generate the necessary security contexts prior to
+running the scheduled refreshes.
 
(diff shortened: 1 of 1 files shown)


docs/content/Configuration/Config.md
```diff
@@ -360,12 +362,63 @@ module.exports = {
...
+
+### scheduledRefreshContexts
+
+If you are using `authInfo` inside `contextToAppId` or `contextToOrchestratorId` without `scheduledRefreshContexts`, 
+you will get `undefined`, because it's impossible to automatically generate context for it.
+
+To help Cube.js know about contexts, you should declare `scheduledRefreshContexts` to explain, what contexts
+are needed to be executed.
+
+You can declare contexts:
```
hassankhan, line 399:
```suggestion
If you are using `securityContext` inside `contextToAppId` or
`contextToOrchestratorId`, you must also set up  `scheduledRefreshContexts`.
This allows Cube.js to generate the necessary security contexts prior to
running the scheduled refreshes.

<!-- prettier-ignore-start -->
[[warning |]]
| Leaving `scheduledRefreshContexts` unconfigured will lead to issues where the
| security context will be `undefined`. This is because there is no way for
| Cube.js to know how to generate a context without the required input.
<!-- prettier-ignore-end -->
```

An example of a configured `scheduledRefreshContexts` is below:

hassankhan (changes requested):
Mostly minor things, but I did rework the `scheduledRefreshContexts` section a little bit. Let me know what you think @ovr :smile:

PR #1923, opened 2021-01-28
docs(multitenancy): Add Warning about scheduledRefreshContexts

ovr:
Hello!

refs https://github.com/cube-js/cube.js/issues/1904

Thanks

### DIFF
--- a/docs/content/Configuration/Config.md
+++ b/docs/content/Configuration/Config.md
@@ -371,12 +371,16 @@ module.exports = {
 };
 ```
 
+Learn more about [scheduled refreshes here](caching#keeping-cache-up-to-date).
+
+Best practice is to run `scheduledRefreshTimer` in a separate worker Cube.js
+instance. For serverless deployments, [REST API](rest-api#api-reference-v-1-run-scheduled-refresh) should be used instead.
+
+You may also need to configure [`scheduledRefreshTimeZones`](#options-reference-scheduled-refresh-time-zones) and [`scheduledRefreshContexts`](#options-reference-scheduled-refresh-contexts).
 ### scheduledRefreshTimeZones
 
-All time-based calculations performed within Cube.js are timezone-aware. Using this property you can specify multiple timezones in 
+All time-based calculations performed within Cube.js are timezone-aware. Using this property you can specify multiple timezones in
 [TZ Database Name](https://en.wikipedia.org/wiki/Tz_database) format e.g. `America/Los_Angeles`. The default value is `UTC`.
-
-This configuration option can be also set using the `CUBEJS_SCHEDULED_REFRESH_TIMEZONES` environment variable.
 
 ```javascript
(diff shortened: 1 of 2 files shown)


docs/content/Configuration/Multitenancy-Setup.md
```diff
@@ -20,6 +20,13 @@ The options are:
 - `preAggregationsSchema`
 - `queryTransformer`
 
+<!-- prettier-ignore-start -->
+[[warning |]]
+| In the [scheduled refresh](caching#pre-aggregations-background-refresh) leaving [`scheduledRefreshContexts`](config#options-reference-scheduled-refresh-contexts)  
+| unconfigured will lead to issues where the security context will be `undefined`. This is because there is no way for
+| Cube.js to know how to generate a context without the required input.
+<!-- prettier-ignore-end -->
+
```
hassankhan, line 29: Not sure about the placement of this - I think we should create a section in this (Multitenancy) page specifically for scheduled refreshes, WDYT @keydunov @ovr?

docs/content/Configuration/Config.md
```diff
@@ -371,13 +371,22 @@ module.exports = {
...
+
+Best practice is to run `scheduledRefreshTimer` in a separate worker Cube.js
+instance. For serverless deployments [REST API](rest-api#api-reference-v-1-run-scheduled-refresh) should be used
+instead of timer.
+
+There are some options that are important for the scheduled refresh:
+
+- [`scheduledRefreshTimeZones`](#options-reference-scheduled-refresh-time-zones)
+- [`scheduledRefreshContexts`](#options-reference-scheduled-refresh-contexts)
+
```
hassankhan, line 384:
```suggestion
You may also need to configure [`scheduledRefreshTimeZones`](#options-reference-scheduled-refresh-time-zones) and [`scheduledRefreshContexts`](#options-reference-scheduled-refresh-contexts).
```

hassankhan (changes requested):
Minor changes 😀 The only "problem" is the edits to Multitenancy - I agree we need to update the docs, but I think it should be more than a warning.

docs/content/Configuration/Multitenancy-Setup.md
ovr: We have the issue: https://github.com/cube-js/cube.js/issues/1904
Where users don't know about that they need to use `scheduledRefreshContexts` with multitenancy. It's why I added a reference as warning block to indicate that it's important.

hassankhan: @ovr I totally agree we should keep the warning 😄 I just meant maybe we should also make a 'Scheduled Refresh' section on this page and then move the warning there. The current placement of the warning seems a bit disconnected from the rest of the text around it.

hassankhan (approved):

PR #1939, opened 2021-01-29
chore: stage message fix

vasilev-alex:


ovr (approved):
I dont know original context why we deprecate this message, but LGTM

PR #1973, opened 2021-02-02
Update server.ts

rongfengliang:
change fetch scheduledRefreshTimer logic,  first fetch from env and then from code  and  `undefined==null` is true so for dev env scheduledRefreshTimer will awalys running


ovr:
Hello @rongfengliang,

Thank you for submitting a PR, but I am going to close this PR in favor of https://github.com/cube-js/cube.js/pull/1975 (which cleanup codebase)

Thanks

PR #1979, opened 2021-02-02
docs(DEPRECATION.md): improve examples for `authInfo` deprecation

hassankhan:

Improve `authInfo` deprecation examples as per conversation [from Slack](https://www.notion.so/Improve-deprecation-doc-6ab921d655cd4464b8313fa9bfdcb215#9f3d1f79355a4b3f8d688b39cb29d2ca).

### DIFF
--- a/DEPRECATION.md
+++ b/DEPRECATION.md
@@ -222,17 +222,37 @@ cube(`visitors`, {
 
 The `authInfo` parameter to `checkAuth` no longer wraps the decoded JWT under
 the `u` property. It has also been renamed to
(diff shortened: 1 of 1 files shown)


keydunov:
Looks good. Maybe would worth adding also something like 

```
// Old way
{
  "sub": "1234567890",
  "u": {
    "user_id": 131
  }
}
// New way
{
  "sub": "1234567890",
   "user_id": 131
}
```

So users would have clear understanding that "u" thing has changed not only in `checkAuth` but globally.

keydunov:
:shipit:

PR #1987, opened 2021-02-04
feat: Block from uploading files and folders (recur) starting from "."

Levhav:
**Check List**

 Block from uploading files and folders (recur) starting from "."


ovr (approved):
LGTM

PR #1996, opened 2021-02-05
fix(server-core): add support for async driverFactory

rsbh:
This will fix the issue of not able to pass async driverFactory in config.


ovr:
Hello @rsbh ,

Good catch 👍  Thank you!

PR #2000, opened 2021-02-08
limit size is too large  change to 200 step to 100

rongfengliang:
default limit size is too large ,change to 200 and step to 100 https://github.com/cube-js/cube.js/issues/1999


vasilev-alex:
@rongfengliang Thanks for the PR! Just discussed it with my teammates and we concluded that `200` limit is way too small as may not be enough when you request time-series data.

So, let's make `5000` by default

rongfengliang:
@vasilev-alex i  change it to 5000 and step to 500

PR #2009, opened 2021-02-08
docs: add healthcheck section to Production Checklist

hassankhan:

Add health check section to Production Checklist and backreferences to the REST API page


rpaik (approved):

docs/content/Deployment/Production-Checklist.md
```diff
@@ -144,3 +144,12 @@ server {
   }
 }
 ```
+
+## Set up health checks
```
igorlukanin, line 148: @hassankhan I think there's more context to these conventional API methods :) AFAIK they come from k8s world, right? Maybe we can have a relevant link here? Also, if there's a popular observability tool (Prometheus? Something AWS/GCP-specific?) which has a guide on working with these endpoints, can we reference it, too?

igorlukanin (approved):

docs/content/Deployment/Production-Checklist.md
hassankhan: Definitely agree on referencing the K8S docs, not too sure about promoting another tool but perhaps we can address that in a future PR?

PR #2026, opened 2021-02-09
fix(api-gateway): bind custom checkAuth functions

TRManderson:
**Check List**

**Description of Changes Made (issue reference is not provided)**

These functions were previously bound so that the API Gateway context could be accessed by `this` (useful for checkAuthFn when trying to use `enforceSecurityChecks` and `apiSecret`. This restores that functionality.

### DIFF
--- a/packages/cubejs-api-gateway/src/gateway.ts
+++ b/packages/cubejs-api-gateway/src/gateway.ts
@@ -736,7 +736,7 @@ export class ApiGateway {
     let showWarningAboutNotObject = false;
 
     return (req, res, next) => {
-      fn(req, res, (e) => {
+      fn.bind(this)(req, res, (e) => {
         // We renamed authInfo to securityContext, but users can continue to use both ways
         if (req.securityContext && !req.authInfo) {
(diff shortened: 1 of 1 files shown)


vasilev-alex (approved):

ovr (commented):
I don't think that it's a good idea to bind these functions to `ApiGateway`, because `ApiGateway` is an internal API that can be changed at any time without BC.

Still, there is a problem that we don't pass any options inside `checkAuth` fn, which makes it's impossible to understand what configuration should be used.

It's better to wait for https://github.com/cube-js/cube.js/pull/1962 and introduce a new argument called `options` as 3rd parameter in the `checkAuth` function.

cC @paveltiunov @keydunov

rpaik:
👋  @TRManderson, as noted above we're still working on `checkAuth` as part of our [Q1 epic](https://github.com/cube-js/cube.js/issues/1738). 

We'll close this PR for now, but feel free to check back with us and open a new issue/PR. Thanks for your contribution 🙇

PR #2054, opened 2021-02-11
Add HTTPS and readOnly support to ClickHouse driver

igorlukanin:
**Added HTTPS support to ClickHouse driver.**

Now it's possible to connect to a HTTPS-enabled ClickHouse database when `CUBEJS_DB_SSL` is set to `true`.

See https://github.com/apla/node-clickhouse#options (the "protocol" option) for implementation details.

**Added read-only support to ClickHouse driver.**

Now it's possible to connect to a ClickHouse server when a user's permissions are restricted to read-only when new config flag `CUBEJS_DB_CLICKHOUSE_READONLY` is set to `true`.

See https://clickhouse.tech/docs/en/operations/settings/permissions-for-queries/#settings_readonly and https://github.com/apla/node-clickhouse#options (the "readonly" option) for implementation details.

---

Documentation is updated in relevant parts.

### DIFF
--- a/packages/cubejs-clickhouse-driver/driver/ClickHouseDriver.js
+++ b/packages/cubejs-clickhouse-driver/driver/ClickHouseDriver.js
@@ -11,6 +11,7 @@ class ClickHouseDriver extends BaseDriver {
       host: process.env.CUBEJS_DB_HOST,
       port: process.env.CUBEJS_DB_PORT,
       auth: process.env.CUBEJS_DB_USER || process.env.CUBEJS_DB_PASS ? `${process.env.CUBEJS_DB_USER}:${process.env.CUBEJS_DB_PASS}` : '',
+      protocol: process.env.CUBEJS_DB_SSL ? 'https:' : 'http:',
       queryOptions: {
         database: process.env.CUBEJS_DB_NAME || config && config.database || 'default'
       },
--- a/packages/cubejs-clickhouse-driver/driver/ClickHouseDriver.js
+++ b/packages/cubejs-clickhouse-driver/driver/ClickHouseDriver.js
@@ -81,6 +81,10 @@ class ClickHouseDriver extends BaseDriver {
     return this.query("SELECT 1");
   }
 
+  readOnly() {
+    return !!this.config.readOnly || this.readOnlyMode;
+  }
+
   query(query, values) {
     const formattedQuery = sqlstring.format(query, values);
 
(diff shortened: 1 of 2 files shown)


hassankhan (approved):
LGTM 👍

rpaik (approved):

packages/cubejs-clickhouse-driver/driver/ClickHouseDriver.js
```diff
@@ -11,6 +11,8 @@ class ClickHouseDriver extends BaseDriver {
       host: process.env.CUBEJS_DB_HOST,
       port: process.env.CUBEJS_DB_PORT,
       auth: process.env.CUBEJS_DB_USER || process.env.CUBEJS_DB_PASS ? `${process.env.CUBEJS_DB_USER}:${process.env.CUBEJS_DB_PASS}` : '',
+      protocol: process.env.CUBEJS_DB_SSL ? 'https:' : 'http:',
+      readonly: Boolean(process.env.CUBEJS_DB_CLICKHOUSE_READONLY),
```
ovr, line 15: I dont think that We should introduce a new env variable called `CUBEJS_DB_CLICKHOUSE_READONLY`, because it's not universal for all drivers.

packages/cubejs-clickhouse-driver/driver/ClickHouseDriver.js
```diff
@@ -20,7 +22,14 @@ class ClickHouseDriver extends BaseDriver {
...
         queryOptions: {
-          join_use_nulls: 1,
+          //
+          //
+          // If ClickHouse user's permissions are restricted with "readonly = 1",
+          // change settings queries are not allowed. Thus, "join_use_nulls" setting
+          // can not be changed
+          //
+          //
+          ...(this.config.readonly ? {} : { join_use_nulls: 1 }),
```
ovr, line 32: @igorlukanin I think there is a misunderstanding about read-only flag for connections and `readOnly` option.

inside Cube.js option called `readOnly` is used to choose another strategy for  pre-aggregation creation. (pre-agg will be done without creating table and type detection will be done from db's response which has limitations - https://cube.dev/docs/pre-aggregations#read-only-data-source-pre-aggregations-limitations)

If you are trying to get another pre-aggregation strategy ☝️ , you should implement readOnly method

```js
  readOnly() {
    return !!this.config.readOnly;
  }
```

And in this case I think we should not do read-only connection.

Thanks

packages/cubejs-clickhouse-driver/driver/ClickHouseDriver.js
igorlukanin: Let me explain! In ClickHouse, there's this [readonly](https://clickhouse.tech/docs/en/operations/settings/permissions-for-queries/#settings_readonly) option which defined the restrictions on the user's permissions:

> 0 — All queries are allowed.
> 1 — Only read data queries are allowed.
> 2 — Read data and change settings queries are allowed.

So, it makes sense to introduce a new env variable called `CUBEJS_DB_CLICKHOUSE_READONLY` to reflect this feature of ClickHouse. As you can see in the [docs](https://github.com/apla/node-clickhouse#options) for the library we use to connect to ClickHouse, they also introduce this kind of config option.

packages/cubejs-clickhouse-driver/driver/ClickHouseDriver.js
igorlukanin: Sure, you're right. I'm not talking about pre-aggregations here.

igorlukanin:
@ovr @paveltiunov I've updated the code according to your suggestions. Are we okay to merge now? :)

PR #2095, opened 2021-02-14
docs: change code of boolean logical operators section example

herculesgabriel:

A code example of Boolean logical operators doesn't work.
This PR makes the first condition from "and" logical operator different than the first condition from "or" logical operator. I only changed the value inside the "values" array, but it could be something else. I don't think it must be more descriptive than that, though.


hassankhan (approved):
LGTM, thanks @herculesgabriel 🙌

PR #2109, opened 2021-02-16
docs: add an example of multi-line environment variable for SSL connections

hassankhan:
Adding an example `.env` file with a multi-line SSL certificate. Based on testing from https://github.com/cube-js/cube.js/pull/2072.


hassankhan:
Duplicate of https://github.com/cube-js/cube.js/pull/2010 😅

PR #2139, opened 2021-02-18
fix(@cubejs-client/core): manage boolean filters for missing members

### LINKED ISSUE #2097: How do i get logical operators to work in the client library
**Problem**

I'm experiencing an uncaught exception using the cubejs-client/core package whenever I try to use logical operators in my filters. Below is the json query passed to the QueryBuilder object.  The error returned is in the resolveMember method:

```
Cannot read property 'split' of undefined
```

which comes from it thinking the memberName is undefined.  I'm assuming that this has to do with my JSON structure, but I've tried every permutation of structure I can think of.  Any advice would be greatly appreciated.
 

**Related Cube.js schema**
```javascript
{
	"chartType": "number",
	"query": {
		"measures": ["PlayerEvents.sumEvents"],
		"timeDimensions": [],
		"filters": [{
			"and": [{
				"member": "PlayerEvents.eventCategory",
				"operator": "equals",
				"values": ["Video"]
			}, {
				"member": "PlayerEvents.eventAction",
				"operator": "equals",
				"values": ["content_play"]
			}, {
				"or": [{
					"member": "PlayerEvents.contentProviderAccountId",
					"operator": "equals",
					"values": ["2"]
				}, {
					"member": "PlayerEvents.placementAccountId",
					"operator": "equals",
					"values": ["2"]
				}, {
					"member": "PlayerEvents.contentProviderAccountId",
					"operator": "equals",
					"values": ["2"]
				}]
			}]
		}]
	}
}
```

**Related Cube.js generated SQL**

```sql
SELECT * FROM ...
```

vasilev-alex:
Closes #2097

### DIFF
--- a/packages/cubejs-client-core/src/utils.js
+++ b/packages/cubejs-client-core/src/utils.js
@@ -5,15 +5,15 @@ export function defaultOrder(query) {
 
   if (granularity) {
     return {
-      [granularity.dimension]: 'asc'
+      [granularity.dimension]: 'asc',
     };
   } else if ((query.measures || []).length > 0 && (query.dimensions || []).length > 0) {
     return {
-      [query.measures[0]]: 'desc'
+      [query.measures[0]]: 'desc',
     };
   } else if ((query.dimensions || []).length > 0) {
     return {
-      [query.dimensions[0]]: 'asc'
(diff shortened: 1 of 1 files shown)


ovr (approved):
LGTM

PR #2166, opened 2021-02-19
examples(auth0): fix example

hassankhan:

Issue discovered during https://github.com/cube-js/cube.js/pull/2143.


ovr:
@hassankhan I think it's better to introduce an env variable called `REACT_APP_AUTH0_REDIRECT_URI` and define it for production example by `.env.production`, next

```
`redirectUri={proccess.env.REACT_APP_AUTH0_REDIRECT_URI || window.location.origin}`
```

What do you think? Thanks

PR #2196, opened 2021-02-23
React client TypeScript fixes, QueryBuilder improvements

SteffeyDev:
**Check List**

### TypeScript type updates

I tried using cube.js react client in a typescript project, and found a number of issues with the provided typescript types.  In many cases, the types were not specific enough to be used in a strict typescript project, but in others the types were simply incorrect.  The included changes were necessary for me to use it in my project, where I use a I have created QueryBuilder wrapper similar to that in the playground, but strongly typed.  I have also modified the types to support the QueryBuilder changes I made, as described below.

#### Convert to uncontrolled

I ran into some issues while making other improvements, and realized that it was because the QueryBuilder was neither fully controller nor uncontrolled, and was using the React anti-pattern of [unconditionally copying props to state](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#anti-pattern-unconditionally-copying-props-to-state).  Following the recommendations in that article and based on my experience as a React developer, I decided to make QueryBuilder uncontrolled.  This means that QueryBuilder controls it's own state, and can't be modified from the outside save for in specific ways that are allowed through props.

This involved removing the `query`, `setQuery`, `vizState`, and `setVizState` props, and adding the `initialVizState` and `onVizStateChanged` props.  From my analysis, the most common use case for accessing the `vizState` outside of the query editor is to save the editor state in persistent storage.  These two new props allow for setting up a QueryBuilder component from a saved state and then listening for changes to save back.  I'm using this in my app to save the `vizState` to a database so I can persist the QueryBuilder state across reloads.

I also added the `defaultQuery` and `defaultChartType` props, so that if `initialVizState` is empty, it will pick up the default values for those. `defaultChartType` was listed in the documentation before, but wasn't actually implemented until now.

#### Refactor `orderMembers`

Another issue I ran into was the interop between `query.order` and `orderMembers`.  Whenever the query changed, `orderMembers` potentially needed to be updated, and every time `orderMembers` changed, the query might need to be updated.  This relationship made reasoning about the various state changes difficult, and especially saving/restoring state and reloads.

I refactored the QueryBuilder so that `orderMembers` is calculated on demand in then `render` method based on the `availableDimensions`, `availableMeasures`, and `query.order`.  Then, I changed `updateOrder` so that it just modified the `query.order` and sent it through the normal query update path.  This makes `query.order` the single source of truth, and removes any difficulties with maintaining the relationship between `query.order` and `orderMembers`.

#### Add `error` render prop

I noticed that if `dryRun` failed, the error was logged to the console but not shown to the user.  Instead of ignoring and letting the `query` call report the same error, I changed it so that `dryRun` errors are reported to the rendered function in the `error` prop.  This is mostly useful when `wrapWithQueryRenderer` is false, as when it is true, the QueryRenderer will produce the same error and overwrite with it's own `error` prop (by design).

#### Add PropTypes

I added PropTypes so that non-typescript users can enjoy the benefit of prop type checking as well.  This should be a lot safer and prevent unexpected props from being passed in.'

### Summary

I know this is a lot, but I believe it makes the QueryBuilder a lot simpler to reason about, easier for fluent React developers like myself to pick it up and use it, and easier for future open source contributors to understand and add to the code.

I did not update the example React dashboard app and other docs yet with these changes, as I wanted to get approval and resolve all issues before I spent the time updating those.  Once this is approved, I will go through and update the docs and examples to support these new props, and make sure the example react dashboard app still works.

As always, I'm hoping for good discussions and will readily make changes to address all concerns.  I look forward to working with the cube.js development team on these changes, and if it goes well, would most likely invest more time into the project in the future.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,5 @@ yarn-error.log
 **/coverage
 lerna-debug.log
 tsconfig.tsbuildinfo
+**/*.swo
+**/*.swp
--- a/packages/cubejs-client-core/index.d.ts
+++ b/packages/cubejs-client-core/index.d.ts
@@ -641,6 +641,7 @@ declare module '@cubejs-client/core' {
      */
     tableColumns(pivotConfig?: PivotConfig): TableColumn[];
 
+    totalRow(): ChartPivotRow;
     query(): Query;
     rawData(): T[];
     annotation(): QueryAnnotations;
--- a/packages/cubejs-client-react/index.d.ts
+++ b/packages/cubejs-client-react/index.d.ts
@@ -5,16 +5,18 @@ import {
   ResultSet,
   Filter,
   PivotConfig,
-  MemberType,
   TCubeMeasure,
   TCubeDimension,
-  TCubeMember,
+  TCubeSegment,
+  TimeDimension,
   ProgressResponse,
   TDryRunResponse,
   TOrderMember,
   QueryOrder,
   TQueryOrderArray,
   TSourceAxis,
+  TimeDimensionComparison,
+  TimeDimensionRanged,
 } from '@cubejs-client/core';
 
 /**
@@ -141,9 +143,8 @@ declare module '@cubejs-client/react' {
   type ChartType = 'line' | 'bar' | 'table' | 'area' | 'number' | 'pie';
 
   type VizState = {
-    [key: string]: any;
+    query?: Query;
     pivotConfig?: PivotConfig;
-    shouldApplyHeuristicOrder?: boolean;
     chartType?: ChartType;
   };
 
@@ -155,12 +156,12 @@ declare module '@cubejs-client/react' {
     /**
      * Default query
      */
-    query?: Query;
-    vizState?: VizState;
+    defaultQuery?: Query;
+    defaultChartType?: ChartType;
+    initialVizState?: VizState;
     /**
      * @default defaultChartType line
      */
-    defaultChartType?: ChartType;
     /**
      * Defaults to `false`. This means that the default heuristics will be applied. For example: when the query is empty and you select a measure that has a default time dimension it will be pushed to the query.
      * @default disableHeuristics false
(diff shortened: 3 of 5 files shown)


SteffeyDev:
Not sure why that integration test failed, but obviously nothing to do with this PR

SteffeyDev:
I'll resolve conflicts once the PR is reviewed

vasilev-alex (changes requested):
Hi @SteffeyDev! Thanks for a great PR!

As I can see it introduces breaking changes which we cannot accept. However, we're ok with deprecating things but keeping them backward compatible for now.

Also, maybe you could consider splitting the PR? It'll make it easier to review and accept the changes.

As for **PropTypes** not sure if need it as we'll end up maintaining both `index.d.ts` and `prop-types` definitions. Moreover, all modern IDEs will take the TS definitions we have into account even for non-TS projects. 

What do you think?

SteffeyDev:
Why can you not accept breaking changes? Maybe not immediately, but into the next major version.  As I explained the QueryBuilder API has some design issues, and thus breaking changes are necessary to get it inline with best practices and developer expectations.  Unfortunately, trying to support fully controlled props in an uncontrolled component is near impossible, and would defeat the purpose of making it uncontrolled in the first place, which I explained above.

There's a few options to move forward, but it really depends on what your schedule is for major versions where breaking changes can happen.  Let me know where you guys are at on that and we can go from there.

SteffeyDev:
As for PropTypes, I think it's just a safer option, but definitely not needed.  If you don't include them, there's always a risk that someone using an older editor might pass in invalid options, which causes internal issues and then that user gets frustrated because it looks like it is the component's fault, not theirs.  Up to you as a maintainer.

SteffeyDev:
After giving it some thought, I think I've found a way to make it backwards compatible with minimal consequences.  I could get to work on that.  While I'm doing that, I might also go ahead and split this into a few PRs.  It will take a lot of time, but I understand why.

Here would be the 5 resulting PRs:
* Typescript type fixes
* Convert QueryBuilder to uncontrolled
* Refactor `orderMembers`
* Add `queryError` handling from `dryRun`
* Add PropTypes to QueryBuilder

Before I spend all the time on these, are any of these PRs that you would reject or have no interest in? I'll only spend the time on PRs that will be accepted (potentially after discussion and modifications).

vasilev-alex:
@SteffeyDev, we're definitely willing to accept all of the PRs. But again, the main concern is that all changes to the API must be backward compatible.

Thanks for your time!

SteffeyDev:
@vasilev-alex I created #2261, #2262, and #2263.  Do you want me to create a PR for adding PropTypes, or do you not want to have the burden of maintaining PropTypes? Makes little difference to me.

Once those are merged in, I'll work on the bigger PR, which is converting to uncontrolled.  That needs to build on top of these three.

vasilev-alex:
@SteffeyDev nice! Let's go without PropTypes for now.

PR #2198, opened 2021-02-23
Add Codecov badge to README

thomasrockhu:
**Check List**

N/A

Tom from Codecov here. I thought I would go ahead and add the Codecov badge here in case it's useful!


rpaik:
@igorlukanin can you help with the review?

hassankhan (commented):
Just wondering if we should add the badge to all the package READMEs to be consistent?

/cc @keydunov

igorlukanin:
@thomasrockhu Thank you! Codecov is great, though I think that we need to configure the coverage across our entire mono repo first, so the info shown on the badge makes sense. For that, community contributions are very much welcome.

thomasrockhu:
@igorlukanin, would the ideal situation be the same badge added to all directories under `packages`?

igorlukanin:
@thomasrockhu Yes, that would do a part of the job. We also need to configure the coverage for every package properly.

PR #2200, opened 2021-02-24
docs(auth0): fix typos in guide

hassankhan:
Story details: https://app.clubhouse.io/cube-dev/story/194


rpaik (approved):

docs/content/Auth/Auth0-Guide.md
```diff
@@ -77,7 +84,7 @@ Cube.js.
 ```javascript
 function (user, context, callback) {
   const namespace = "http://localhost:4000/";
-  context.accessToken[namespace] =
+  context.idToken[namespace] =
```
keydunov, line 87: @hassankhan @ovr Just want to make sure it is correct. We're having `idToken` in one example and `accessToken` in different.

hassankhan: @keydunov I'd incorrectly used `accessToken` before; it should be `idToken` for both examples. Tested and verified using the OpenID Connect Playground:

image

ovr: We should use `accessToken`, because `idToken` is used for OpenID Connect spec and looks likes it's deprecated and will be removed.

hassankhan: I've just tested this using [the Auth0 example](https://github.com/cube-js/cube.js/tree/master/examples/auth0) and the custom claims only seem to work when added to `idToken`, **not** `accessToken`.

PR #2250, opened 2021-03-02
feat(server-core): add support for async `dbType()`

rsbh:
**Check List**

The current type for the `dbType` option in cube config is `string | sync function` and as per [documentation](https://cube.dev/docs/config#options-reference-db-type), it is used to dynamically select a database type depending on the user's context. but due to its sync type, we cant make any async call (eg. Network call) in it.

The `driverFactory` is already async type so it will be better to change `dbType` to async function.

### DIFF
--- a/packages/cubejs-server-core/src/core/CompilerApi.js
+++ b/packages/cubejs-server-core/src/core/CompilerApi.js
@@ -119,13 +119,13 @@ export class CompilerApi {
     return createQuery(
       compilers,
       dbType, {
-      ...query,
-      dialectClass,
-      externalDialectClass: this.options.externalDialectClass,
-      externalDbType: this.options.externalDbType,
-      preAggregationsSchema: this.preAggregationsSchema,
-      allowUngroupedWithoutPrimaryKey: this.allowUngroupedWithoutPrimaryKey
-    }
+        ...query,
+        dialectClass,
+        externalDialectClass: this.options.externalDialectClass,
+        externalDbType: this.options.externalDbType,
+        preAggregationsSchema: this.preAggregationsSchema,
+        allowUngroupedWithoutPrimaryKey: this.allowUngroupedWithoutPrimaryKey
+      }
     );
   }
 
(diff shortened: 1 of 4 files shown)


ovr:
@rsbh 

First, Thank you for submitting the PR.

Cube.js provides guarantees for caching async functions from configuration inside the core because:

1. It's harder to implement caching inside user's configuration
2. Everyone thinks that caching will be done inside cube.js core

When we are going to introduce that `dbType` can be an `async` function, it requires us to implement caching for it (`dbType` should be called once with caching to the context and it must be denounced to protect multiple async calls in one time for it).

There are helpers for it inside `@cubejs-backend/shared`, but I suggest another plan.

1. Quickly solve your task by yet another small PR:

FYI: Cube.js defines `driverFactory` & `dialectFactory` by default:

```js
      driverFactory: () => typeof dbType === 'string' && CubejsServerCore.createDriver(dbType),
      dialectFactory: (ctx) => CubejsServerCore.lookupDriverClass(ctx.dbType).dialectClass &&
        CubejsServerCore.lookupDriverClass(ctx.dbType).dialectClass(),
```

I suggest you to do in your `cube.js`:

```
const dbTypeCache = memoize(async (ctx) => {
     // select your configuration
    const { dbType, credentials } = await //

    const driver = createDriver(dbType);

    return {
         driver,
         dialect: createDialectFromDriver(driver),
    }
});

export {
   driverFactory: async (ctx) => (await dbTypeCache(ctx)).driver,
   dialectFactory: async (ctx) => (await dbTypeCache(ctx)).dialect,
}
```

To do ☝️ 

- You should create a PR that export `createDriver` & `createDialectFromDriver` and mark this functions with comment `// @internal`
- Defines `dialectClass` for each driver that doesn't have it.


2. Finish with this PR by introducing caching & debouncing it to the context and more write more tests.

Thanks

rsbh:
Hey @ovr, I am closing the PR right now.
I have to read more of the codebase and understand the internals more.
Will raise another PR when I have a better solution.

PR #2256, opened 2021-03-02
chore(cubestore): warn if some operations run longer than expected

ilya-biryukov:
In hopes to find places that cause slowdown.

### DIFF
--- a/rust/cubestore/src/sys/malloc.rs
+++ b/rust/cubestore/src/sys/malloc.rs
@@ -1,3 +1,6 @@
+use crate::util::time_span::warn_long;
+use std::time::Duration;
+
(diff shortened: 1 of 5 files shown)


rust/cubestore/src/sys/malloc.rs
```diff
@@ -1,3 +1,6 @@
+use crate::util::time_span::warn_long;
+use flatbuffers::bitflags::_core::time::Duration;
```
ovr, line 2: It looks like that IDE create an import wrongly from `flatbuffers` instead of `std::time::Duration`
is it?

ilya-biryukov: Yep, thanks.

PR #2261, opened 2021-03-03
TypeScript fixes

SteffeyDev:
**Check List**

Broken off of #2196

I tried using cube.js react client in a typescript project, and found a number of issues with the provided typescript types. In many cases, the types were not specific enough to be used in a strict typescript project, but in others the types were simply incorrect. The included changes were necessary for me to use it in my project, where I use a I have created QueryBuilder wrapper similar to that in the playground, but strongly typed.

### DIFF
--- a/packages/cubejs-client-core/index.d.ts
+++ b/packages/cubejs-client-core/index.d.ts
@@ -34,10 +34,25 @@ declare module '@cubejs-client/core' {
    * @order 3
    */
   export class HttpTransport implements ITransport {
+    /**
+     * @hidden
+     */
     protected authorization: TransportOptions['authorization'];
+    /**
+     * @hidden
+     */
     protected apiUrl: TransportOptions['apiUrl'];
+    /**
+     * @hidden
+     */
     protected method: TransportOptions['method'];
+    /**
+     * @hidden
+     */
     protected headers: TransportOptions['headers'];
+    /**
+     * @hidden
+     */
     protected credentials: TransportOptions['credentials'];
     constructor(options: TransportOptions);
     request(method: string, params: any): () => Promise<any>;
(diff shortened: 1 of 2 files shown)


packages/cubejs-client-core/index.d.ts
```diff
@@ -22,17 +22,23 @@ declare module '@cubejs-client/core' {
...
-    request(method: string, params: any): () => Promise<void>;
+    request(method: string, params: Record<string, unknown>): () => Promise<void>;
   }
 
   /**
    * Default transport implementation.
    * @order 3
    */
   export class HttpTransport implements ITransport {
+    protected authorization: TransportOptions['authorization'];
```
vasilev-alex, line 37: could you please annotate each of these with the `@hidden` annotation?

packages/cubejs-client-react/index.d.ts
```diff
@@ -477,25 +483,47 @@ declare module '@cubejs-client/react' {
...
+
+  type FilterWithExtraFields = Omit<Filter, 'dimension'> & {
+    dimension: TCubeDimension | TCubeMeasure;
+    operators: { name: string; title: string }[];
+  };
+  type TimeDimensionWithExtraFields = Omit<TimeDimension, 'dimension'> & {
+    dimension: TCubeDimension & { granularities: { name: string; title: string }[] };
+  };
+
+  type MemberUpdater<T> = {
```
vasilev-alex, line 495: Please move this type to the bottom of the docs block above, as we use it to generate docs.

packages/cubejs-client-react/index.d.ts
```diff
@@ -477,25 +483,47 @@ declare module '@cubejs-client/react' {
...
+    add: (member: T) => void;
+    remove: (member: { index: number }) => void;
+    update: (member: { index: number }, updateWith: T) => void;
   };
+  type DimensionUpdater = MemberUpdater<TCubeDimension>;
+  type MeasureUpdater = MemberUpdater<TCubeMeasure>;
+  type SegmentUpdater = MemberUpdater<TCubeSegment>;
+  // Only require the fields that are actually used (otherwise fields like `operators` are required just to add/update)
+  type TimeDimensionUpdater = MemberUpdater<
```
vasilev-alex, line 504: Would it be possible to simplify this type? While it may be absolutely correct, we use them for docs generation (typedoc) and if we run it we'll get

```
TimeDimensionUpdater: MemberUpdater‹object | object & object›
```

It is ok to duplicate some types, as the main goal here is clear typedoc output.

vasilev-alex (changes requested):

packages/cubejs-client-react/index.d.ts
SteffeyDev: Got it, installing typedoc now so I can preview output and make it look pretty :)

SteffeyDev:
Ok, I improved typedoc output at the expensive of maintainability and conciseness, improving existing docs as well as docs for my changes.

PR #2263, opened 2021-03-03
Refactoring QueryBuilder orderMembers for better logic flow

SteffeyDev:
**Check List**

Broken off from #2196

In the React QueryBuilder, an issue I ran into was the interop between query.order and orderMembers. Whenever the query changed, orderMembers potentially needed to be updated, and every time orderMembers changed, the query might need to be updated. This relationship made reasoning about the various state changes difficult, and especially saving/restoring state and reloads.

I refactored the QueryBuilder so that orderMembers is calculated on demand in then render method based on the availableDimensions, availableMeasures, and query.order. Then, I changed updateOrder so that it just modified the query.order and sent it through the normal query update path. This makes query.order the single source of truth, and removes any difficulties with maintaining the relationship between query.order and orderMembers.

I confirmed in my own usage that the `orderMembers` prop that is put out is consistent with before, and the `updateOrder` prop also has the same API and works exactly the same.

### DIFF
--- a/packages/cubejs-client-react/src/QueryBuilder.jsx
+++ b/packages/cubejs-client-react/src/QueryBuilder.jsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { prop, uniqBy, indexBy, fromPairs, equals } from 'ramda';
+import { prop, uniqBy, equals } from 'ramda';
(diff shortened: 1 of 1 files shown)


vasilev-alex (approved):
Looks great!

PR #2281, opened 2021-03-04
misc(ci): Cache docs/node_modules + use only one cache

ovr:
Hello!

Problems:

- docs/node_modules are not cached
- 2 caches, which is slow for downloading and persistence after job

Before

image

- 9 minutes total (pretty random thing)
- 1 19 yarn cache
- 2 11 lerna cache

After

image

Thanks


vasilev-alex:
Awesome!

vasilev-alex (approved):

PR #2303, opened 2021-03-05
Refactoring React QueryBuilder to be uncontrolled

SteffeyDev:
**Check List**

Broken off from #2196 

I ran into some issues while making other changes to the React QueryBuilder, and realized that it was because the QueryBuilder was neither fully controller nor uncontrolled, and was using the React anti-pattern of [unconditionally copying props to state](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#anti-pattern-unconditionally-copying-props-to-state).  Following the recommendations in that article and based on my experience as a React developer, I decided to make QueryBuilder uncontrolled.  This means that QueryBuilder controls it's own state, and can't be modified from the outside save for in specific ways that are allowed through props.

This involved deprecating the `query`, `setQuery`, `vizState`, and `setVizState` props, and adding the `initialVizState` and `onVizStateChanged` props.  From my analysis, the most common use case for accessing the `vizState` outside of the query editor is to save the editor state in persistent storage.  These two new props allow for setting up a QueryBuilder component from a saved state and then listening for changes to save back.  I'm using this in my app to save the `vizState` to a database so I can persist the QueryBuilder state across reloads.

I also added the `defaultQuery` and `defaultChartType` props, so that if `initialVizState` is empty, it will pick up the default values for those. `defaultChartType` was listed in the documentation before, but wasn't actually implemented until now.

### DIFF
--- a/packages/cubejs-client-react/index.d.ts
+++ b/packages/cubejs-client-react/index.d.ts
@@ -157,14 +157,21 @@ declare module '@cubejs-client/react' {
      */
     cubejsApi: CubejsApi;
     /**
-     * Default query
+     * State for the QueryBuilder to start with. Pass in the value previously saved from onVizStateChanged to restore a session.
      */
-    query?: Query;
-    vizState?: VizState;
+    initialVizState?: VizState;
     /**
-     * @default defaultChartType line
+     * Called by the `QueryBuilder` when the viz state has changed. Use it to save state outside of the `QueryBuilder` component.
+     */
+    onVizStateChanged?: (vizState: VizState) => void;
+    /**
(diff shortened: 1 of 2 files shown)


DrewHoo:
> This involved deprecating the query, setQuery, vizState, and setVizState props

@SteffeyDev are those deprecated properties being tracked anywhere for when they'll be removed?

SteffeyDev:
I think I put comments in the code to mark them as deprecated, but nothing past that.

SteffeyDev:
They can't be removed until the next major version, but unless they are maintaining a list of things to remove I doubt the maintainers will remember to remove them when it comes time to release V1.

PR #2326, opened 2021-03-10
fix(cubestore): update arrow, fix test merge sort over unions

ilya-biryukov:


Levhav (commented):
I haven't worked with cubestore or Rust. Maybe someone else needs to do a review.

PR #2327, opened 2021-03-10
fix(cubestore): CubeTableExec always produces sorted output

ilya-biryukov:
By using `MergeSortExec` instead of `MergeExec`.
This allows more optimizations.

After this change, we use inplace aggregation instead of hash
aggregation for partitions with extra chunks.


ilya-biryukov:
This change makes things quite a bit slower if we do not need to aggregate.
https://github.com/cube-js/cube.js/pull/2330 takes care of only creating `MergeSortExec` when aggregations later consume it, we should land it instead.

PR #2330, opened 2021-03-10
fix(cubestore): use merge sort exec when aggregations are required

ilya-biryukov:
The new physical plan optimization aims to change the consumer of
`CubeTableExec` from `MergeExec` to `MergeSortExec`, but only if
results get consumed by aggregations.

This lets us use more efficient inplace aggregations in more cases.


ilya-biryukov:
We need to land https://github.com/cube-js/arrow/pull/20 first and update the arrow version here.

PR #2332, opened 2021-03-10
fix(cubestore): fix crash on empty results from workers

ilya-biryukov:


rust/cubestore/src/cluster/mod.rs
```diff
@@ -713,12 +714,12 @@ impl ClusterImpl {
         #[cfg(target_os = "windows")]
         {
             // TODO optimize for no double conversion
-            let res = SerializedRecordBatchStream::write(
-                self.query_executor
-                    .execute_worker_plan(plan_node.clone(), remote_to_local_names)
-                    .await?,
-            );
-
+            // TODO optimize for no double conversion
```
ilya-biryukov, line 717: Ooops, this is accidental, sorry. Feel free to amend or land like this and I'll fix this tomorrow.

PR #2351, opened 2021-03-12
Prevent calling onVizStateChanged twice unless needed

SteffeyDev:
**Check List**

Related to #2303 

Prevents `onVizStateChanged` from being called twice unnecessarily by doing an `equals` object comparison before calling the second time.

After this change, in all my testing, it was only called once.  I was unable to find a test case where the query or pivotConfig was changed after `dry-run`.  I'd be happy to test more if you can identify a sequence of actions that could lead to that.  Looking at the code, it _should_ run twice if changed, but I have not proven that.


vasilev-alex:
Thanks @SteffeyDev!

PR #2356, opened 2021-03-12
chore(cubestore): choose index during planning, not in serialized plan

ilya-biryukov:


ilya-biryukov:
Merged as part of another PR

PR #2371, opened 2021-03-15
Logging in typescript

jrylander:
Moved logging code in order to use it from redis code (I still experience spurious errors in our system). Pls check if logging code is to your liking and has not changed intent.


packages/cubejs-server-core/tsconfig.json
```diff
@@ -3,7 +3,7 @@
   "include": [
     "src",
     "test"
-  ],
+, "../cubejs-backend-shared/src/logger.js"  ],
```
ovr, line 6: Weird. Do you have a problem with `yarn link`?

packages/cubejs-server-core/yarn.lock
```diff
@@ -0,0 +1,7516 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
+
+
```
ovr, line 4: We migrated to yarn workspaces.

packages/cubejs-query-orchestrator/src/orchestrator/IORedisFactory.ts
```diff
@@ -77,6 +86,14 @@ export async function createIORedisClient(url: string, opts: RedisOptions): Prom
     ...options,
     ...opts,
   });
+
+  client.on('connect', () => logger('IORedis connected'));
```
ovr, line 90: These events can produce a lot of noise inside logs.

packages/cubejs-query-orchestrator/src/orchestrator/IORedisFactory.ts
```diff
@@ -1,10 +1,14 @@
 import Redis, { Redis as redis, RedisOptions } from 'ioredis';
-import { getEnv } from '@cubejs-backend/shared';
+import { getEnv, devLogger, prodLogger } from '@cubejs-backend/shared';
 import AsyncRedisClient from './AsyncRedisClient';
 import { parseRedisUrl } from './utils';
 
 export type IORedisOptions = RedisOptions;
 
+const logger = process.env.NODE_ENV !== 'production'
```
ovr, line 8: The logger should be re-used from the `server-core`, instead of creating a new one.

packages/cubejs-backend-shared/src/logger.ts
```diff
@@ -0,0 +1,132 @@
+import SqlString from 'sqlstring';
```
ovr, line 1: As I understand you moved the logger to `@cubejs-backend/shared` to create a new instance of it, but the logger should be re-used from the server-core, instead of creating a new one in each place where it should be used.

ovr:
Thank you @jrylander, for submitting the PR. I left some comments.

packages/cubejs-backend-shared/src/logger.ts
```diff
@@ -0,0 +1,132 @@
...
+import R from 'ramda';
+
+interface logParams {
```
vasilev-alex, line 4: Let's prefer PascalCase for interface names and keep each element on a separate line

packages/cubejs-server-core/tsconfig.json
jrylander: Ehum. Not sure what automatic something did that :D

packages/cubejs-server-core/yarn.lock
jrylander: Should all yarn.lock files go?

packages/cubejs-query-orchestrator/src/orchestrator/IORedisFactory.ts
jrylander: Only on trace level, right?

packages/cubejs-backend-shared/src/logger.ts
jrylander: Ah. Fell into the logging pattern we use where I work.

jrylander:
Hm. It is so terse to just write and I hope I do contribute more than mess. If I do mess more than contribute, please feel free to tell me and let me get better. I moved the logging because otherwise there is a circular dependency. We could just move the LoggingFn type but it felt strange to have that in another package from the implementation.

packages/cubejs-server-core/yarn.lock
jrylander: Never mind. Diff shows no change in yarn.lock files

packages/cubejs-query-orchestrator/src/orchestrator/IORedisFactory.ts
ovr: Right! But the current design of the logger doesn't allow it.

paveltiunov:
Hey @jrylander ! It seems like you're trying to debug IORedis in production. Could you please elaborate on what are the issues you have?

jrylander:
> Hey @jrylander ! It seems like you're trying to debug IORedis in production. Could you please elaborate on what are the issues you have?

Hi! Pls forgive me for late answer. I have not had time to continue my investigation but the error lies in the system not being able to reconnect when connection to Sentinel goes down and comes back.

paveltiunov:
@jrylander Gotcha. Most likely it's due to the pool doesn't do tests on borrow. We're working on this issue right now. You may want to try running without a pool and see if this issue persists for you.

jrylander:
I will pull back this since I do not have time to work on it. Tnx for all the feedback on both this and earlier pr!

PR #2378, opened 2021-03-16
Skip `refreshCubesRefreshKey` If A Cube Has No Pre-Aggregations

acham1:
**Check List**

N/A

Skip `refreshCubesRefreshKey` if a Cube has no pre-aggregations defined for it.

# Background

- I'm trying to trigger a scheduled refresh using the `/v1/run-scheduled-refresh` API endpoint
- I have two Cubes: one named `X` with `preAggregations` and one named `Y` without `preAggregations`
- I have a Big Query data source that requires a partition filter on all queries, so I reference that partition column `X.p` in the `timeDimensions` part of the API call's query options
- The Cube server emits an error `Error: Can't find join path to join 'Y', 'X'`, presumably because of the query option referencing a column from `X`

I am proposing this change to skip `refreshCubesRefreshKey` if a Cube's `preAggregations` field is empty. Thanks!


paveltiunov:
Hey @acham1 ! Could you please share the schema which causes the problem?

rpaik:
👋 @acham1 following-up on @paveltiunov's question 👆

PR #2413, opened 2021-03-22
refactor(cubejs-cli): cube-cloud methods as separate package

RusovDmitriy:


packages/cubejs-backend-cloud/src/deploy.ts
```diff
@@ -43,7 +43,7 @@ export class DeployDirectory {
     }
 
     // blacklist
-    if (['dashboard-app', 'node_modules'].includes(baseName)) {
+    if (['.git', 'dashboard-app', 'node_modules'].includes(baseName)) {
```
ovr, line 46: It already ignores all names started with by `return baseName.charAt(0) !== '.';`.

packages/cubejs-backend-cloud/tsconfig.json
```diff
@@ -0,0 +1,19 @@
...
+  "extends": "../../tsconfig.base.json",
+  "include": [
+    "src",
+    "test"
+  ],
+  "compilerOptions": {
+    "outDir": "dist",
+    "rootDir": ".",
+    "baseUrl": ".",
+    "allowJs": true,
```
ovr, line 11: I don't think that you are using `js` inside the new package.

Let's drop

```
allowJs
checkJs
```

packages/cubejs-backend-cloud/tsconfig.json
```diff
@@ -0,0 +1,19 @@
...
+    "checkJs": true,
+    "noImplicitAny": false,
```
ovr, line 13: I think it's better to enable it. Don't forget that it's possible to do `Record<string, any>`.

PR #2462, opened 2021-03-30
fix: sqlAlias on not partioned rollups

ovr:
Hello!

Thanks

### DIFF
--- a/packages/cubejs-schema-compiler/src/adapter/PreAggregations.js
+++ b/packages/cubejs-schema-compiler/src/adapter/PreAggregations.js
@@ -13,11 +13,16 @@ export class PreAggregations {
    * @return {unknown[]}
    */
   preAggregationsDescription() {
-    return R.pipe(R.unnest, R.uniqBy(desc => desc.tableName))(
-      [this.preAggregationsDescriptionLocal()].concat(
-        this.query.subQueryDimensions.map(d => this.query.subQueryDescription(d).subQuery)
-          .map(q => q.preAggregations.preAggregationsDescription())
-      )
+    const preAggregations = [this.preAggregationsDescriptionLocal()].concat(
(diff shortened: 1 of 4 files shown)


packages/cubejs-schema-compiler/src/adapter/PreAggregations.js
```diff
@@ -107,8 +112,8 @@ export class PreAggregations {
 
   preAggregationDescriptionFor(cube, foundPreAggregation) {
     const { preAggregationName, preAggregation } = foundPreAggregation;
-    const name = foundPreAggregation.sqlAlias || preAggregationName;
```
ovr, line 110: `foundPreAggregation ` doesn't contain `sqlAlias`. `foundPreAggregation. preAggregation` should be used.

packages/cubejs-schema-compiler/src/adapter/PreAggregations.js
```diff
@@ -593,18 +599,21 @@ export class PreAggregations {
     return this.query.cubeEvaluator.evaluatePreAggregationReferences(cube, aggregation);
   }
 
-  originalSqlPreAggregationTable(preAggregation) {
-    if (this.canPartitionsBeUsed(preAggregation)) {
-      return this.partitionUnion(preAggregation);
+  originalSqlPreAggregationTable(preAggregationDescription) {
```
hassankhan, line 602: Might be silly, but should the variable `preAggregationDescription` be called `preAggregationDescriptor` instead?

ovr: @hassankhan First, I renamed it to descriptor, but I realize that we have a method that called description.

hassankhan: Makes sense, thanks for clarifying 👍

PR #2488, opened 2021-04-04
misc(birdbox): Run birdbox on CLI

ovr:
Hello!

Thanks

### DIFF
--- a/packages/cubejs-testing/src/birdbox.ts
+++ b/packages/cubejs-testing/src/birdbox.ts
@@ -1,18 +1,30 @@
 import { DockerComposeEnvironment } from 'testcontainers';
 import path from 'path';
+import { spawn } from 'child_process';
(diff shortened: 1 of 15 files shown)


packages/cubejs-testing/src/birdbox.ts
```diff
@@ -1,18 +1,30 @@
...
-export async function startBidBoxContainer(options: BirdBoxTestCaseOptions) {
+export interface BirdBox {
+  stop: () => Promise<void>,
+  configuration: {
+    playgroundUrl: string,
+    apiUrl: string,
+  }
+}
+
+export async function startBidBoxFromContainer(options: BirdBoxTestCaseOptions): Promise<BirdBox> {
```
vasilev-alex, line 19: `bird`?

ovr: Good catch!

PR #2492, opened 2021-04-05
feat: Make apiSecret optional, when JWK_URL is specified

ovr:
Hello!

Thanks


vasilev-alex (approved):

packages/cubejs-server-core/src/core/server.ts
```diff
@@ -404,6 +392,30 @@ export class CubejsServerCore {
...
+        'dbType'
+      ];
+
+      if (!options.jwt?.jwkUrl) {
+        // apiSecret is required only for auth by JWT, for JWK it's not needed
+        fieldsForValidation.push('apiSecret');
+      }
+
+      if (!fieldsForValidation.every((field) => options[field] !== undefined)) {
+        throw new Error(
```
vasilev-alex, line 413: can we show which option is missing exactly?

vasilev-alex (approved):

PR #2516, opened 2021-04-08
fix: Typo JWK_KEY -> JWT_KEY

ovr:
Hello!

Thanks


docs/content/Configuration/Config.md
```diff
@@ -478,8 +478,8 @@ using `CUBEJS_JWK_URL`.
 
 #### key
 
-A JSON string that represents a cryptographic key and its' properties. Can also
-be set using `CUBEJS_JWK_KEY`.
+A JSON string that represents a cryptographic key. Similar to `API_SECRET`, which will be deprecated in the feature. Can also
```
rpaik, line 481:
```suggestion
A JSON string that represents a cryptographic key. Similar to `API_SECRET`, which will be deprecated in the future. Can also
```

rpaik: I assume this is what you meant? If you meant `feature`, I think you can make this simpler by just saying `which will be deprecated`.

docs/content/Configuration/Environment-Variables-Reference.md
```diff
@@ -35,13 +35,13 @@ options][link-config].
...
-| `CUBEJS_JWT_ISSUER`                  | An issuer value which will be used to enforce the [`iss` claim from inbound JWTs][link-jwt-ref-iss]                                                                           | `https://<AUTH0-SUBDOMAIN>.auth0.com/`                                                                |
-| `CUBEJS_JWT_SUBJECT`                 | A subject value which will be used to enforce the [`sub` claim from inbound JWTs][link-jwt-ref-sub]                                                                           | `person@example.com`                                                                                  |
+| `CUBEJS_JWT_KEY`                     | The secret key used to sign and verify JWTs. Similar to `CUBEJS_API_SECRET`, which will be deprecated in the feature                                                                             | A JSON object conforming to [JWK specification RFC 7517 Section 4][link-jwk-ref], encoded as a string |
```
rpaik, line 39: same comment as above regarding `feature`.

rpaik:
@ovr, just one minor suggestion.

docs/content/Configuration/Config.md
hassankhan: +1 to @rpaik's suggestion. Also, we should add `API_SECRET` to  `DEPRECATION.md`, but let's do that as part of a separate PR

hassankhan (approved):

keydunov:
@hassankhan @ovr I'm not sure we need to say anything about potential and future deprecation. Once (and if) we decide to deprecate it - then it would be a better time to mention it.

PR #2575, opened 2021-04-19
fix: RedisPool issue with broken release on exception

ovr:
refs https://github.com/cube-js/cube.js/issues/2574


ifokeev:
had problems with this too

PR #2585, opened 2021-04-21
feat(@cubejs-client/core): add type definition to the categories method

rsbh:
**Check List**
- [ x ] Tests has been run in packages where changes made if available
- [ x ] Linter has been run for changed code

There is a `categories` method on `resultSet` which is even shown in cube.js homepage examples.
but due to the absence of its type definitions, I am unable to use it in a TS project without `ts-ignore`

### DIFF
--- a/packages/cubejs-client-core/index.d.ts
+++ b/packages/cubejs-client-core/index.d.ts
@@ -22,11 +22,11 @@ declare module '@cubejs-client/core' {
      */
     headers?: Record<string, string>;
     credentials?: 'omit' | 'same-origin' | 'include';
-    method?: 'GET' | 'PUT' | 'POST' | 'PATCH'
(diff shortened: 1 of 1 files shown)


ovr (approved):
LGTM

chamabreu:
Hi, i had the same Problem.
I updated my package.json and now it works.

But if i try to map categories() I get a ChartPivotRow as type back. I guess thats not right.

```

const createLabels = (): string[] => {
    const dateLabels = resultSet.categories().map((wrongTypedProperty) => {
      return new Date(wrongTypedProperty.category).toLocaleDateString('de-DE', { weekday: 'short', day: '2-digit', month: '2-digit', year: '2-digit' })
    })

    return dateLabels
}

```

the 'wrongTypedProperty' shows me its a ChartPivotRow and has only x and xValue als properties, but i can access the .category property.

PR #2608, opened 2021-04-26
chore(cubestore): refactor `RemoteFs`

ilya-biryukov:
The goal is to reduce code duplication and separate local files
management from the cloud storage API wrappers.

We now have `RemoteStorage` for the latter.


ilya-biryukov:
Closing to remove some noise from my github. Feel free to re-open/copy commit if needed.

PR #2611, opened 2021-04-27
fix(cubestore-driver): Ping connection only when it's OPEN

ovr:
Hello!

Thanks


packages/cubejs-cubestore-driver/src/WebSocketConnection.ts
```diff
@@ -21,13 +21,15 @@ export class WebSocketConnection {
       webSocket.readyPromise = new Promise<WebSocket>((resolve, reject) => {
         webSocket.lastHeartBeat = new Date();
         const pingInterval = setInterval(() => {
-          if (webSocket.readyState !== WebSocket.CLOSED) {
+          if (webSocket.readyState === WebSocket.OPEN) {
```
ovr, line 24: I think we should ping only when the connection is opened.

image

PR #2752, opened 2021-05-17
fix(@cubejs-client/core): compareDateRange pivot fix

### LINKED ISSUE #2739: fillMissingDates does not work when using compareDateRange
Hello! thanks for the awesome library.

**Describe the bug**

When using `compareDateRange`, `fillMissingDates` does not work, on `chartPivot` and `tablePivot`.

The reason is, that after calling `normalizePivotConfig`, `x` property is always more, than 1, which causes `fillMissingDates` to be ignored.

For parameters: 

when calling `normalizePivotConfig` with parameters:
```
     {
        x: ['SalesDaily.reportDate'],
        fillMissingDates: true,
      }
```
it returns:
```
    {
      fillMissingDates: true;
      joinDateRange: false;
      x: (2)[('compareDateRange', 'SalesDaily.reportDate.day')];
      y: ['measures'];
    }
```

Is it desired behavior? Can we override `normalizePivotConfig`, if I provide x values manually?

**To Reproduce**
run any query using `compareDateRange`
use chartPivot

vasilev-alex:
Closes #2739

### DIFF
--- a/packages/cubejs-client-core/src/ResultSet.js
+++ b/packages/cubejs-client-core/src/ResultSet.js
@@ -255,7 +255,7 @@ class ResultSet {
     const dimensionFilter = (key) => allDimensions.includes(key) || key === 'measures';
 
     pivotConfig.x = pivotConfig.x.concat(
-      allDimensions.filter(d => !allIncludedDimensions.includes(d))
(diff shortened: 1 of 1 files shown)


ovr (approved):
LGTM

PR #2765, opened 2021-05-17
docs(cloud): add guides for AWS Athena, AWS Redshift, BigQuery and Snowflake

hassankhan:
Story details: https://app.clubhouse.io/cube-dev/story/531


igorlukanin (commented):
Great job and brilliant idea, Hassan! I also suggest that every database's connecting info should be put onto a dedicated page in the docs. Developers rarely need to connect to multiple databases so it's really no need to prevent them from navigating to a dedicated page. However, it would allow us to present the information in a focused yet detailed manner, without creating a single mega-page.

keydunov (approved):
LGTM!

PR #2771, opened 2021-05-18
docs(caching): use correct port for Cube Store in Docker Compose example

hassankhan:


docs/content/Caching/Running-in-Production.md
```diff
@@ -130,7 +130,6 @@ services:
       - CUBESTORE_WORKERS=cubestore_worker_1:9001,cubestore_worker_2:9001
       - CUBESTORE_REMOTE_DIR=/cube/data
     expose:
-      - 9999 # This exposes the Metastore endpoint
```
ilya-biryukov, line 133: This won't work.
You need to expose the metastore port in docker compose so that workers and router can communicate.

Put the default value for the port here?

ilya-biryukov (changes requested):

keydunov (approved):

PR #2787, opened 2021-05-20
chore(docs): Remove unnecessary port 3000 from Docker Compose docs

igorlukanin:
**Check List**
—

—

Fixing what looks like a mistype in the `docker-compose.yml`.

### DIFF
--- a/docs/content/Getting-Started/Getting-Started-Docker-Compose.md
+++ b/docs/content/Getting-Started/Getting-Started-Docker-Compose.md
@@ -26,11 +26,7 @@ services:
   cube:
     image: cubejs/cube:latest
     ports:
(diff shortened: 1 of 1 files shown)


igorlukanin:
@hassankhan It might be interesting to track how this 3000 port made it to this `docker-compose.yml` in the first place. Can we have anything similar in other places?

hassankhan:
@igorlukanin It came from a freshly-generated project via the Cube.js CLI tool; might be good to also update the templates in that case 😀

igorlukanin:
@hassankhan OMG, we have plenty to update: https://github.com/cube-js/cube.js/search?q=3000%3A3000

Issue #960, opened 2020-08-06
Reduce bundle size of @cubejs-client


mgebeily:
**Is your feature request related to a problem? Please describe.**
The cube.js client is an awesome way to interact with the server. However, it's [rather large](https://bundlephobia.com/result?p=@cubejs-client/core@0.19.56) for a client side library, with the bulk of the size of it being comprised of dependencies, and half of which come from a single library (moment). There are also polyfills bundled with the library that could be deferred to the user's own browser support needs. This could lead to a large increase in the size of javascript apps that utilize the cube client.

**Describe the solution you'd like**
There are some smaller alternatives to moment that provide the same functionality. [dayjs](https://github.com/iamkun/dayjs) is one of the most popular and is 5% the size. Most of the usage of rambda could be replaced with native ES6 and polyfilled by the user depending on the browsers they support. There may also be value in re-arranging some of the code to allow for tree shaking.

**Describe alternatives you've considered**
The alternative to this would be to interact directly with the server through http. The cube client provides some really useful functionality and would be great to use if it was a little more lightweight. It seems that the core of the package has almost all of the logic and makes up a small percentage of the size.

**Additional context**
Any moment replacement would have to be strenuously tested to make sure that it can parse the same date formats, as users can pass arbitrary strings to moment. Even then, all of the above suggestions would probably be considered breaking. I wanted to see if others have the same concerns and if there was interest in a pull request before I try to create one. Let me know if I completely missed the mark with the above.

I'm really looking forward to working with cube. Reporting has always been a pain point for me and this seems like an excellent solution. Thanks in advance!

vasilev-alex:
hi @mgebeily! Thanks for posting it here. Yes, it's difficult to disagree with you on the bundle size. 

We could've just replaced `moment` with `moment-mini` that comes without unnecessary locales but we also depend on `moment-range`.  I guess we can try `dayjs` if we can find an alternative to `moment-range`.

As for `ramda` we should opt for smaller bundle size as well, and find more lightweight alternatives for the methods we can't simply replace.

And, of course, PRs are very welcomed!

vasilev-alex:
Let's keep it open then and track all related discussions here

sveisvei:
I would like to help with this, how would you like this approached? Ramda I think is small enough that is does not matter, moment on the other hand is very large.

[...]

PR #2794, opened 2021-05-21
refactor(@cubejs-client/core): replace `moment` with `dayjs`

mgebeily:
**Check List**

**#960**

Removes a dependency on moment / moment-range and replaces it with dayjs. This reduces bundle size by 60%. Areas of interest would be in differences between date string parsing methodologies of moment and dayjs.


packages/cubejs-client-core/src/tests/drill-down.test.js
```diff
@@ -123,7 +123,7 @@ describe('drill down query', () => {
       timeDimensions: [
         {
           dimension: 'Orders.ts',
-          dateRange: ['2020-08-01T00:00:00.000', '2020-08-01T23:59:59.999'],
+          dateRange: ['2020-08-01T00:00:00.000', '2020-08-01T23:59:59.000'],
```
vasilev-alex, line 126: Is there any reason you truncate time here? Feels like it may lead to unpredicted behavior.

vasilev-alex (changes requested):

deckchairlabs:
Perhaps better to use a https://github.com/dmtrKovalenko/date-io and abstract away the date implementation? That way users can use any library supported by date-io?

packages/cubejs-client-core/src/tests/drill-down.test.js
mgebeily: Sorry for the careless mistake, I thought this was a semantic change in how dayjs parses but realized I did in fact use the wrong stringify method.

mgebeily:
> Perhaps better to use a https://github.com/dmtrKovalenko/date-io and abstract away the date implementation? That way users can use any library supported by date-io?

I don't think this is necessary, as dates are passed around as strings and each of the libraries supported by date-io have enough string parsing capabilities to handle the incoming formats. The only benefit I see to it would be for lack of redundancy in bundle size but dayjs is small enough to be considered negligible here.

packages/cubejs-client-core/src/ResultSet.js
```diff
@@ -45,6 +35,25 @@ const groupByToPairs = (keyFn) => {
...
 };
 
+const dayRange = (from, to) => ({
+  by: (value) => {
+    const results = [];
+
+    let start = dayjs(from);
+    const end = dayjs(to);
+
+    while (start.isBefore(end)) {
```
vasilev-alex, line 45: I believe it should be inclusive. We should add the `end` date to the results

vasilev-alex (changes requested):

packages/cubejs-client-core/src/ResultSet.js
vasilev-alex: @mgebeily added a small test for that, so please pull from the upstream

rpaik:
@mgebeily, just checking back. Will you be able to incorporate @vasilev-alex's suggestions 👆 ?

mgebeily:
> @mgebeily, just checking back. Will you be able to incorporate @vasilev-alex's suggestions point_up_2 ?

Very sorry, let this get away from me a bit. I just pushed up the fix.

rpaik:
@vasilev-alex, 🏓 back to you 😄

vasilev-alex (approved):

PR #2837, opened 2021-05-27
chore: track setter

vasilev-alex:


packages/cubejs-playground/src/events.ts
```diff
@@ -8,18 +8,23 @@ let baseProps = {
   sentFrom: 'frontend',
 };
 let telemetry: boolean | undefined;
+let track: null | ((event: Record<string, any>, telemetry?: boolean) => Promise<any>) = null;
```
ovr, line 11: Maybe it's better to assign a default implementation to protect a case when someone is using the playground as the library and it will be a BC for him.

ovr (approved):
LGTM

PR #2838, opened 2021-05-27
feat(redshift-driver): Support UNLOAD (direct export to S3)

ovr:
Hello!

Thanks


packages/cubejs-redshift-driver/src/RedshiftDriver.ts
```diff
@@ -0,0 +1,165 @@
...
+      if (emptyKeys.length) {
+        throw new Error(
+          `Unsupported configuration exportBucket, some configuration keys are empty: ${emptyKeys.join(',')}`
+        );
+      }
+
+      return <RedshiftDriverExportAWS>exportBucket;
+    }
+
+    return undefined;
```
vasilev-alex, line 62: `return` returns `undefined` by default

vasilev-alex (approved):

packages/cubejs-redshift-driver/src/RedshiftDriver.ts
ovr: There is a problem with that because our eslint configuration doesn't allow that.

https://eslint.org/docs/rules/consistent-return#treatundefinedasunspecified

Issue #2831, opened 2021-05-27
Incorrect column type used for time dimensions when using MSSQL as an external preaggregation store.


davefrancis:
**Describe the bug**
When using a `timeDimensionReference` in the pre-aggregation definition I am encountering this error: `RequestError: Cannot insert an explicit value into a timestamp column. Use INSERT with a column list to exclude the timestamp column, or insert a DEFAULT into the timestamp column.` 

While debugging I've noticed that the table created in the external database uses a `timestamp` data type for time dimension columns. In the case of SQL Server, this data type is actually used to version table rows and not (as the name would suggest) used to store a timestamp. 
https://docs.microsoft.com/en-us/sql/t-sql/data-types/rowversion-transact-sql?view=sql-server-ver15

**Expected behavior**
When creating external preaggregation tables in SQL Server, use datetime or datetime2 column types for time dimensions.

PR #2846, opened 2021-05-28
fix(@cubejs-backend/mssql-driver): add column type mappings for MSSQL external pre-aggregations

davefrancis:
Hi, this PR provides type mappings from generic types to MSSQL data types. Fixes bugs when creating external pre-aggregation tables in a MSSQL database.

**Check List**

#2831

### DIFF
--- a/packages/cubejs-mssql-driver/driver/MSSqlDriver.js
+++ b/packages/cubejs-mssql-driver/driver/MSSqlDriver.js
@@ -4,8 +4,7 @@ const { BaseDriver } = require('@cubejs-backend/query-orchestrator');
 const GenericTypeToMSSql = {
   string: 'nvarchar(max)',
   text: 'nvarchar(max)',
-  // should this be timezone aware? use datetimeoffset
-  timestamp: 'datetime',
+  timestamp: 'datetime2',
 };
 
 class MSSqlDriver extends BaseDriver {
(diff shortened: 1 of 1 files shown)


davefrancis:
Hi @rpaik, here is the PR regarding the MSSQL bug. I was unsure whether the `timestamp` generic type is time zone aware. If it is the PR should changed before it is merged. I also found the string and text type mappings to be incorrect as well and they are also fixed in this PR. Thanks!

rpaik:
Thanks @davefrancis 😄 

@ovr can you help with the review?

packages/cubejs-mssql-driver/driver/MSSqlDriver.js
```diff
@@ -1,6 +1,13 @@
 const sql = require('mssql');
 const { BaseDriver } = require('@cubejs-backend/query-orchestrator');
 
+const GenericTypeToMSSql = {
+  string: 'nvarchar(max)',
+  text: 'nvarchar(max)',
+  // should this be timezone aware? use datetimeoffset
+  timestamp: 'datetime',
```
ovr, line 8: I am thinking about `datetime2`, but it's supported in MSSQL >= 2008

davefrancis: @ovr I think you're right that `datetime2` would be a better option. It seems to be the suggested type from Microsoft going forward. This is from their documentation: 

> Use the time, date, datetime2 and datetimeoffset data types for new work. These types align with the SQL Standard. They are more portable. time, datetime2 and datetimeoffset provide more seconds precision. datetimeoffset provides time zone support for globally deployed applications.

I will update the source branch to use `datetime2`. Thanks for the review.

PR #2849, opened 2021-05-28
fix(@cubejs-client/core): decompose type

### LINKED ISSUE #2814: decompose() with compareDateRange in typescript
**Problem**

How can i decompose my values from a query with a 'compareDateRange' timeDimension?

My query:
```
const { resultSet, error } = useCubeQuery({
    "measures": [
      myMeasure
    ],
    "timeDimensions": [
      {
        "dimension": "My.Dimension",
        "compareDateRange": ['Today', '6 days ago']
      }
    ],
  });
```
When i try to decompose() the resultSet in typescript, it returns a `Object | null`. Thats what the documentation says.
But isn't it actually an array?
https://cube.dev/docs/@cubejs-client-core#result-set-decompose
Here it's used with .forEach() but i can't do this in Typescript because of the `Object | null` typing.

Is this a bug/error or am i doing something wrong?

Maybe there is a better way to extract my 2 resultSet Data?

vasilev-alex:
Closes #2814

### DIFF
--- a/packages/cubejs-client-core/index.d.ts
+++ b/packages/cubejs-client-core/index.d.ts
@@ -317,7 +317,7 @@ declare module '@cubejs-client/core' {
      * });
      * ```
      */
-    decompose(): Object;
+    decompose(): ResultSet[];
 
     /**
      * @hidden


ovr (approved):
ЛГТМ

PR #2854, opened 2021-05-28
Support multi-value filtering on same column through FILTER_PARAMS

omab:
**Check List**

No related issue.

This PR adds support to properly process multiple filters over the same dimension when accessing them through `FILTER_PARAMS` context variable, this becomes particularly handy when the expression involved is an `ILIKE` operator (or similar).

With this change it's possible to define SQLs like:

```js
cube('Foo', {
  sql: `SELECT id, name FROM foo WHERE ${FILTER_PARAMS.Foo.name.filter('name')}`,
  dimensions: {
    id: {
      type: 'number',
      sql: 'id',
      primaryKey: true
    },
    name: {
      type: 'string',
      sql: 'name'
    }
  }
}
```

and if the filter passed are in the form:

```js
filters: [
  {
    dimension: 'Foo.name',
    operator: 'contains',
    values: ['Foo']
  }, {
    dimension: 'Foo.name',
    operator: 'contains',
    values: ['Bar']
  }
]
```

Will result in an SQL query like:

```sql
SELECT id, name FROM foo WHERE name ILIKE '%Foo%' AND name ILIKE '%Bar%';
```

### DIFF
--- a/packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
+++ b/packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
@@ -2223,26 +2223,32 @@ export class BaseQuery {
         const cubeName = this.cubeEvaluator.cubeNameFromPath(name);
         return new Proxy({ cube: cubeName }, {
           get: (cubeNameObj, propertyName) => {
-            const filter =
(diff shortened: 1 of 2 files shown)


paveltiunov:
@omab Hey Matias! Sorry for the delay! Thanks for contributing it! Looks great!

omab:
Thanks!

PR #2876, opened 2021-06-01
feat(@cubejs-client/ngx): async CubejsClient initialization

### LINKED ISSUE #1992: Angular: How to set token after user is logged into the app?
**Problem**

A very common use case is to only request the CubeJS token once the user is authenticated to my app. Then I can provide a token initialize CubeJS.

The way it is on the docs (https://cube.dev/docs/@cubejs-client-ngx#api), you set the config on the app.module, hence, before the user has a chance to login to the app.

How would I set the token AFTER I know who the user is (after he has authenticated to my app)?

Thanks

vasilev-alex:
Closes #1992


kodeine:
@vasilev-alex so i am having difficulty using the async. if there's a delay from the API like 300ms in getting the response for token, cubejs will fail to load().


```
  const config$ = from([]).pipe(
    mergeMap(() =>
      of({
        token:
'sometoken',
        options: {
          apiUrl: environment.cubeJs.API_URL,
        },
      })
    ),
    delay(350)
  );

  return new CubejsClient(config$);
```

image

image

PR #2894, opened 2021-06-03
feat(cubestore): Use NPM's proxy settings in post-installer

ovr:
Hello!

Before this PR Cube Store installer didn't use proxy settings from NPM, which will be a problem for users who are behind the proxy.

Thanks


vasilev-alex (approved):

vasilev-alex:
выглядит хорошо для меня

PR #2903, opened 2021-06-04
Converted likeIgnoreCase to contains search in Elasticsearch

vignesh-123:
Converted likeIgnoreCase to contains search instead of fuzzy match

likeIgnoreCase function of ElasticSearchQuery was doing a fuzzy match. This leads to unexpected results in contains queries
I have changed to LIKE query as in the other Databases we support


igorlukanin:
This is most likely stale. Please feel free to reopen if needed.

PR #2927, opened 2021-06-09
docs: Remove mentions of non-existent autorollup feature

igorlukanin:
We're getting questions about `autorollup` pre-aggregations. However, this feature is not supported. Let's remove mentions of this feature from docs so we don't confuse Cube.js users.


keydunov:
👍

PR #2939, opened 2021-06-10
fix: extDbType warning

vasilev-alex:


packages/cubejs-server-core/src/core/server.ts
```diff
@@ -710,6 +710,7 @@ export class CubejsServerCore {
       }),
       redisPrefix: orchestratorId,
       orchestratorOptions: {
+        // todo: dynamic externalDbType?
```
ovr, line 713: It's already dynamic.

```
    const externalDbType = this.contextToExternalDbType(context);
```

ovr (approved):
LGTM

PR #2941, opened 2021-06-10
Picking up allowUngroupedWithoutPrimaryKey from env

vignesh-123:
allowUngroupedWithoutPrimaryKey is an option that is not available to set over the environment variables
I am trying to run the service as a docker and this will help with that
If this name is fine, I can also update the environment variables doc


paveltiunov:
Hey @vignesh-123 ! Thanks for contributing it! Could you please do it through `getEnv()` util as all other variables? Thanks!

vignesh-123:
@paveltiunov I have made the change

paveltiunov:
Thanks for contributing it, @vignesh-123 !

PR #2960, opened 2021-06-15
feat: Make MySQL/PostgreSQL readOnly by default

ovr:
Hello!

Thanks


packages/cubejs-redshift-driver/src/RedshiftDriver.ts
```diff
@@ -24,6 +24,7 @@ export class RedshiftDriver extends PostgresDriver<RedshiftDriverConfiguration>
 
   protected getInitialConfiguration(): Partial<RedshiftDriverConfiguration> {
     return {
+      readOnly: true,
```
ovr, line 27: todo: Validate that unload supported without creating table

ovr: works.!

PR #2991, opened 2021-06-22
fix(@cubejs-client/playground): invalid token

vasilev-alex:


packages/cubejs-server-core/src/core/DevServer.ts
```diff
@@ -490,4 +492,8 @@ export class DevServer {
       res.json({ token });
     }));
   }
+
+  protected getIdentifier(apiSecret: string): string {
+    return crypto.createHash('md5').update(apiSecret).digest('hex').replace(/[^\d]/g, '');
```
ovr, line 497: Maybe `.substr(0, 10)`?

ovr (approved):
LGTM

PR #2992, opened 2021-06-22
test: standard cron syntax

ovr:
Hello!

Thanks

### DIFF
--- a/docs/content/Schema/cube.md
+++ b/docs/content/Schema/cube.md
@@ -300,20 +300,56 @@ cube(`OrderFacts`, {
 });
 ```
 
-`every` can accept only equal time intervals - so  "Day of month" and "month" intervals in CRON expressions are not supported.
+`every` can accept only equal time intervals - so "Day of month" and "month" intervals in CRON expressions are not supported.
+
+<!-- prettier-ignore-start -->
+[[warning |]]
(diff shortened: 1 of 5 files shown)


docs/content/Schema/cube.md
```diff
@@ -300,18 +300,25 @@ cube(`OrderFacts`, {
 });
 ```
 
-`every` can accept only equal time intervals - so  "Day of month" and "month" intervals in CRON expressions are not supported.
+`every` can accept only equal time intervals - so "Day of month" and "month" intervals in CRON expressions are not supported.
+
+[[warning |]]
+| Cube.js supports two different formats of CRON expressions: with leading seconds (6 parts) and leading minutes (5 parts).
```
hassankhan, line 307: @ovr any chance we could add an example for both formats, please? 😀

hassankhan (approved):

docs/content/Schema/cube.md
ovr: @hassankhan shall I add it to this warning or below?

hassankhan: I think to the warning would be best; something like:

```
Cube.js supports two different formats of CRON expressions; with leading seconds (EXAMPLE_HERE) and with leading minutes (EXAMPLE_HERE).
```

hassankhan (approved):
LGTM 🙌

PR #2993, opened 2021-06-22
feat: Execute refreshKeys in externalDb (only for every)

ovr:
Hello!

## Internal

First, there is an internal change, that we started to pass query options as 3rd parameter in Query tuple. 

## Motivation

We are going to execute refresh keys (only for every, just now) in an external database. In the feature, we are planning to improve support for every via changes in the cube store (support more useful SQL things).

Thanks


packages/cubejs-query-orchestrator/src/orchestrator/PreAggregations.ts
```diff
@@ -267,28 +267,29 @@ class PreAggregationLoadCache {
...
-    if (!this.queryResults[this.queryCache.queryRedisKey(keyQuery)]) {
-      this.queryResults[this.queryCache.queryRedisKey(keyQuery)] = await this.queryCache.cacheQueryResult(
-        Array.isArray(keyQuery) ? keyQuery[0] : keyQuery,
-        Array.isArray(keyQuery) ? keyQuery[1] : [],
-        keyQuery,
+  protected async keyQueryResult(sqlQuery: QueryWithParams, waitForRenew, priority) {
+    if (!this.queryResults[this.queryCache.queryRedisKey(sqlQuery)]) {
+      const [query, values, queryOptions]: QueryTuple = Array.isArray(sqlQuery) ? sqlQuery : [sqlQuery, [], {}];
+
+      this.queryResults[this.queryCache.queryRedisKey(sqlQuery)] = await this.queryCache.cacheQueryResult(
```
ovr, line 274: // [query, values]

Issue #3048, opened 2021-07-01
non-root user for cubestore


loremaps:
**Is your feature request related to a problem? Please describe.**

Follow [non-root user best practice](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user) for the docker image of cubestore.

**Describe the solution you'd like**

Be able to deploy cubestore in Kubernetes using non-root user. e.g.
```yaml
   ...
    spec:
      containers:
        - name: cubestore
          image: cubejs/cubestore:latest
         ...
      securityContext:
        runAsUser: 1000
```

**Describe alternatives you've considered**

N/A

**Additional context**

Right now if I run cubestore as non-root I get the following error:

```
2021-07-01 14:06:47,261 INFO  [cubestore::metastore] 1> Creating metastore from scratch in /cube/.cubestore/data/metastore
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error { message: "Failed to create RocksDB directory: `Os { code: 13, kind: PermissionDenied, message: \"Permission denied\" }`." }', /build/cubestore/cubestore/src/metastore/mod.rs:1542:40
stack backtrace:
   0: rust_begin_unwind
             at /rustc/a143517d44cac50b20cbd3a0b579addab40dd399/library/std/src/panicking.rs:493:5
   1: core::panicking::panic_fmt
             at /rustc/a143517d44cac50b20cbd3a0b579addab40dd399/library/core/src/panicking.rs:92:14
   2: core::option::expect_none_failed
             at /rustc/a143517d44cac50b20cbd3a0b579addab40dd399/library/core/src/option.rs:1300:5
   3: cubestore::metastore::RocksMetaStore::new
   4: cubestore::metastore::RocksMetaStore::load_from_remote::{{closure}}
   5:  as core::future::future::Future>::poll
   6:  as core::future::future::Future>::poll
   7:  as core::future::future::Future>::poll
   8:  as core::future::future::Future>::poll
   9:  as core::future::future::Future>::poll
  10:  as core::future::future::Future>::poll
  11: cubestored::main::{{closure}}
  12: tokio::runtime::enter::Enter::block_on
  13: tokio::runtime::thread_pool::ThreadPool::block_on
  14: tokio::runtime::Runtime::block_on
  15: cubestored::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

loremaps:
[...]

PR #3049, opened 2021-07-01
fix(cubestore): non-root user

loremaps:
**Check List**

#3048

I added a non-root user.

### DIFF
--- a/rust/Dockerfile
+++ b/rust/Dockerfile
@@ -42,4 +42,8 @@ EXPOSE 3306
 
 ENV RUST_BACKTRACE=true
 
+RUN chown nobody /cube
+
+USER nobody
+
 CMD ["./cubestored"]


ilya-biryukov:
Could you add a bit more context.
Why would we prefer to run as a non-root user inside a container?

ilya-biryukov:
Thanks, that's a good read. I did know Docker was running its daemon as root, but wasn't sure about the alternatives.
I do agree that running as root is the wrong default. But is something that should be handled on the `docker`/`k8s`/whoever runs the containers, not on the containers themselves. One really needs to enable user namespaces/create users and sandbox each container separately.

I would be up for changing the default, but  I don't think user `1000` is a good one. On my desktop machine that's the default user, so it may feel like that's what the users want most of the time. But that is **accidental**. Using `1000` on a shared system is no good as we can hit an account that has reasonable permissions. Even on my system, I would rather take an account that has no permissions than the one that can read my home folder.

Are there better alternatives?
Thinking about *nobody* (id `65534`). That gives very little access to the system, hard to do anything interesting with it.
There's also a high chance that this user is present on the host system.

ilya-biryukov:
How are you mounting the volume? Here's the variation that works for me in the container:
```
RUN chown nobody /cube
USER nobody
```
If I mount a folder on the host into `/cube`, e.g. by running `docker run -v /home/myuser/db:/cube`, I also have to ensure container user has the rights to write into that folder, e.g. with `docker run … -u myuser`. The latter effectively overrides the `USER` directive.
Same goes for K8s, one needs to say:
```
  securityContext:
    runAsUser: …
    fsGroup: …
```
No matter which option we choose, we can't guarantee permissions on the mounted folders match inside container. It's a burden on people configuring containers/pods, we don't have control over that as image owners.
I have stressed why I think `1000` above is bad idea, same applies to `999` and `1001`.

loremaps:
Yes, I messed up sorry 😬  I have updated the PR.

ilya-biryukov:
I have realized this change will break existing CubeStore setups.

Here's what will happen:
- we publish the new images,
- setups who don't set the user explicitly (the default) will pull the new image,
- cubestore will try to run as "nobody" and won't be able to write into volumes.

This happens because prior to the patch, the default setup wrote files and directories as root, so other users can't overwrite them.
I'm not sure how many instances we have running in production, but I'm reluctant to break them.

loremaps:
Is it on K8s? with persisted volumes? On top of my head, a possible migration plan would be to use `initContainers` for your deployments. For example:

```yaml
     ....
      initContainers:
        - name: volume-permissions
          image: busybox
          command:
            - sh
            - -c
            - chown -R nobody:nobody /cube
          securityContext:
            runAsUser: 0
          volumeMounts:
            - mountPath: /cube
              name: cubestore-volume
```

But ofc I do not know the details, so sorry if this suggestion is irrelevant. 
However, since those instances are in production, it might be a good idea to consider this security improvement. Thanks

ilya-biryukov:
`CUBE_DATA_DIR` must point to the directory that the user you run with has access to.
It's `/cube/.cubestore` by default inside the container, so this works fine:
```
docker run --user=1000 -v $PWD:/cube/.cubestore cubejs/cubestore
```

loremaps:
Thanks @ilya-biryukov, this is useful! If this PR is not working for you plz feel free to close it. (I am leaving it open because I think it leads to a be experience). Thanks again for your time!

ilya-biryukov:
Closing the PR.
Thanks for bringing this up, though, we should definitely mention that running with non-root user should probably be preferred in production.
@hassankhan let's update the [relevant docs](https://cube.dev/docs/caching/running-in-production)? Feel free to ping me about this.

PR #3067, opened 2021-07-05
fix(docs): top banner on mobile

YakovlevCoded:
**Check List**

[For example #12]

[Description goes here]


hassankhan (changes requested):
Doesn't seem to be loading up on my iPhone with Safari

YakovlevCoded:
> Doesn't seem to be loading up on my iPhone with Safari

it seems like api is been disabled with banner data :(

PR #3072, opened 2021-07-06
fix(@cubejs-client/core): Long Query 413 URL too large

telunc:
**Check List**

#1558 

**Description**
In HttpTransport, `this.method = this.method || (url.length < 2000 ? 'GET' : 'POST');` the later url length logic will only get run once if method is undefined. Any request afterwards will be taken from `this.method`. For apps that uses the client as a singleton like angular, client will never fire a `POST` request if query is too long.


vasilev-alex (changes requested):
I assume we no longer need the `this.method` field

telunc:
Yes, exactly.

vasilev-alex (approved):

PR #3092, opened 2021-07-12
docs(schema): create new Concepts page under new Fundamentals section

hassankhan:
Story details: https://app.clubhouse.io/cube-dev/story/784

### DIFF
--- a/docs/content/Schema/Fundamentals/Concepts.md
+++ b/docs/content/Schema/Fundamentals/Concepts.md
@@ -1,0 +1,306 @@
+---
+title: Data Schema Concepts
+permalink: /schema/fundamentals/concepts
+category: Data Schema
+menuOrder: 1
+subCategory: Fundamentals
+---
+
+Cube.js borrows a lot of terminology from data science and [OLAP
+theory][wiki-olap], and this document is intended for both newcomers and regular
+users to refresh their understanding.
+
+We'll use a sample e-commerce database with two tables, `orders` and
+`line_items` to illustrate the concepts throughout this page:
+
+**`orders`**
+
+| **id** | **status** | **completed_at**           | **created_at**             |
+| ------ | ---------- | -------------------------- | -------------------------- |
+| 1      | completed  | 2019-01-05 00:00:00.000000 | 2019-01-02 00:00:00.000000 |
+| 2      | shipped    | 2019-01-17 00:00:00.000000 | 2019-01-02 00:00:00.000000 |
+| 3      | completed  | 2019-01-27 00:00:00.000000 | 2019-01-02 00:00:00.000000 |
+| 4      | shipped    | 2019-01-09 00:00:00.000000 | 2019-01-02 00:00:00.000000 |
+| 5      | processing | 2019-01-29 00:00:00.000000 | 2019-01-02 00:00:00.000000 |
+
+**`line_items`**
+
+| **id** | **product_id** | **order_id** | **quantity** | **price** | **created_at**             |
+| ------ | -------------- | ------------ | ------------ | --------- | -------------------------- |
+| 1      | 31             | 1            | 1            | 275       | 2019-01-31 00:00:00.000000 |
+| 2      | 49             | 2            | 6            | 248       | 2021-01-20 00:00:00.000000 |
+| 3      | 89             | 3            | 6            | 197       | 2021-11-25 00:00:00.000000 |
+| 4      | 71             | 4            | 8            | 223       | 2019-12-23 00:00:00.000000 |
+| 5      | 64             | 5            | 5            | 75        | 2019-04-20 00:00:00.000000 |
+
+## Cubes
+
+A cube represents a dataset in Cube.js, and is conceptually similar to a [view
+in SQL][wiki-view-sql]. Cubes are typically declared in separate files with one
+cube per file. Within each cube are definitions of [dimensions][self-dimensions]
+and [measures][self-measures]. Typically, a cube points to a single table in
+your database using the [`sql` property][ref-schema-ref-sql]:
+
+```javascript
+cube('Orders', {
+  sql: `SELECT * FROM orders`,
+});
+```
+
+The `sql` property of a cube is flexible enough to accommodate more complex SQL
+queries too:
+
+```javascript
+cube('Orders', {
+  sql: `
+SELECT
+  *
+FROM
+  orders,
+  line_items
+WHERE
+  orders.id = line_items.order_id
+  `,
+});
+```
+
+## Dimensions
+
+Dimensions represent the properties of a **single** data point in the cube.
+[The `orders` table](#top) contains only dimensions, so representing them in the
+`Orders` cube is straightforward:
+
+```javascript
+cube('Orders', {
+
+  ...,
+
+  dimensions: {
+    id: {
+      sql: `id`,
+      type: `number`,
+      // Here we explicitly let Cube.js know this field is the primary key
+      // This is required for de-duplicating results when using joins
+      primaryKey: true
+    },
+    status: {
+      sql: `status`,
+      type: `string`
+    },
+  },
+});
+```
+
+[The `line_items` table](#top) also has a couple of dimensions which can be
+represented as follows:
+
+```javascript
+cube('LineItems', {
+
+  ...,
+
+  dimensions: {
+    id: {
+      sql: `id`,
+      type: `number`,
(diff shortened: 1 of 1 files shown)


docs/content/Schema/Fundamentals/Concepts.md
```diff
@@ -0,0 +1,328 @@
...
+## Cubes
+
+A cube represents a dataset in Cube.js, and is conceptually similar to a [view
+in SQL][wiki-view-sql]. Cubes are typically declared in separate files with one
+cube per file. Within each cube are definitions of measures and dimensions.
+Typically, a cube points to a single table in your database using the [`sql`
+property][ref-schema-ref-sql]:
+
+<!-- prettier-ignore-start -->
+[[info | ]]
```
rpaik, line 44: I assume this will be deleted as there's a duplicate below in the Dimensions section?

docs/content/Schema/Fundamentals/Concepts.md
```diff
@@ -0,0 +1,328 @@
...
+| 2      | 49             | 2            | 6            | 248       | 2021-01-20 00:00:00.000000 |
+| 3      | 89             | 3            | 6            | 197       | 2021-11-25 00:00:00.000000 |
+| 4      | 71             | 4            | 8            | 223       | 2019-12-23 00:00:00.000000 |
+| 5      | 64             | 5            | 5            | 75        | 2019-04-20 00:00:00.000000 |
+
+## Cubes
+
+A cube represents a dataset in Cube.js, and is conceptually similar to a [view
+in SQL][wiki-view-sql]. Cubes are typically declared in separate files with one
+cube per file. Within each cube are definitions of measures and dimensions.
```
rpaik, line 39: Would it make sense to make `measures` and `dimensions` links to the respective sections below?

rpaik:
Thanks @hassankhan, looks great 😊 Just added a few comments...

docs/content/Schema/Fundamentals/Concepts.md
```diff
@@ -0,0 +1,323 @@
...
+```
+
+Dimensions can be of different types, and you can find them all
+[here][ref-schema-dimension-types].
+
+### Time Dimensions
+
+<!-- prettier-ignore-start -->
+[[info | ]]
+| Time-based properties are generally used as dimensions in Cube.js. // say why
```
hassankhan, line 125: Is there a specific reason temporal attributes in a dataset are dimensions? I would imagine its because timestamps aren't quite measures (although that's not _completely_ true either). Would <3 some clarification

/cc @keydunov @paveltiunov

docs/content/Schema/Fundamentals/Concepts.md
```diff
@@ -0,0 +1,323 @@
...
+  measures: {
+    total: {
+      sql: `price`,
+      type: `sum`,
+    },
+  },
+})
+```
+
+> Add a mention of how to create measures from dimensions (SUM something?)
```
hassankhan, line 216: @keydunov Is it worth talking about converting dimensions to measures here or should we do that in advanced concepts/recipes instead?

docs/content/Schema/Fundamentals/Concepts.md
```diff
@@ -0,0 +1,303 @@
...
+});
+```
+
+Dimensions can be of different types, and you can find them all
+[here][ref-schema-dimension-types].
+
+### Time Dimensions
+
+Time-based properties should be represented as dimensions with type `time`. Time
+dimensions allow granulating the result set for queries sent to Cube.js.
```
igorlukanin, line 120: While it's correct to say "time dimensions allow granulating the result set", I'm not completely sure that it's enough to say this so everyone understands. IDK what might work better though. Maybe "t. d. allow grouping by time units such as hours, days, or months"?

docs/content/Schema/Fundamentals/Concepts.md
```diff
@@ -0,0 +1,303 @@
...
+
+  measures: {
+    count: {
+      type: `count`,
+    },
+  },
+});
+```
+
+In our `LineItems` cube, we can We could also create a measure to sum up the
```
igorlukanin, line 186: "we can We could" — looks suspicious :)

docs/content/Schema/Fundamentals/Concepts.md
```diff
@@ -0,0 +1,303 @@
...
+```
+
+Measures can be of different types, and you can find them all
+[here][ref-schema-measure-types].
+
+## Joins
+
+Joins define the relationships between cubes, which then allows accessing and
+comparing properties from two or more cubes at the same time. In Cube.js, all
+joins are `LEFT JOIN`s by default.
```
igorlukanin, line 210: I wonder if we should keep "by default" here. There's no way to change these defaults, right?

docs/content/Schema/Fundamentals/Concepts.md
```diff
@@ -0,0 +1,303 @@
...
+
+<!-- prettier-ignore-start -->
+[[info | ]]
+| An `INNER JOIN` can be replicated with Cube.js; when making a Cube.js query,
```
igorlukanin, line 214: I also wonder if we should explain the difference between `LEFT JOIN`s and `INNER JOIN`s here — or at least provide a relevant link.

docs/content/Schema/Fundamentals/Concepts.md
```diff
@@ -0,0 +1,303 @@
...
+
+```javascript
+cube('Orders', {
+
+  ...,
+
+  joins: {
+    LineItems: {
+      relationship: `belongsTo`,
+      sql: `${CUBE}.id = ${LineItems}.order_id`,
```
igorlukanin, line 229: Suddenly, we're using `${CUBE}` to reference the cube. I wonder if there should be a place in the docs (maybe not on this page) where we recommend always using `${CUBE}` or a specific cube reference to prevent name clashes.

docs/content/Schema/Fundamentals/Concepts.md
```diff
@@ -0,0 +1,303 @@
...
+      sql: `${CUBE}.status = 'completed'`
+    },
+  },
+});
+```
+
+## Pre-Aggregations
+
+Pre-aggregations are a powerful way of caching frequently-used, expensive
+queries and keeping the cache up-to-date on a periodic basis. Within a data
```
igorlukanin, line 267: I assumed that we want users to use pre-aggregations for **all** queries rather than just **frequently-used, expensive** ones because otherwise we cannot provide any guarantees w.r.t. performance. I think it would be great to come to a shared understanding here. Cc @hassankhan @keydunov @ryanpei

docs/content/Schema/Fundamentals/Concepts.md
igorlukanin: Also, it's great to explain `${CUBE}.column` vs `${CUBE.member}`.

igorlukanin (approved):
As always, marvelous job, @hassankhan! 😻

 I've also left a few comments that you might find useful.

docs/content/Schema/Fundamentals/Concepts.md
hassankhan: Definitely can add a link, do you have one in mind? 😀

docs/content/Schema/Fundamentals/Concepts.md
hassankhan: Agreed that we should explain the above; but also not sure if this is the best place to do it. Also, do you think I should update _all_ the examples on this page to use the `CUBE` global?

adnanrahic (commented):
I left a few comments. :smile:

docs/content/Schema/Fundamentals/Concepts.md
```diff
@@ -0,0 +1,303 @@
+---
+title: Concepts
+permalink: /schema/fundamentals/concepts
+category: Data Schema
+menuOrder: 1
+subCategory: Fundamentals
+---
+
+Cube.js borrows a lot of terminology from data science and [OLAP
+theory][wiki-olap], and this document is intended for both newcomers and regular
```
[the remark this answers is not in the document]
hassankhan: I think both words still mean the same thing (in this context), but happy to change if you think it should be

keydunov:
Great job @hassankhan. It is a solid foundation and I'm glad we finally have this page.

adnanrahic (approved):

PR #3104, opened 2021-07-13
chore: adding the core server version

vasilev-alex:


packages/cubejs-server-core/src/core/server.ts
```diff
@@ -115,6 +115,8 @@ export class CubejsServerCore {
   public constructor(opts: CreateOptions = {}, protected readonly systemOptions?: SystemOptions) {
     optionsValidate(opts);
 
+    this.coreServerVersion = version;
```
ovr, line 118: I think we can drop this version from the property because we get this version from package.json. m?

PR #3121, opened 2021-07-19
Update DruidQuery.ts

rongfengliang:
bugfix for support like https://github.com/cube-js/cube.js/issues/3109

### DIFF
--- a/packages/cubejs-druid-driver/src/DruidQuery.ts
+++ b/packages/cubejs-druid-driver/src/DruidQuery.ts
@@ -11,7 +11,18 @@ const GRANULARITY_TO_INTERVAL: Record<string, (date: string) => string> = {
   year: date => `DATE_TRUNC('year', ${date})`
 };
 
+class DruidFilter extends BaseFilter {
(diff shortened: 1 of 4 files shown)


rongfengliang:
@ovr  thanks

paveltiunov:
@rongfengliang Thanks for contributing it! Is there any chance you can add the simple test in `druid-driver.test.ts`?

rongfengliang:
> @rongfengliang Thanks for contributing it! Is there any chance you can add the simple test in `druid-driver.test.ts`?

this is a query test so I create another test file `druid-query.test.ts`

paveltiunov:
@rongfengliang It seems `LIKE` is case-sensitive. We probably should use `ICONTAINS_STRING` instead.

rongfengliang:
> @rongfengliang It seems `LIKE` is case-sensitive. We probably should use `ICONTAINS_STRING` instead.

yes. from source code I seen  it use uppercase string so it should be works

ovr:
Thank you @rongfengliang 👍  🍰 🥮

PR #3170, opened 2021-07-29
Support reading AWS config and credentials files

etburke:
Modified Athena drive config to not override aws-sdk reading config/creds from ~/.aws directory

**Check List**

[For example #12]

[Description goes here]


paveltiunov:
@etburke Hey Thomas! Thanks for contributing it! Looks great! The only thing `config.accessKeyId || process.env.CUBEJS_AWS_KEY` and `config.secretAccessKey || process.env.CUBEJS_AWS_SECRET` still should take precedence over `AWS_SDK_LOAD_CONFIG` variable to maintain backward compatibility and to be consistent with what we have in terms of configuration.

PR #3214, opened 2021-08-09
docs(schema): document using `every` and `sql` in `refreshKey` together

hassankhan:
Story details: https://app.clubhouse.io/cube-dev/story/882

### DIFF
--- a/docs/content/Caching/Using-Pre-Aggregations.md
+++ b/docs/content/Caching/Using-Pre-Aggregations.md
@@ -21,37 +21,21 @@ Pre-Aggregations reference in the data schema section][ref-schema-ref-preaggs].
 ## Refresh Strategy
 
 Refresh strategy can be customized by setting the
-[refreshKey][ref-schema-ref-preaggs-refresh-key] property for the
+[`refreshKey`][ref-schema-ref-preaggs-refresh-key] property for the
 pre-aggregation.
 
-The default value of `refreshKey` is `every: '1 hour'`. It can be redefined
-either by providing SQL:
+The default value of [`refreshKey`][ref-schema-ref-preaggs-refresh-key] is
+`every: '1 hour'`. It can be redefined either by overriding the default value of
+the [`every` property][ref-schema-ref-preaggs-refresh-key-every]:
 
 ```javascript
 cube(`Orders`, {
-  // ...
+
+  ...,
 
   preAggregations: {
     amountByCreated: {
-      measures: [amount],
-      timeDimension: createdAt,
-      granularity: `month`,
-      refreshKey: {
-        sql: `SELECT MAX(created_at) FROM orders`,
-      },
-    },
-  },
-});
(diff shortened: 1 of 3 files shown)


docs/content/Caching/Using-Pre-Aggregations.md
```diff
@@ -25,7 +25,29 @@ Refresh strategy can be customized by setting the
...
+      granularity: `month`,
+      refreshKey: {
+        every: `12 hour`,
+      },
+    },
+  },
+});
+```
+
+Or by providing a [`sql` property][ref-schema-ref-preaggs-refresh-key-sql]
```
keydunov, line 49: we need to confirm with core and then mention it here - 

**what is the default value of every when only `sql` is specified?**

I assume it is 1 hour, but worth to double check.

docs/content/Schema/Reference/pre-aggregations.md
```diff
@@ -493,6 +493,30 @@ cube(`Orders`, {
 For possible `every` parameter values please refer to
 [`refreshKey`][ref-cube-refreshkey] documentation.
 
+You can also use `every` with `sql`:
+
+<!-- prettier-ignore-start -->
+[[warning | ]]
+| When using `every` and `sql` together, the value of `every` should match
```
keydunov, line 500: Is it coming from the core team?

keydunov (commented):
There are two important questions to resolve before merging. Highlighted them in the comments.

docs/content/Caching/Using-Pre-Aggregations.md
hassankhan: I'll add a link to https://cube.dev/docs/caching#in-memory-cache-default-refresh-keys which defines the defaults

docs/content/Schema/Reference/pre-aggregations.md
hassankhan: That's what I understood from the task in Notion, will confirm with @ovr

docs/content/Caching/Using-Pre-Aggregations.md
```diff
@@ -25,7 +25,30 @@ Refresh strategy can be customized by setting the
...
+instead, and leaving [`every` unchanged from its' default
```
keydunov, line 50: in-memory and pre-aggs have different default.

hassankhan: I'll confirm with @ovr and update the PR accordingly

keydunov:
@hassankhan Can we add an explanation, so if used together, Cube will check `sql` condition on the cadence defined in every. And if SQL condition is satisfied will refresh the pre-aggregation.

keydunov (changes requested):
Commented above.

keydunov (approved):
Looks good to merge.

PR #3216, opened 2021-08-10
Fix client hangs when loading big responses (node)

vieira:
**Check List**
- [X] Tests for the changes have been added if not covered yet (no new code paths added)
- [X] Docs have been added / updated if required (no docs changes required)

The cloned response stream and the original stream have to be resolved in parallel, otherwise if the response is bigger than the `highWaterMark` (16kB in Node.js) the text stream will hang.

In theory, this can also affect browsers but the `highWaterMark` is usually much bigger there.

See https://github.com/node-fetch/node-fetch/issues/151 and [node-fetch and `highWaterMark`](https://github.com/node-fetch/node-fetch#custom-highwatermark).

### DIFF
--- a/packages/cubejs-client-core/src/index.js
+++ b/packages/cubejs-client-core/src/index.js
@@ -109,11 +109,12 @@ class CubejsApi {
       }
 
       let body = {};
-
+      let text = '';
       try {
-        body = await response.clone().json();
+        text = await response.text();
+        body = JSON.parse(text);
       } catch (_) {
-        body.error = await response.text();
+        body.error = text;
       }
 
       if (body.error === 'Continue wait') {


packages/cubejs-client-core/src/index.js
```diff
@@ -109,11 +109,17 @@ class CubejsApi {
...
-
+      let json = {};
+      let text = '';
+      // XXX: The cloned response stream and the original stream have to be
+      // resolved in parallel, otherwise if the response is bigger than the
+      // highWaterMark (16kB in node.js) the text stream will hang.
       try {
-        body = await response.clone().json();
+        const r1 = await response.clone();
+        ([json, text] = await Promise.all([response.json(), r1.text()]));
```
vasilev-alex, line 119: This is not gonna work as if either of the promises fails `Promise.all` fails as well

vasilev-alex (changes requested):

packages/cubejs-client-core/src/index.js
vieira: Hello @vasilev-alex,

Oops, you are of course right. I guess the `continue wait` I was testing this with does come inside a JSON response and does not exercise the catch path.

I guess we can avoid the clone altogether by reading the stream as text and then trying to parse the resulting text as JSON. If it fails, use the already read text for body.error?

I have amended by previous commit with this other approach. It still works for my test cases. What do you think?

vasilev-alex (approved):

tomasrb:
@vasilev-alex how long before this makes it to a published version? We're running into this issue in production.  Thanks!

vasilev-alex:
@tomasrb can release it now

tomasrb:
@vasilev-alex and @vieira, really appreciate the great responsiveness! Will test and validate right away.

tomasrb:
@vasilev-alex looks like after upgrading we're still experiencing this same hanging behavior.  Payload size from invoking a curl /load call bypassing the Node/cube client is 78KB (Node.js version v15.14.0 and cube v0.28.17), should I report a new issue?

vasilev-alex:
@tomasrb yes, please

tomasrb:
@vasilev-alex done: https://github.com/cube-js/cube.js/issues/3235, I'm in the Slack channel so feel free to ping me there if any questions or any workarounds we could use. Thanks!

tomasrb:
@vasilev-alex just fyi I closed 3235, there was an issue on our side because when i re-deployed/restarted our clusters and this fix works, thanks again!

Issue #3210, opened 2021-08-07
Incorrect port in OracleDriver.js


jaipgit:
File `OracleDrive.js` has the code (incorrect `port` definition):
```
this.config = config || {
      user: process.env.CUBEJS_DB_USER,
      password: process.env.CUBEJS_DB_PASS,
      db: process.env.CUBEJS_DB_NAME,
      host: process.env.CUBEJS_DB_HOST,
      port: 1521,
      poolMin: 0,
      poolMax: 50,
    };
```

Should be:
```
    this.config = config || {
      user: process.env.CUBEJS_DB_USER,
      password: process.env.CUBEJS_DB_PASS,
      db: process.env.CUBEJS_DB_NAME,
      host: process.env.CUBEJS_DB_HOST,
      port: process.env.CUBEJS_DB_PORT,
      poolMin: 0,
      poolMax: 50,
    };
```

jaipgit:
I want to make Pull Request. How can I do this?

PR #3253, opened 2021-08-13
correct port definition in OracleDriver.js

jaipgit:
#3210


paveltiunov:
@jaipgit Looks great! Can we probably split name and port changes into 2 different PRs as they seem not related?

PR #3265, opened 2021-08-17
docs: ensure all Data Schema reference pages have consistent table of contents

hassankhan:
Story details: https://app.clubhouse.io/cube-dev/story/896

### DIFF
--- a/docs/content/Schema/Reference/segments.md
+++ b/docs/content/Schema/Reference/segments.md
@@ -12,7 +12,7 @@ redirect_from:
 
 Segments are predefined filters. You can use segments to define complex
 filtering logic in SQL. For example, users for one particular city can be
-treated as a segment.
+treated as a segment:
 
 ```javascript
 cube(`Users`, {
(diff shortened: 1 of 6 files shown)


docs/content/Schema/Reference/segments.md
```diff
@@ -12,7 +12,7 @@ redirect_from:
 
 Segments are predefined filters. You can use segments to define complex
```
hassankhan, line 13: Not really sure how to make this page consistent, as it only has one parameter: `sql`.

Also not sure what to do about the "Segments vs Dimension Filters" section, as it isn't really reference documentation. Could/should we move it to its' own recipe perhaps, and then keep this file purely as reference docs for segments?

docs/content/Schema/Reference/joins.md
```diff
@@ -34,6 +34,23 @@ The semantics of `INNER JOIN` can be achieved with additional filtering. For
 example, a simple check of whether the column value `IS NOT NULL` by using [set
 filter][ref-restapi-query-filter-op-set] satisfies this requirement.
 
+## Naming
+
+The name of a join should match the [name of the external
+cube][ref-schema-cube-naming]. For an example where a `Products` cube is being
```
rpaik, line 40:
```suggestion
cube][ref-schema-cube-naming]. For example when a `Products` cube is being
```

PR #3274, opened 2021-08-18
correction of oracle port definition

jaipgit:
#3210


packages/cubejs-oracle-driver/driver/OracleDriver.js
```diff
@@ -47,7 +47,7 @@ class OracleDriver extends BaseDriver {
       password: process.env.CUBEJS_DB_PASS,
       db: process.env.CUBEJS_DB_NAME,
       host: process.env.CUBEJS_DB_HOST,
-      port: 1521,
+      port: process.env.CUBEJS_DB_PORT,
```
ovr, line 50: I suggest not to do backward compatibility break, let's continue to use 1521 as the default value for port:

```suggestion
      port: process.env.CUBEJS_DB_PORT || 1521,
```

ovr:
Merged, Thank you @jaipgit 👍

PR #3277, opened 2021-08-18
feat: add cubestore helm chart

lvauvillier:
This is a first PR to add helm charts to ease cube deployments.

I made it as generic as possible to cover most of deployment cases.

This PR covers only `cubestore` chart.

I think separating `cubejs` and `cubestore` in two different charts is a more granular design. In the future, we can add  a third umbrella chart `cube-stack` to deploy all the stack (including redis if needed)

This cubestore chart covers all the following use cases:
- remoteDir and external storage
- persistance for local data
- single node deployment or router + workers (default is a router and two workers)
- supports common k8s attributes (affinity, annotation, etc.)
- supports all cubestore configs (no upload, job runners, query timeout, etc.)

See values.yaml file for all possibilities

### Project path

I add this PR to the folder `helm-charts`of the cubejs project. Tell me if there is a better location.

### Known issue

Sometimes, some workers dont initialize properly and get the error message:
```
2021-08-18 18:57:37,934 ERROR [cubestore::cluster] <pid:1> Failed to get warmup partitions: Internal: Connection refused (os error 111)
```

This is due to the fact that, at init time for a worker, the router is not ready. Cubstore dont seems to have a retry strategy.
We can add a workaround using an `initcontainer` to wait for the router before starting the worker.

It will be great If i can have some guidance here to properly write this init check.

### Global config

All the following params are global and will be applied both to the router and workers:
- noUpload
- jobRunners
- querytimeout
- selectWorkers
- walSplitThreshold

I need confirmation if all of them are global or only for router/worker

### Next steps

- Add cubejs chart
- Host a repository in github pages


helm-charts/cubestore/templates/pvc.yaml
```diff
@@ -0,0 +1,22 @@
...
+  {{- end }}
+  labels: 
+    app.kubernetes.io/component: remote-storage
+    {{- include "cubestore.selectorLabels" . | nindent 4 }}
+  {{- if .Values.remoteDir.persistence.annotations }}
+  annotations: {{- toYaml .Values.remoteDir.persistence.annotations | nindent 4 }}
+  {{- end }}
+spec:
+  accessModes:
+    - ReadWriteMany
```
vainkop, line 18: @lvauvillier it's always better not to use VPs with ReadWriteMany as well as other shared NFS-like storages vs just using S3 buckets. 
There are always problems with performance & permissions with NFS-like storages. 
All cloud providers have buckets & on prem can easily deploy smth like minio.

helm-charts/cubestore/templates/workers/statefulset.yaml
```diff
@@ -0,0 +1,104 @@
...
+            {{- $workers := list }}
+            {{- range $e, $i := until ($.Values.workers.workersCount | int) }}
+            {{- $workers = append $workers (printf "%s-workers-%d.%s:%d" $fullName $i $headlessServiceName $workerPort) }}
+            {{- end }}
+            - name: CUBESTORE_WORKERS
+              value: {{ join "," $workers | quote }}
+            - name: CUBESTORE_META_ADDR
+              value: {{ printf "%s-router:%d" (include "cubestore.fullname" .) (.Values.router.metaPort | int) }}
+            - name: CUBESTORE_DATA_DIR
+              value: /cube/.cubestore/data
```
vainkop, line 50: @lvauvillier why use local mount path if the preferred method is the use of S3/GCP buckets? I'd make it "if not cloudprovider then" use local path (basically same as https://github.com/cube-js/cube.js/pull/3277/commits/793cc2db48dd728d17b5517feb6dbc95032e93e2#r691975909 )

helm-charts/cubestore/templates/workers/statefulset.yaml
```diff
@@ -0,0 +1,104 @@
...
+            {{- if and (not .Values.cloudStorage.gcp.bucket) (not .Values.cloudStorage.aws.bucket) }}
+            - name: CUBESTORE_REMOTE_DIR
+              value: /cube/data
```
vainkop, line 53: @lvauvillier why use local mount path if the preferred method is the use of S3/GCP buckets? I'd make it "if not cloudprovider then" use local path (basically same as https://github.com/cube-js/cube.js/pull/3277/commits/793cc2db48dd728d17b5517feb6dbc95032e93e2#r691975909 )

adnanrahic:
Hi @lvauvillier ! :wave: 
This is absolutely lovely. Thanks a lot for your contribution!

> I think separating cubejs and cubestore in two different charts is a more granular design. In the future, we can add a third umbrella chart cube-stack to deploy all the stack (including redis if needed)

It's a super clever solution. Adding an umbrella chart with Redis would be the next step.

As mentioned in https://github.com/cube-js/cube.js/pull/3266 it would be great if you could move the Helm files into `examples/helm-charts`.

lvauvillier:
Hi @adnanrahic, files are now in the example folder.

rpaik:
🙏 for this @lvauvillier. I was also discussing this with @adnanrahic, but is it a standard practice to add descriptions and author info in the `Chart.yaml` file or do people also create separate README files? 

This may seem a dumb question (as I'm not an expert on Helm Charts), but it'd be good to have a standard here for future contributions. I'm OK either way and I think it'd be good to add some of the details you already have in the PR description (e.g. bullet points under Cubestore chart, exclusion of external pre-aggregations dbs, etc.) either in `Chart.yaml` or README.

adnanrahic:
I agree with you @rpaik , using a readme to add an explanation is pretty great.

I took the liberty of making this as an example we could use in the future: https://github.com/cube-js/cube.js/blob/examples-kubectl/examples/kubectl/prod/readme.md

Let me know what you think. :smile:

lvauvillier:
You are right. I'm currently working to add a README for both charts

lvauvillier:
All possible options are now listed in both README.
I also added installation instruction and deployment examples.

lvauvillier:
External db for preaggregation are now supported. All configs are now covered :)

adnanrahic:
Hi @lvauvillier !

I've had a look at the Helm charts and ran some tests. It all looks awesome.
I did run into one issue with `CUBEJS_API_SECRET`. I can't seem to set it through the Helm `values.yaml`. I also can't seem to find it at all in the `common-env.tpl`. Am I missing something?

For reference, [here's where I added it in the `Kubectl` sample](https://github.com/cube-js/cube.js/pull/3311/files#diff-39b7342b688124b269b4265166083c528a88673ff431f6028130652b31b8f8efR23-R24).

PR #3288, opened 2021-08-19
fix(@cubejs-client/core): prevent redundant auth updates

### LINKED ISSUE #3190: Prevent redundant auth token server requests when initially loading a page with server Cube queries
Per my conversion with @vasilev-alex on Slack here https://cube-js.slack.com/archives/CC0403RRR/p1627945309042800

Is there an optimal way of preventing several redundant requests to my `jwt` endpoint for authorization on an initial page load when a token does not exist or is expired? For example, I have a page with 7 different graphs I need to load, all of which are separate requests. As a result, my `/auth/cubejs-token`  endpoint that issues a `jwt` token is called 7 times since the token does not exist. My client configuration is straight forward, just checks if the token exists or is expired in `localstorage` , if not fetch a new token. What is the best route to take? Can I somehow delay the client from attempting to fetch another token while requesting one?

```
const cubejsClient = cubejs(async () => {
  let token = ....
  // CHECK IF TOKEN EXISTS IN STORAGE OR IS EXPIRED

  // IF TOKEN DOES NOT EXIST, FETCH NEW
  if (!token) {
    const response = await fetch(`http://127.0.0.1:4001/auth/cubejs-token`, {
      ...
    });
    const jsonResponse = await response.json();
    
    // SET TOKEN VALUE AND STORE IN LOCALSTORAGE
    if (jsonResponse.token) {
      token = jsonResponse.token;

      await Storage.set({
        key: 'cubejsAuthToken',
        value: token
      })
    }
  }

  return `${token}`
}, {
  apiUrl: `${process.env.NEXT_PUBLIC_CUBEJS_API_URL}`
})
```

Here's an image of my network request logs:
image (1)

Requests after initial page load work perfectly since the token is not expired and exists in storage, but that initial page load does request the endpoint quite a few times!

I would like to be able to debounce(?) an auth token request if one is currently being requested for the page so that my server isn't hit with a burst of token requests every time a user visits their dashboard.

vasilev-alex:
Closes #3190

### DIFF
--- a/packages/cubejs-client-core/src/index.js
+++ b/packages/cubejs-client-core/src/index.js
@@ -194,7 +194,7 @@ class CubejsApi {
       return;
     }
     
-    if (typeof this.apiToken === 'function' && !this.updateAuthorizationPromise) {
+    if (typeof this.apiToken === 'function') {
       this.updateAuthorizationPromise = new Promise(async (resolve, reject) => {
         try {
           const token = await this.apiToken();
--- a/packages/cubejs-client-core/src/index.js
+++ b/packages/cubejs-client-core/src/index.js
@@ -10,9 +10,17 @@ let mutexCounter = 0;
 
 const MUTEX_ERROR = 'Mutex has been changed';
 
-const mutexPromise = (promise) => new Promise((resolve, reject) => {
-  promise.then(r => resolve(r), e => e !== MUTEX_ERROR && reject(e));
-});
+function mutexPromise(promise) {
+  return new Promise(async (resolve, reject) => {
+    try {
+      resolve(await promise);
+    } catch (error) {
+      if (error !== MUTEX_ERROR) {
+        reject(error);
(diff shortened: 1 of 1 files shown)


packages/cubejs-client-core/src/index.js
```diff
@@ -10,9 +10,17 @@ let mutexCounter = 0;
 
 const MUTEX_ERROR = 'Mutex has been changed';
 
-const mutexPromise = (promise) => new Promise((resolve, reject) => {
-  promise.then(r => resolve(r), e => e !== MUTEX_ERROR && reject(e));
-});
+function mutexPromise(promise) {
```
ovr, line 13: There is a function `asyncDebounce` in `@cubejs-backend/shared`, probably we need to solve this sort of problem in one way.

ovr (approved):
LGTM 🥮

PR #3319, opened 2021-08-25
feat(cubestore): readiness and liveness probes 

ilya-biryukov:
Available only on router nodes via `/livez` and `/readyz`.
Use `CUBESTORE_STATUS_BIND_ADDR` or `CUBESTORE_STATUS_PORT` for
endpoint configuration. Default is `0.0.0.0:3031`.

Runs as a separate HTTP server to simplify implementation. Main HTTP
server does waits for metastore from remote before it starts listening
on a socket.

The checks are very basic. Liveness check tries to execute a trivial
query. Readiness check after metastore got loaded from remote and
can serve another trivial query.

### DIFF
--- a/rust/cubestore/src/bin/cubestored.rs
+++ b/rust/cubestore/src/bin/cubestored.rs
@@ -1,5 +1,6 @@
 use cubestore::app_metrics;
 use cubestore::config::{Config, CubeServices};
+use cubestore::http::status::serve_status_probes;
 use cubestore::telemetry::track_event;
 use cubestore::util::logger::init_cube_logger;
 use cubestore::util::metrics::init_metrics;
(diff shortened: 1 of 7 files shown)


lvauvillier:
Hi @ilya-biryukov, is the worker warmup failure when router is not ready covered by this PR?
More info here:
https://github.com/cube-js/cube.js/pull/3311#issuecomment-904659361
and workaround here:
https://github.com/cube-js/cube.js/pull/3311#issuecomment-904969646

ilya-biryukov:
Hi @lvauvillier.

Nope. We'll need to tackle this separately.

The failure happens on k8s, right?

lvauvillier:
Yes. It happens at init time when the first worker is deployed and the router is not ready. Cusbestore dont seems having a retry policy here when retrieving metadata at warmup.

Restarting manually the worker pod fix the issue.

A workaround is to add an initContainer with a wait script. But having a hard dependency in startup order is not a good design for k8s. I think workers should fail in this case to be auto restarted by k8s.

sh8mi:
Are these healthcheck endpoints 3031/livez and 3031/readyz only available in router?
When it's the worker, these endpoints are not reachable.

lvauvillier:
@sh8mi yes, probes only checks the router. I'm not sure how to properly check each workers.

PR #3364, opened 2021-09-02
feat: Add ability to pass through generic-pool options

### LINKED ISSUE #3340: Redis pool options not passed through to underlying connection pool.
Several options under `orchestratorOptions.redisPoolOptions` never get passed to the underlying connection pool, and are thus unused. Rather than just PRing the very short fix myself I've created this issue for some discussion around the issue, because I'd also like to take this opportunity to discuss exposing additional options from the underlying pool.,

Points of consideration
- `generic-pool` has `millis` configuration options for `idleTimeout` and `softIdleTimeout` but Cube takes `seconds` options, should this be changed?
- Should we expose other `generic-pool` options that Cube has defaults set for? (`acquireTimeoutMillis`, `evictionRunIntervalMillis`)
- Should we provide a way to pass configuration to _every_ option that `generic-pool` provides, beyond the current cube options?

**Version:**
Definitely in 0.27, 0.28, and master.

**Additional context**

We've had a whole bunch of issues relating to Redis usage, including just ongoing CPU usage relating to idle connections being held open for ages because `generic-pool` will only evict 3 resources per eviction run, and Cube provides no way to configure these options directly. The reason for me opening this issue is obviously that certain options don't actually get passed through and that's clearly a bug, but this additional context is my motivation for asking the broader "what should be exposed?" question.

TRManderson:
**Check List**
- [x] Docs have been added / updated if required (no relevant docs currently exist, am happy to add some)

Resolves #3340, and serves as a proposal for the extensions to `redisPoolOptions` that I talk about in that issue

Adds a new key to `orchestratorOptions.redisPoolOptions` that is passed directly to the `generic-pool` constructor used for creating the Redis connection pool.

Also actually passes through the `softIdleTimeoutSeconds` and `idleTimeoutSeconds` keys that are accepted currently, but not actually passed through to the connection pool. Despite actually passing these through, I'm of the opinion that these should be deprecated in favour of eg `orchestratorOptions.redisPoolOptions.poolOptions.idleTimeoutSeconds`, passing through to `generic-pool` far more directly.


paveltiunov:
@TRManderson Hey Taylor! It seems `idleTimeoutSeconds` and `softIdleTimeoutSeconds` were never working. Thanks for fixing it!

PR #3368, opened 2021-09-02
fix(client-core): keep order of elements within tableColumns

dodgex:
Prefix the key used to build the schema with an underscore to ensure the key is treated as a string and not as a numeric and thus as an array index.

**Check List**

https://github.com/cube-js/cube.js/issues/3367

### DIFF
--- a/packages/cubejs-client-core/src/ResultSet.js
+++ b/packages/cubejs-client-core/src/ResultSet.js
@@ -537,15 +537,15 @@ class ResultSet {
         let currentItem = schema;
 
         yValues.forEach((value, index) => {
-          currentItem[value] = {
+          currentItem[`_${value}`] = {
             key: value,
             memberId: normalizedPivotConfig.y[index] === 'measures'
               ? value
               : normalizedPivotConfig.y[index],
-            children: (currentItem[value] && currentItem[value].children) || {}
+            children: (currentItem[`_${value}`] && currentItem[`_${value}`].children) || {}
           };
 
-          currentItem = currentItem[value].children;
+          currentItem = currentItem[`_${value}`].children;
         });
       }
     });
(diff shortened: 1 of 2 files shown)


packages/cubejs-client-core/src/tests/ResultSet.test.js
```diff
@@ -1224,4 +1224,101 @@ describe('ResultSet', () => {
       );
     });
   });
+
+  describe('tableColumns',()=> {
```
vasilev-alex, line 1228: Looks good! One small thing, could you please move this test to `table.test.js`?

vasilev-alex (changes requested):

packages/cubejs-client-core/src/tests/ResultSet.test.js
dodgex: No problem, i just saw the `ResultSet.test.js` and added it here, did not realize that there are more specific test files too.

Just one question: should i add a new sub-describe or can i just add the test in the top level one?

vasilev-alex: @dodgex if it's just a single test there's no need to wrap it in another `describe`

packages/cubejs-client-core/src/ResultSet.js
```diff
@@ -537,15 +537,15 @@ class ResultSet {
         let currentItem = schema;
 
         yValues.forEach((value, index) => {
-          currentItem[value] = {
+          currentItem['_' + value] = {
```
vasilev-alex, line 540: could you also fix the `eslint` issues?

packages/cubejs-client-core/src/tests/ResultSet.test.js
dodgex: Done! :)

packages/cubejs-client-core/src/ResultSet.js
dodgex: Done, no idea why my editor didnt show it before. just started it to fix and it showed the error :D

PR #3390, opened 2021-09-06
fix(helm-charts): Fix live/ready probe path and expose options

lvauvillier:
Fix live/ready probe path and expose options


adnanrahic (approved):
Looks great! Thanks for your contribution @lvauvillier ! :clap:

PR #3395, opened 2021-09-07
fix(docs): update links to documentation in Playground and API Gateway

hassankhan:
Story details: https://app.clubhouse.io/cube-dev/story/1047

### DIFF
--- a/packages/cubejs-playground/src/components/RollupDesigner/RollupDesigner.tsx
+++ b/packages/cubejs-playground/src/components/RollupDesigner/RollupDesigner.tsx
@@ -295,7 +295,7 @@ export function RollupDesigner({
       return (
         <Paragraph>
           <Link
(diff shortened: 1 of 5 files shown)


packages/cubejs-playground/src/components/RollupDesigner/RollupDesigner.tsx
```diff
@@ -285,7 +285,7 @@ export function RollupDesigner({
           . Please consider removing not additive measures like `countDistinct`
           or `avg`. You can also try to use{' '}
           <Link
-            href="https://cube.dev/docs/schema/reference/pre-aggregations#parameters-type-originalsql"
+            href="https://cube.dev/docs/schema/reference/pre-aggregations#type-originalsql"
```
ovr, line 288: This href doesn't work.

hassankhan: Yep, looks like all H4 tags broke when we moved to MDX, thanks for catching it ❤️

PR #3400, opened 2021-09-08
feat(docs): feedback block

YakovlevCoded:
Preview: https://deploy-preview-3400--cubejs-docs-staging.netlify.app/recipes or any other docs page with a few exceptions


docs/src/components/Button/styles.module.scss
```diff
@@ -0,0 +1,164 @@
+$danger-text: #ef404a;
+$success-text: #59a36c;
+$purple: #7A77FF;
+$white: #FFF;
+$purple-text: #665DE8;
+$cera-pro-font: "Cera Pro", "Avenir Next", Avenir, "Helvetica Neue", Helvetica,
+      Ubuntu, "DejaVu Sans", Arial, sans-serif;
```
hassankhan, line 7: Can we move these to https://github.com/cube-js/cube.js/blob/master/docs/static/styles/variables.scss

hassankhan (commented):
Minor nitpick, great work @YakovlevCoded 🙌

igorlukanin:
@Leonid Yakovlev Great job! Could you please increase the margin between the feedback block and the page content? I think we can safely make ~3 times bigger than currently. Attached the image below:

Screenshot 2021-09-09 at 20 40 57

rchkv:
It looks nice 👍 @YakovlevCoded 

Just a few questions about the workflow (maybe they should address to @igorlukanin):
1. We send a query to the internal analytics for every click on the like/dislike buttons. So, a user might cancel feedback sending but we will get an event. Is that correct? It seems to me that we should do this after clicking on the `Send` button.  (please see the gif in the attachment)
2. Will the message be mandatory or not?

https://user-images.githubusercontent.com/18264667/132799818-f7ddc12c-0ccb-4127-a04d-563b47958270.mov

YakovlevCoded:
> increase

increased

YakovlevCoded:
> @YakovlevCoded Great job! Could you please increase the margin between the feedback block and the page content? I think we can safely make ~3 times bigger than currently. Attached the image below:
> 
> Screenshot 2021-09-09 at 20 40 57

increased

YakovlevCoded:
> It looks nice 👍 @YakovlevCoded
> 
> Just a few questions about the workflow (maybe they should be addressed to @igorlukanin):
> 
> 1. We send a query to the internal analytics for every click on the like/dislike buttons. So, a user might cancel feedback sending but we will get an event. Is that correct? It seems to me that we should do this after clicking on the `Send` button.  (please see the gif in the attachment)
> 2. Will the message be mandatory or not?
> 
>  Screen.Recording.2021-09-10.at.09.29.12.mov

We send an event when the user click to like or dislike because the user may don't leave a comment and click to the send button :( And Anton asked to be able to send an empty form.

rchkv:
> We send an event when a user clicks to like or dislike because the user may don't leave a comment or click on the send button :( And Anton asked to be able to send an empty form.

I see. But I think we can send the like/dislike event with the form posting event. In my opinion, an action cannot be considered complete until the user has explicitly submitted the form.

igorlukanin:
@rchkv1. We cannot expect

hassankhan (approved):

PR #3454, opened 2021-09-20
fix(cubestore): fix Docker Hub repository name

ilya-biryukov:
We used to publish CubeStore image into the CubeJS repository.


ovr (approved):
LGTM, my bad :(

PR #3524, opened 2021-10-07
Fix: updated dynamic-angular-dashboard example's dependencies

kiuru:
Dynamic Angular Dashboard example was getting error during the build process, because it contains old dependencies.

Also I got an error when I was trying to add a chart to the Dashboard, which related to the following issue with apollo-client: apollographql/apollo-client/issues/8566.


rpaik:
Thanks @kiuru for your first contribution 😄

rpaik:
@vasilev-alex can you help with the review?

Issue #3433, opened 2021-09-15
GraphQL query Support


lvauvillier:
## Context

**Is your feature request related to a problem? Please describe.**
A previous issue was opened on this topic (#1765) but was closed due to query format design and ResultSet issue.

As we already run a working graphql API on top of cubeJS REST API on production, I wanted to share here the design API we choosed and how we solved it.

I hope this will start an official support of graphQL API.

**Describe the solution you'd like**
There is two possible solutions:
- A graphql API on top of the existing REST API (the GraphQL API act as a "proxy")
- A standalone graphql API (a new graphQLApiGateway module in the cubejs server core?)

We will here choose the first solution to benefit of existing optimization, security, query rewrite etc...

### Context

GraphQL APIs returns data in the same shape as they were requested. This constraint dont allow a perfect mapping to the existing REST api and its response.

⚠️ The ResultSet object will not be usable for data blending and has to be rewrited for graphQL responses. (Or provide a GraphQLResultSet object). This task is not hard to do, if we flatten the graphQL object, most of the existing part of the current ResultSet class can be reused.

### Example

```graphql

{
    events (
       filters: [
       {
            member: "publishedAt"
            operator: inDateRange
            values: ["2021-08-01", "2021-08-30"]
        }
    ]) {
        measures {
            count
            pageViewsCount
        }
        dimensions {
            country
        }
    }
}
```

Response
```json
{
    "events": [{
        "measures": {
            "count": 145,
            "pageViewCount": 35
        },
        "dimensions": {
            "country": "US"
        } 
    },
    {
        "measures": {
            "count": 23,
            "pageViewCount": 9
        },
        "dimensions": {
            "country": "FR"
        } 
    },
    {
        "measures": {
            "count": 45,
            "pageViewCount": 12
        },
        "dimensions": {
            "country": "DE"
        } 
    }]
}

```

### Spec

```graphql
enum CubeFilterOperator {
  afterDate
  beforeDate
  contains
  equals
  gt
  gte
  inDateRange
  lt
  lte
  notContains
  notEquals
  notInDateRange
  notSet
  set
}

enum CubeOrder {
  asc
  desc
}

enum CubeGranularity {
  second
  minute
  hour
  day
  week
  month
  year
}

input CubeFilterInput {
  member: String!
  operator: CubeFilterOperator!
  values: [String]
  or: [CubeFilterInput!]
  and: [CubeFilterInput!]
}

type Root {
[...]

PR #3555, opened 2021-10-18
feat(gateway): Add GraphQL endpoint

lvauvillier:
This PR is an implementation of the GraphQL spec described in issue #3433

I think i covered all the main functionalities of the REST API.

## Description

A new endpoint `/cubejs-api/graphql` is added in the gateway.
All security middlewares (checkAuth, queryRewrite, etc.) are still used.

In dev mode, graphiql UI is available.

## How it works
- GraphQL schema is automatically generated using the compilerApi.
- GraphQL AST info are used to generate cubequery at root cube entity level
- Results are remapped to the graphql query shape

## Warning
Current implementation will raise an error for a `continueWait` response. Long polling is not implemented (and not possible with this API design).

## Example

original query:

```json
{
  "measures": ["Stories.count"],
  "dimensions": ["Stories.category"],
  "filters": [{
    "member": "Stories.isDraft",
    "operator": "equals",
    "values": ["No"]
  }],
  "timeDimensions": [{
    "dimension": "Stories.time",
    "dateRange": ["2015-01-01", "2015-12-31"],
    "granularity": "month"
  }],
  "limit": 100,
  "offset": 50,
  "order": {
    "Stories.time": "asc",
    "Stories.count": "desc"
  },
  "timezone": "America/Los_Angeles"
}
```

is equivalent to graphql query:
```graphql
{
  load(limit: 100, offset: 50, timezone: "America/Los_Angeles") {
    stories(granularity: {time: month}, orderBy: { time: asc, count: desc }, where: {isDraft: {equals: "NO"}, time: {inDateRange: ["2015-01-01", "2015-12-31"]}}) {
      count
      category
      time
    }
  }
}

```

directives `@skip` and `@include` are supported:

```graphql
query GetStories($byCategory: Boolean = false) {
  stories(granularity: { time: month }) {
      count
      category @include(if: $byCategory)
      time
  }
} 
```

### DIFF
--- a/packages/cubejs-api-gateway/package.json
+++ b/packages/cubejs-api-gateway/package.json
@@ -33,6 +33,7 @@
     "body-parser": "^1.19.0",
     "chrono-node": "^2.2.6",
     "express-graphql": "^0.12.0",
+    "graphql": "^15.3.0",
     "graphql-scalars": "^1.10.0",
     "jsonwebtoken": "^8.3.0",
     "jwk-to-pem": "^2.0.4",
@@ -60,9 +61,6 @@
     "should": "^13.2.3",
     "supertest": "^4.0.2",
     "typescript": "~4.1.5"
-  },
-  "peerDependencies": {
-    "graphql": "^14.7.0 || ^15.3.0"
   },
   "license": "Apache-2.0",
   "eslintConfig": {
--- a/packages/cubejs-api-gateway/package.json
+++ b/packages/cubejs-api-gateway/package.json
@@ -32,11 +32,15 @@
     "@hapi/joi": "^15.1.1",
     "body-parser": "^1.19.0",
     "chrono-node": "^2.2.6",
+    "express-graphql": "^0.12.0",
+    "graphql-scalars": "^1.10.0",
     "jsonwebtoken": "^8.3.0",
     "jwk-to-pem": "^2.0.4",
     "moment": "^2.24.0",
     "moment-timezone": "^0.5.27",
     "node-fetch": "^2.6.1",
+    "nexus": "^1.1.0",
+    "querystring": "^0.2.1",
     "ramda": "^0.27.0",
     "uuid": "^8.3.2"
   },
@@ -56,6 +60,9 @@
     "should": "^13.2.3",
     "supertest": "^4.0.2",
     "typescript": "~4.1.5"
+  },
+  "peerDependencies": {
+    "graphql": "^14.7.0 || ^15.3.0"
   },
   "license": "Apache-2.0",
   "eslintConfig": {
--- a/packages/cubejs-api-gateway/src/gateway.ts
+++ b/packages/cubejs-api-gateway/src/gateway.ts
@@ -283,19 +283,23 @@ export class ApiGateway {
     const guestMiddlewares = [];
 
     app.use(`${this.basePath}/graphql`, userMiddlewares, (() => {
-      let schema; // memoized schema
+      let cache;
       return (async (req, res) => {
-        if (!schema) {
+        if (!cache) {
           const metaConfig = await this.getCompilerApi(req.context).metaConfig({
             requestId: req.context.requestId,
           });
-          schema = makeSchema(metaConfig);
+          cache = {
+            metaConfig,
+            schema: makeSchema(metaConfig)
+          };
         }
 
         return graphqlHTTP({
-          schema,
+          schema: cache.schema,
           context: {
             req,
+            metaConfig: cache.metaConfig,
             apiGateway: this
           },
           graphiql: getEnv('nodeEnv') !== 'production'
--- a/packages/cubejs-api-gateway/src/gateway.ts
+++ b/packages/cubejs-api-gateway/src/gateway.ts
@@ -282,30 +282,26 @@ export class ApiGateway {
     // @todo Should we pass requestLoggerMiddleware?
     const guestMiddlewares = [];
 
-    app.use(`${this.basePath}/graphql`, userMiddlewares, (() => {
-      let cache;
-      return (async (req, res) => {
-        if (!cache) {
-          const metaConfig = await this.getCompilerApi(req.context).metaConfig({
-            requestId: req.context.requestId,
-          });
-          cache = {
-            metaConfig,
-            schema: makeSchema(metaConfig)
-          };
-        }
+    app.use(`${this.basePath}/graphql`, userMiddlewares, async (req, res) => {
+      const compilerApi = this.getCompilerApi(req.context);
+      let schema = compilerApi.getGraphQLSchema();
+      if (!schema) {
+        const metaConfig = await compilerApi.metaConfig({
+          requestId: req.context.requestId,
+        });
+        schema = makeSchema(metaConfig);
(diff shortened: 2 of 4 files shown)


paveltiunov:
@lvauvillier On a side note how would query with members from multiple cubes look like?

lvauvillier:
@paveltiunov I just push an update and it works nicely :)

The new filters works great with autocompletion and we can now join cubes.

Filters are autogenerated based on the member type with some changes from the REST API (heavily inspired from hasura and prisma filtering api):
- it accepts number values for number types (instead of string values everywhere)
- `notSet` is removed in favor of `{ set:  false }`
- `equals` and `notEquals` accept a single value, but we have now `in`and `notIn` for multiple values
- `AND` and `OR` fields for boolean operator
- filters can be set at `load` level or `<cubeName>` level. Using `load` level allow user to use boolean operator accros multiple cubes.

Example:
```graphql
{
  load(offset: 20, limit: 5, where: {users: {firstName: {notEquals: "Golden"}}}) {
    orders(granularity: {createdAt: day}, where: {OR: [{createdAt: {afterDate: "2021-01-01"}}, {status: {equals: "shipped"}}]}, orderBy: {createdAt: asc}) {
      count
      status
      createdAt
    }
    users {
      firstName
    }
  }
}
```
will generate:
```json
{
  "measures": [
    "Orders.count"
  ],
  "dimensions": [
    "Orders.status",
    "Users.firstName"
  ],
  "timeDimensions": [
    {
      "dimension": "Orders.createdAt",
      "granularity": "day"
    }
  ],
  "order": {
    "Orders.createdAt": "asc"
  },
  "limit": 5,
  "offset": 20,
  "filters": [
    {
      "member": "Users.firstName",
      "operator": "notEquals",
      "values": [
        "Golden"
      ]
    },
    {
      "or": [
        {
          "member": "Orders.createdAt",
          "operator": "afterDate",
          "values": [
            "2021-01-01"
          ]
        },
        {
          "member": "Orders.status",
          "operator": "equals",
          "values": [
            "shipped"
          ]
        }
      ]
    }
  ]
}
```

paveltiunov:
@lvauvillier Looks really good and it seems to play nicely overall. What do you think about putting granularity inside the time dimension field like this?
```
{
  load(offset: 20, limit: 5, where: {users: {firstName: {notEquals: "Golden"}}}) {
    orders(where: {OR: [{createdAt: {afterDate: "2021-01-01"}}, {status: {equals: "shipped"}}]}, orderBy: {createdAt: asc}) {
      count
      status
      createdAt { day }
    }
    users {
      firstName
    }
  }
}
```
The reason for it is generally speaking you can query multiple granularities of the same time dimension. It's rare but still valid use case.

Could you please also rebase so we can run the test and be ready to merge?

lvauvillier:
@paveltiunov If we add granularity subfields to time dimensions, we need to keep the ability use the time dimension without granularity.

We can add a new `raw` field to achieve this (i'm not sure of the name)

With granularity

```graphql
{
  load(offset: 20, limit: 5, where: {users: {firstName: {notEquals: "Golden"}}}) {
    orders(where: {OR: [{createdAt: {afterDate: "2021-01-01"}}, {status: {equals: "shipped"}}]}, orderBy: {createdAt: asc}) {
      count
      status
      createdAt {
        day
      }
    }
    users {
      firstName
    }
  }
}
```

Without granularity

```graphql
{
  load(offset: 20, limit: 5, where: {users: {firstName: {notEquals: "Golden"}}}) {
    orders(where: {OR: [{createdAt: {afterDate: "2021-01-01"}}, {status: {equals: "shipped"}}]}, orderBy: {createdAt: asc}) {
      count
      status
      createdAt {
        raw
      }
    }
    users {
      firstName
    }
  }
}
```

I also think about using a union type, but scalars are not allowed: we can't use a simple `DateTime | Granularity`.
Union types also require to use fragment notation in the query:
```graphql
createdAt {
  ...on Granularity {
    day
  }
}
```

paveltiunov:
@lvauvillier Yeah. This is a really great point! `raw` might work. Other way is to have explicit dimensions resolver like this to avoid reserved names introduction:
```graphql
{
  load(offset: 20, limit: 5, where: {users: {firstName: {notEquals: "Golden"}}}) {
    orders(where: {OR: [{createdAt: {afterDate: "2021-01-01"}}, {status: {equals: "shipped"}}]}, orderBy: {createdAt: asc}) {
      count
      status
      dimensions {
        createdAt
      }
    }
    users {
      firstName
    }
  }
}
```

paveltiunov:
@lvauvillier I guess we can go with `raw` for now and see how it progresses over time. Could you please add it? Thanks!

lvauvillier:
@paveltiunov I just pushed a new version. I finally chose the name `value` instead of `raw`.
It now works with multiple granularities at the same time.

I also added the missing `quarter` granularity and add a nonNull constraint to all members.

New graphQL spec:

```graphql
input DateTimeFilter {
  afterDate: String
  beforeDate: String
  equals: String
  in: [String]
  inDateRange: [String]
  notEquals: String
  notIn: [String]
  notInDateRange: [String]
  set: Boolean
}

input FloatFilter {
  contains: Float
  equals: Float
  in: [Float]
  notContains: Float
  notEquals: Float
  notIn: [Float]
  set: Boolean
}

input StringFilter {
  contains: String
  equals: String
  in: [String]
  notContains: String
  notEquals: String
  notIn: [String]
  set: Boolean
}

type TimeDimension {
  day: DateTime!
  hour: DateTime!
  minute: DateTime!
  month: DateTime!
  quarter: DateTime!
  second: DateTime!
  value: DateTime!
  week: DateTime!
  year: DateTime!
}

enum OrderBy {
  asc
  desc
}

type Query {
  load(limit: Int, offset: Int, renewQuery: Boolean, timezone: String, where: RootWhereInput): [Result!]!
}

type Result {
  <cubeName>(orderBy: <CubeName>OrderByInput, where: <CubeName>WhereInput): <CubeName>Members!
  ...
}

type <CubeName>Members {
  <member>: String! | Float!
<timeMember>: TimeDimension!
  ...
}

input RootWhereInput {
  AND: [RootWhereInput!]
  OR: [RootWhereInput!]
  <cubeName>: <CubeName>WhereInput
  ...
}

input <CubeName>WhereInput {
  AND: [<CubeName>WhereInput!]
  OR: [<CubeName>WhereInput!]
  <member>: FloatFilter | DateTimeFilter | StringFilter
  ...
}

input <CubeName>OrderByInput {
 <member>: OrderBy
 ...
}
```

paveltiunov:
@lvauvillier We're almost there! The small thing I noted is you introduced cache for the schema. If we cache it we should respect multi-tenancy. To not withhold merging of this big PR we can consider either removing caching (I realize it won't be suitable for production workload yet) or storing it as a part of `CompilerApi`. You can introduce a setter for that. Should be an easy fix.

lvauvillier:
@paveltiunov Oh, you are right, schemas can de dynamic if a `repositoryFactory` is defined.

It also raises the point of extending graphiQL UI to add the ability to define security contexts like in the playground. This can be part of a new PR.

paveltiunov:
@lvauvillier Hey Luc! Looks great! Thanks again for such a big effort!

Issue #3534, opened 2021-10-12
cubejs-schema-compiler BaseQuery.dimensionsJoinCondition not working in Clickhouse


antnmxmv:
**Describe the bug**
Clickhouse not supports OR conditions in JOIN ON statements ([link to docs](https://clickhouse.com/docs/en/sql-reference/statements/select/join/#on-section-conditions)). So when [BaseQuery.dimensionsJoinCondition](https://github.com/cube-js/cube.js/blob/269c20bf9749ec95c0a6dfe4c1932c079bfa0ea0/packages/cubejs-schema-compiler/src/adapter/BaseQuery.js#L680) method used clickhousedb throws exception.

**To Reproduce**
1. Use Clickhouse as source database
2. Build query that uses BaseQuery.fullKeyQueryAggregate

**Version:**
master

**Additional context**
[There is the issue](https://github.com/ClickHouse/ClickHouse/issues/22509#issuecomment-812538170) in clickhouse repository containing question about handling with OR in join conditions. It contains answer with options from a collaborator

PR #3600, opened 2021-10-28
fixed clickhouse joins in cubejs-schema-compiler

antnmxmv:
**Check List**

#3534

### DIFF
--- a/packages/cubejs-schema-compiler/src/adapter/ClickHouseQuery.js
+++ b/packages/cubejs-schema-compiler/src/adapter/ClickHouseQuery.js
@@ -92,6 +92,16 @@ export class ClickHouseQuery extends BaseQuery {
     return `parseDateTimeBestEffort(${value})`;
   }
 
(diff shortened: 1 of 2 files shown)


paveltiunov:
@antnmxmv Hey Anton! Looks great! I believe it'd be beneficial to add a test for this use case.

antnmxmv:
> @antnmxmv Hey Anton! Looks great! I believe it'd be beneficial to add a test for this use case.

Hi! Test for this usecase has been added already and there was the comment line saying that it crashes because of issue that this PR covers. But after I fixed the issue, test has started to fail because of timestamp type that is not contains time zone. So after I make sure, that query has been generated and expected result appeared (but time zone), I have changed the comment above this test.

paveltiunov:
@antnmxmv Do you think you can fix the test and enable it? It would be really beneficial.

antnmxmv:
@paveltiunov done.

paveltiunov:
Hey @antnmxmv ! Looks good to me! Thanks for contributing it!

PR #3606, opened 2021-10-30
fix(client-core): dayjs locale conflict

LvtLvt:
**Check List**

Hii^^ I'm using cubeJS for analytics API^^ 
First of all, thank you for creating a really great platform. thanks to you, I'm so into cubeJs these days..

if I change the locale value of dayjs on the front-end, no value is displayed when granularity is week.

Currently, if you look at the ResultSet code, the weekStart must be set to 1 to work properly.

However, the value can be modified by the client as much as possible (https://day.js.org/docs/en/i18n/changing-locale)

so it seems necessary to guide or cope with it. 

To reproduces the problem, you can overwrite the locale of dayjs with ko or another value and test it.
please refer to the test code..

thank you^^

### DIFF
--- a/packages/cubejs-client-core/src/ResultSet.js
+++ b/packages/cubejs-client-core/src/ResultSet.js
@@ -12,12 +12,8 @@ import { aliasSeries } from './utils';
 dayjs.extend(quarterOfYear);
 
 // When granularity is week, weekStart Value must be 1. However, since the client can change it globally (https://day.js.org/docs/en/i18n/changing-locale)
-// So the proxy code below has been added.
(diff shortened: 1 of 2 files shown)


paveltiunov:
@LvtLvt Hey David! Looks great! The only minor thing it seems `Proxy` can be replaced by a function as we don't use much of API from `dayjs` currently.

LvtLvt:
@paveltiunov Hii^^ thank you for review. I think you are right. proxy code seems overkill.. your comment helped me a lot..

paveltiunov:
@LvtLvt Looks great! Thanks for contributing it!

PR #3626, opened 2021-11-03
docs(caching): document that Cube Store can use EC2 instance metadata for IAM credentials

hassankhan:
Story details: https://app.shortcut.com/cube-dev/story/1095

### DIFF
--- a/docs/content/Caching/Running-in-Production.mdx
+++ b/docs/content/Caching/Running-in-Production.mdx
@@ -215,6 +215,22 @@ services:
       - cubestore_router
 ```
 
+### AWS
+
+Cube Store can retrieve security credentials from instance metadata
+automatically. This means you can skip defining the
+`CUBESTORE_AWS_ACCESS_KEY_ID` and `CUBESTORE_AWS_SECRET_ACCESS_KEY` environment
+variables.
+
+<WarningBox>
+
+Cube Store currently does not take the key expiration time returned from
+instance metadata into account. For this reason, the key expiration time is
+defined by `CUBESTORE_AWS_CREDS_REFRESH_EVERY_MINS`, which is set to `180` by
+default.
+
+</WarningBox>
+
 ## Security
 
 Cube Store currently does not have any in-built authentication mechanisms. For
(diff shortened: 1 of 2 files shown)


docs/content/Configuration/Environment-Variables-Reference.md
```diff
@@ -166,17 +164,18 @@ databases [in this guide][link-connecting-to-db].
...
-| `CUBESTORE_GCP_KEY_FILE`          | The path to a JSON key file for connecting to Google Cloud. Required when using Google Cloud Storage    | [A valid Google Cloud JSON key file][link-gcp-creds-json]                               |
-| `CUBESTORE_GCS_BUCKET`            | The name of a bucket in GCS. Required when using GCS                                                    | A valid bucket name in the Google Cloud account                                         |
-| `CUBESTORE_GCS_SUB_PATH`          | The path in a GCS bucket to store pre-aggregations. Optional                                            | -                                                                                       |
+| Environment variable                     | Description                                                                                                                             | Possible Values                                                                         |
```
rpaik, line 167: @hassankhan, would it make sense to create a separate column in the table to indicate whether an env. variable is required vs. option? Not sure if that's overkill. Feel free to ignore this 😄

hassankhan: It was previously discussed and we decided against it because it would add too many columns to the data. Definitely need to revisit how this page looks in the future though

docs/content/Caching/Running-in-Production.mdx
```diff
@@ -215,6 +215,22 @@ services:
...
+automatically. This means you can skip defining the
+`CUBESTORE_AWS_ACCESS_KEY_ID` and `CUBESTORE_AWS_SECRET_ACCESS_KEY` environment
+variables.
+
+<WarningBox>
+
+Cube Store currently does not take the key expiration time returned from
+instance metadata into account. For this reason, the key expiration time is
+defined by `CUBESTORE_AWS_CREDS_REFRESH_EVERY_MINS`, which is set to `180` by
+default.
```
hassankhan, line 230:
```suggestion
instance metadata into account; instead the refresh duration for the key is
defined by `CUBESTORE_AWS_CREDS_REFRESH_EVERY_MINS`, which is set to `180` by
default.
```

PR #3664, opened 2021-11-15
fix(client-core): nullish measure values [v0.29]

vasilev-alex:
All `undefined`/`null` measure values will now be converted to `0`


vprus:
@vasilev-alex : Alex, could I ask what was the motivation for this breaking change? In UI, it's important to distinguish between "no data yet" and "all our business metrics are 0", and this change forces the metrics to be '0'. After that, UI has no way to distinguish between those semantically different situations. Am I missing something?

PR #3681, opened 2021-11-17
refactor(@cubejs-backend/mysql-driver): return promise from `MySqlDriver` pool `destroy()` method

rongfengliang:
destroy method would return promise not to call it.

### DIFF
--- a/packages/cubejs-mysql-driver/src/MySqlDriver.ts
+++ b/packages/cubejs-mysql-driver/src/MySqlDriver.ts
@@ -112,7 +112,7 @@ export class MySqlDriver extends BaseDriver implements DriverInterface {
         }
         return true;
       },
-      destroy: (connection) => promisify(connection.end.bind(connection))(),
+      destroy: (connection) => promisify(connection.end.bind(connection)),
     }, {
       min: 0,
       max: process.env.CUBEJS_DB_MAX_POOL && parseInt(process.env.CUBEJS_DB_MAX_POOL, 10) || 8,


packages/cubejs-mysql-driver/src/MySqlDriver.ts
```diff
@@ -112,7 +112,7 @@ export class MySqlDriver extends BaseDriver implements DriverInterface {
         }
         return true;
       },
-      destroy: (connection) => promisify(connection.end.bind(connection))(),
+      destroy: (connection) => promisify(connection.end.bind(connection)),
```
ovr, line 115: Hello!

What issue do you try to resolve? Because it looks strange, because `destroy` should return a Promise, but without calling promisified connection. end, it will return function. for example:

````js
const pool = { destroy: (connection) => connection.end };
pool.destroy({ end: async () => {} })
````

result:

```js
async () => {}
```

Thanks

rongfengliang: I do some tests with MySQL like databases.  when fetching database tables it will block. and I do some tests find that caused by connection pool destroy method . 

and  in ` testConnection `
https://github.com/cube-js/cube.js/blob/master/packages/cubejs-mysql-driver/src/MySqlDriver.ts#L173
it call like this ` await (<any> this.pool)._factory.destroy(conn);` 
so I think this would  return promise  method with one argument .

rongfengliang: and  this  also works 
```
const pool = { destroy: (connection) => connection.end() };
````

Comment (rongfengliang) [reply packages/cubejs-mysql-driver/src/MySqlDriver.ts:115]: or  destroy would like this 

```code
destroy: (connection) => {promisify(connection.end.bind(connection))()}
```

rongfengliang: @ovr 

same as JDBC driver code

https://github.com/cube-js/cube.js/blob/master/packages/cubejs-jdbc-driver/src/JDBCDriver.js#L81

```code
      destroy: async (connection) => promisify(connection.close.bind(connection)),
```
I saw the current connect pool destroy method not all the same

igorlukanin:
This is most likely stale. Please feel free to reopen if needed.

PR #3706, opened 2021-11-19
feat(helm-charts): Add sqlPort config to enable SQL Api

lvauvillier:
- Add `config.sqlPort` config to enable SQL Api (disabled by default)
- Remove image.tag v* restriction.
- Lint

### DIFF
--- a/examples/helm-charts/cubejs/values.yaml
+++ b/examples/helm-charts/cubejs/values.yaml
@@ -33,6 +33,10 @@ config:
   ##
   apiPort: 4000
 
+  ## The port for a Cube.js deployment to listen to SQL connections on
+  ##
+  sqlPort:
+
   ## If true, enables development mode.
   ##
   devMode: false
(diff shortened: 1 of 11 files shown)


rpaik:
@adnanrahic can you help with the review?

examples/helm-charts/cubejs/values.yaml
```diff
@@ -33,6 +33,10 @@ config:
   ##
   apiPort: 4000
 
+  ## The port for a Cube.js deployment to listen to SQL connections on
+  ##
+  sqlPort:
```
adnanrahic, line 38: Do you think it's useful to add the default SQL port here? Similar to how you added the default API port.
I'm just asking. If you think it is not needed, let's skip it.

adnanrahic (approved):
Thanks so much for this @lvauvillier ! Looks good, we can merge it once we decide on if we need a default value for the `sqlPort`. :smile:

examples/helm-charts/cubejs/values.yaml
lvauvillier: According to the doc: https://cube.dev/docs/backend/sql, the sql API is enabled only if we set a port.

I reflected here this behavior: no value disabled, value enabled. The API is disabled by default.

PR #3841, opened 2021-12-22
Build cubestore Docker image for arm64

maximlopin:


.github/workflows/rust-cubestore.yml
```diff
@@ -59,14 +59,26 @@ jobs:
 
   docker-image-latest:
```
ovr, line 60: We need to do the same thing for publishing & master workflows

Issue #3831, opened 2021-12-20
Adding session tokens and more flexibility to the AthenaDriver constructor


lrichomme:
**Is your feature request related to a problem? Please describe.**
I'm doing a project where I must use AWS session tokens and change the s3 output location per query (using driverFactory in the config file). And the current state of the Athena Driver doesn't support session tokens and uses only the env value CUBEJS_AWS_S3_OUTPUT_LOCATION for the output location.

**Describe the solution you'd like**
I would need the driver to take the session token in its credentials and the possibility to set a new s3 ouput location when calling the driver.

**Describe alternatives you've considered**
I don't think there are any alternatives with my problem as I'm not able to use aws credentials without session tokens. And I really need to be able to change the s3 ouput location with each query as I'm accessing different databases with different accounts meaning different sets of s3 buckets.

**Additional context**
I already modified cubejs' AthenaDriver constructor in my local version, and everything works. I also modified every other parameter so they can be set as well when calling the driver in case someone would need to do this later. Also I guessed that adding a new parameter in the config variable would mean setting a new variable in the .env file so I added CUBEJS_AWS_SESSION_TOKEN in it. Here's the modified constructor :
```javascript
constructor(config = {}) {
    super();

    this.config = {
      credentials: {
        accessKeyId: config.accessKeyId || process.env.CUBEJS_AWS_KEY,
        secretAccessKey: config.secretAccessKey || process.env.CUBEJS_AWS_SECRET,
        // Added session token support
        sessionToken: config.sessionToken || process.env.CUBEJS_AWS_SESSION_TOKEN
      },
      // Added the possibility to set region, s3 output location and workgroup values
      region: config.region || process.env.CUBEJS_AWS_REGION,
      S3OutputLocation: config.s3OutputLocation || process.env.CUBEJS_AWS_S3_OUTPUT_LOCATION,
      workGroup: config.workGroup || (process.env.CUBEJS_AWS_ATHENA_WORKGROUP || 'primary'),
      ...config,
      pollTimeout: (config.pollTimeout || getEnv('dbPollTimeout')) * 1000,
      pollMaxInterval: (config.pollMaxInterval || getEnv('dbPollMaxInterval')) * 1000,
    };

    this.athena = new AWS.Athena(this.config);
  }
```
If you're okay with this request I can follow the contributing guide, I'm just not sure how to run tests properly.

ivan-vdovin:
@lrichomme 
Thank you!
Let me add @ovr so he can take a look.

[...]

PR #3850, opened 2021-12-24
Add session token and some flexibility to Athena Driver's constructor

lrichomme:
**Check List**

#3831 

[Description goes here]

### DIFF
--- a/packages/cubejs-athena-driver/driver/AthenaDriver.js
+++ b/packages/cubejs-athena-driver/driver/AthenaDriver.js
@@ -22,6 +22,8 @@ class AthenaDriver extends BaseDriver {
       pollTimeout: (config.pollTimeout || getEnv('dbPollTimeout')) * 1000,
       pollMaxInterval: (config.pollMaxInterval || getEnv('dbPollMaxInterval')) * 1000,
     };
+
+    this.athena = new AWS.Athena(this.config);
   }
 
   readOnly() {
--- a/packages/cubejs-athena-driver/driver/AthenaDriver.js
+++ b/packages/cubejs-athena-driver/driver/AthenaDriver.js
@@ -13,16 +13,15 @@ class AthenaDriver extends BaseDriver {
       credentials: {
         accessKeyId: config.accessKeyId || process.env.CUBEJS_AWS_KEY,
         secretAccessKey: config.secretAccessKey || process.env.CUBEJS_AWS_SECRET,
+        sessionToken: config.sessionToken || undefined
       },
-      region: process.env.CUBEJS_AWS_REGION,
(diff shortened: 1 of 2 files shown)


packages/cubejs-athena-driver/driver/AthenaDriver.js
```diff
@@ -13,10 +13,11 @@ class AthenaDriver extends BaseDriver {
       credentials: {
         accessKeyId: config.accessKeyId || process.env.CUBEJS_AWS_KEY,
         secretAccessKey: config.secretAccessKey || process.env.CUBEJS_AWS_SECRET,
+        sessionToken: config.sessionToken || undefined
       },
-      region: process.env.CUBEJS_AWS_REGION,
-      S3OutputLocation: process.env.CUBEJS_AWS_S3_OUTPUT_LOCATION,
-      workGroup: process.env.CUBEJS_AWS_ATHENA_WORKGROUP || 'primary',
+      region: config.region || process.env.CUBEJS_AWS_REGION,
```
ovr, line 18: I don't think that we need to change these lines, because below these changes there is a line: `...config` which will override these variables if the user passes config in the constructor.

wdyt?

packages/cubejs-athena-driver/driver/AthenaDriver.js
```diff
@@ -13,10 +13,11 @@ class AthenaDriver extends BaseDriver {
       credentials: {
         accessKeyId: config.accessKeyId || process.env.CUBEJS_AWS_KEY,
         secretAccessKey: config.secretAccessKey || process.env.CUBEJS_AWS_SECRET,
+        sessionToken: config.sessionToken || undefined
```
ovr, line 16: What issue do you try to solve with || undefined? If `sessionToken` is not defined in `config` it will be undefined anyway.

packages/cubejs-athena-driver/driver/AthenaDriver.js
ovr: BTW: You can use:


````js
let config = {
  credentials: {
    accessKeyId: 'myaccessid',
    secretAccessKey: 'myaccesskey',
    sessionToken: 'mysessiontoken',
  }
};

let driver = new AthenaDriver(config);
````

lrichomme: I tested the workaround and it works, thanks! I didn't know that ...config would overwrite things. So I guess no change is needed.

Also sorry for the late response I didn't have access to a PC until now to test this.

lucafaggianelli:
Is this PR going forth? Why it stopped?
I'd like as well to use temporary credentials for Athena via env vars, if needed I can take over the PR.

Luca

paveltiunov:
@lucafaggianelli It would be great if you can pick it up!

PR #3855, opened 2021-12-28
chore: unhardcore show variables queries

vasilev-alex:


rust/cubesql/cubesql/src/mysql/mod.rs
```diff
@@ -70,71 +69,62 @@ impl Backend {
             _ => false,
         };
 
-        if query_lower.eq("set autocommit=1, sql_mode = concat(@@sql_mode,',strict_trans_tables')")
+        let stmt = parse_sql_to_statement(&query)?;
```
ovr, line 72: This will not work, because the parser still doesn't support some cases.

show full tables from db
show columns

ovr:
I am going to close this PR due https://github.com/cube-js/cube.js/pull/3960/files

Thanks

PR #3868, opened 2022-01-03
fix(schema-compiler): Handle `null` Values in Security Context

joealden:
**Check List**

(if it is okay, please validate the above, as doing so should be a lot easier for someone that already has things setup locally)

This change should allow `null` to exist as a value of a property within the security context. Prior to this change, if a `null` value was encountered in the security context, a `Cannot create proxy with a non-object as target or handler` error would be thrown, as `typeof null === "object"`. This PR should resolve this by explicitly checking for `null` before passing the property value to `contextSymbolsProxy`.


paveltiunov:
Hey @joealden ! Thanks for fixing that!

PR #3877, opened 2022-01-05
fix(redshift-driver): Use Redshift-specific schema query.

### LINKED ISSUE #3876: Use Redshift-appropriate query for information schema discovery
**Describe the bug**

The `RedshiftDriver` class inherits `createSchemaIfNotExists()` from the `BaseDriver` class, by way of the `PostgresDriver` class.  The implementation of that method doesn't error on Redshift, but [it only returns results for schemas owned by the querying user](https://stackoverflow.com/a/49280051). This results in breakage when Cube.js accesses Redshift with credentials that can access a schema but is not the owner of the schema, because the existing implementation of `createSchemaIfNotExists()` will return no results, and will then attempt to create the pre-aggregation schema. If the calling user does not have permission to create schemas, this will result in an error being thrown without any explanation of which query failed.

**To Reproduce**
1. Provision a Redshift cluster
2. Create a `pre_aggregation` schema in that cluster
3. Create a user in that cluster, with access to the `pre_aggregation` schema (e.g., `GRANT USAGE ON SCHEMA pre_aggregation TO ...;`
4. Create a Cube.js app with a one-cube schema containing a simple pre-aggregation
5. Run the Cube.js app configured for scheduled refresh, to use the Redshift cluster as the default source DB and to use credentials for the user created in step 3
6. Watch it fail with a permission error

**Expected behavior**

Expected Cube.js to function correctly when running as a user with appropriate permissions in the pre-aggregation schema (`SELECT`, `CREATE TABLE`, etc).

**Screenshots**

n/a

**Version:**

0.29.15

emmanuel:
Fixes #3876.

**Check List**

Fixes #3876 

Override `BaseDriver.createSchemaIfNotExists()` in `RedshiftDriver` with a schema-discovery query that is compatible with Redshift information schema permission semantics. See details in #3876.

### DIFF
--- a/packages/cubejs-redshift-driver/src/RedshiftDriver.ts
+++ b/packages/cubejs-redshift-driver/src/RedshiftDriver.ts
@@ -22,6 +22,19 @@ export class RedshiftDriver extends PostgresDriver<RedshiftDriverConfiguration> 
     super(options);
   }
 
+  /**
+   * @param {string} schemaName
+   * @return {Promise<Array<unknown>>}
+   */
+   async createSchemaIfNotExists(schemaName: string) {
+    const schemaExistsQuery = `SELECT nspname FROM pg_namespace where nspname = ${this.param(0)}`;
+    const schemas = await this.query(schemaExistsQuery, [schemaName])
+    if (schemas.length === 0) {
+      return this.query(`CREATE SCHEMA IF NOT EXISTS ${schemaName}`);
+    }
+    return null;
+  }
+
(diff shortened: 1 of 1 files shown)


rpaik:
cc/ @paveltiunov

paveltiunov:
@emmanuel Thanks for contributing that! Could you please fix TS compilation errors?

rpaik:
@emmanuel not sure if you saw @paveltiunov's comment 👆?

KSDaemon:
Closing in a favor of https://github.com/cube-js/cube/pull/9363

PR #3889, opened 2022-01-07
test: enable TypeScript transpiling when running Jest. Pilot for `@cubejs-backend/query-orchestrator`

cristipp:
**Check List**

Enables Typescript transpiling when running jest. This allows running unit tests from WebStorm.

Pilot for @cubejs-backend/query-orchestrator, we should consider adopting this for all js/ts packages.

### DIFF
--- a/packages/cubejs-query-orchestrator/package.json
+++ b/packages/cubejs-query-orchestrator/package.json
@@ -18,9 +18,9 @@
     "tsc": "tsc",
     "watch": "tsc -w",
     "test": "npm run unit && npm run integration",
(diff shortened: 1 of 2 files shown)


ovr (commented):
We special don't use ts-node, because it can be dangerous thing in some places

cristipp:
ts-jest is different than and does not depend on ts-node. See https://github.com/kulshekhar/ts-jest/blob/main/package.json.

ovr (approved):

hassankhan (approved):

PR #3957, opened 2022-01-20
fix: postgres driver add enum support

lxb:
The user defined enum types were being dropped in the Postgres driver.
As per the postgres documentation these are stored internally as four
bytes on disk however the translations occur at query time to the label
from pg_enum which specifies:

"The length of an enum value's textual label is limited by the NAMEDATALEN
setting compiled into PostgreSQL; in standard builds this means at most
63 bytes." [1]

For the purposes of cubejs it should be acceptable to coerce into a
string.

[1] https://www.postgresql.org/docs/9.1/datatype-enum.html

https://github.com/cube-js/cube.js/issues/3946


ovr:
Thank you @lxb for your contribution! Great work 🍰 🚀

lxb:
@ovr Thanks for the quick turn around. You guys rock :)

PR #3993, opened 2022-01-26
chore: improve type coverage and convert files to TypeScript

hassankhan:
**Check List**


packages/cubejs-backend-shared/src/types.ts
```diff
@@ -0,0 +1,10 @@
+export interface FileContent {
+  fileName: string;
+  content: string;
+  isModule?: boolean;
+}
+
+export interface SchemaFileRepository {
```
ovr, line 7: We already have this type in `@cubejs-backend/server-core`

hassankhan: We do, but we needed access to that type from the `@cubejs-backend/schema-compiler` package. I wasn't comfortable making `server-core` a (dev)dependency of `schema-compiler`, but happy to do that if you prefer.

PR #4056, opened 2022-02-08
cubestore: Use 'path style' bucket for custom S3 regions

saul:
**Check List**

N/A

This change uses 'path style' buckets for the Cube Store S3 backend when a custom region is specified. This allows Cube Store to use on-premise S3 solutions by specifying the S3 endpoint in the `CUBESTORE_S3_REGION` environment variable.


paveltiunov:
Hey @saul ! Thanks for contributing it! Looks great! I'm wondering if a custom region always implies a path style flag? AFAIK there're regions not on the list in this s3 library however those are still AWS. If it's the case probably makes sense to have a separate env variable for that.

saul:
We're using the MinIO backend instead as it provides the same functionality as this PR. Many thanks

PR #4095, opened 2022-02-16
Added support for array search and text vector search

shlok97:
**Check List**

[For example #12]

[Description goes here]


rpaik:
@paveltiunov @ovr could you help with the review?

paveltiunov:
@shlok97 Hey Shlok! Thanks for contributing this one! I guess if we want to come up with the support for new filter operators we need to figure out the plan on supporting those in different databases Cube supports and also in Cube Store. Do you have a sense of how those operators should be implemented in other databases Cube supports?

rpaik:
@shlok97 just following up on this...

shlok97:
The operator I added is only specific to postgressql, let me see What can
be done to make this work

PR #4122, opened 2022-02-18
docs(schema): add instructions on how to use `FILTER_PARAMS` with extended cubes

hassankhan:


docs/content/Schema/Advanced/Code-Reusability-Extending-Cubes.mdx
```diff
@@ -93,6 +93,17 @@ cube(`PageViews`, {
 });
 ```
 
+## Usage with FILTER_PARAMS
+
+If the base cube is using [`FILTER_PARAMS`][ref-schema-ref-cube-filter-params]
+in any `sql` property, then extending cubes can do one of two things:
+
+- Override the `sql` property in each target cube.
+
```
rpaik, line 102: does it make sense to add an "Or" between the two items?

PR #4130, opened 2022-02-22
chore(testing): Updates postgres tests to use csv instead of postgres dump

cristipp:
**Check List**

N/A

Updates postgres tests to use github-events-2015-01-01.1000.csv instead of postgres dump.
Goal: All e2e tests should use the same data regardless of the source db.

Note: CI will pass only after https://github.com/cube-js/testing-fixtures/pull/1 is merged.


buntarb (approved):
It was impossible to implement this without changing snap files? As far as I understand these changes related with the changed data, so maybe it's better to keep the same data in new files?

cristipp:
@buntarb The number of rows in the test file was reduced from 337903(?) to 1000. The snapshots had to be updated accordingly.

cristipp:
Thanks for the feedback. Merging.

PR #4135, opened 2022-02-23
docs(schema): improve `rollupJoin' example

hassankhan:


docs/content/Schema/Reference/pre-aggregations.mdx
```diff
@@ -252,6 +257,29 @@ cube('Orders', {
 });
 ```
 
+<InfoBox>
+
+`rollupJoin` is not required to join cubes from the same data source; instead
```
rpaik, line 262:
```suggestion
`rollupJoin` is not required to join cubes from the same data source; instead, you can
```

Issue #4139, opened 2022-02-24
Dremio driver does not accept CUBEJS_DB_NAME


mlilius:
**Describe the bug**
The value of CUBEJS_DB_NAME is ignored. It defaults to 'default'. This causes an error if you don't have a default space defined in Dremio.

**To Reproduce**
1. Setup dremio driver
2. Use a dremio server without a default space
3. Set the CUBEJS_DB_NAME environment variable to a space that does exist.
4. Error happens

**Expected behavior**
It is expected that the value of CUBEJS_DB_NAME should be used.

**Screenshots**
N/A

**Version:**
v0.29.28

paveltiunov:
Hey @mlilius ! Could you please provide an example of what configuration is provided and what doesn't work as expected?

PR #4140, opened 2022-02-24
Dremio driver default database fix and documentation

mlilius:
**Check List**

#4139


paveltiunov:
Hey @mlilius ! Could you please elaborate on a change a little bit? It seems env would always take precedence over configuration provided through `cube.js` config file which is not quite correct.

rpaik:
@mlilius just following up on this....

igorlukanin:
I think this change doesn't make much sense.

PR #4157, opened 2022-03-01
fix(compiler): add strictness to booleans in definition

zpencerq:
**Check List**

Updates schema definitions to be strict on all boolean fields. This aligns with the schema defined in [`cubejs-client-core`](https://github.com/cube-js/cube.js/blob/master/packages/cubejs-client-core/index.d.ts). Previously it was valid to pass along the string `"false"` to fields like `shown` in the cube measure definition, this would be reflected in API responses to clients. As a result, the client library wouldn't effectively filter on just `measure.isVisible` due of truthiness semantics in JS.

This is a bit of a blind change across the board so some review of the other boolean fields should be done before merging.


paveltiunov:
Hey @zpencerq ! Great catch! Looks good to me! Thanks for contributing it!

PR #4191, opened 2022-03-09
fix(helm-charts): add missing config for sqlAPI

lvauvillier:
- Add missing config env vars for sqlAPI
- Remove unnecessary service config to keep the it as simple as possible and allow to expose both http and api ports
- fix s3OutputLocation in values.yaml


rpaik:
@adnanrahic can you help with this PR as well? 🙏

adnanrahic (approved):

PR #4239, opened 2022-03-18
refactor(api-gateway): add query type, response format

dmitrika:
**Check List**

Variable `query` had `any` type, after changing it to actual typing noticed a few errors raised by TS compiler and fixed them.
Also changed npm to yarn in package json to remove warning when run them, like:
```
❯ yarn test
yarn run v1.22.17
$ npm run unit
npm WARN lifecycle The node binary used for scripts is /var/folders/21/cwdmycgd6jl_l5ngd53h6ktc0000gn/T/yarn--1647596852752-0.47937816308405234/node but npm is using /Users/dmitrika/.nvm/versions/node/v14.19.0/bin/node itself. Use the `--scripts-prepend-node-path` option to include the path for the node binary npm was executed with.
```

### DIFF
--- a/packages/cubejs-api-gateway/package.json
+++ b/packages/cubejs-api-gateway/package.json
@@ -14,9 +14,9 @@
   "main": "dist/src/index.js",
   "typings": "dist/src/index.d.ts",
   "scripts": {
-    "test": "npm run unit",
+    "test": "yarn run unit",
     "unit": "jest --coverage dist/test",
-    "build": "rm -rf dist && npm run tsc",
+    "build": "rm -rf dist && yarn run tsc",
     "tsc": "tsc",
     "watch": "tsc -w",
     "lint": "eslint src/* test/* --ext .ts,.js",
--- a/packages/cubejs-api-gateway/src/gateway.ts
+++ b/packages/cubejs-api-gateway/src/gateway.ts
@@ -200,7 +200,7 @@ class ApiGateway {
         res: this.resToResultFn(res)
       });
     }));
-    
+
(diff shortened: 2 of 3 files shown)


packages/cubejs-api-gateway/package.json
```diff
@@ -14,9 +14,9 @@
   "main": "dist/src/index.js",
   "typings": "dist/src/index.d.ts",
   "scripts": {
-    "test": "npm run unit",
+    "test": "yarn run unit",
```
ovr, line 17: For this type of change, it's better to change it across the whole repository, because different variants in multiple places can produce misunderstandings.

Do you have a strong opinion that we need to use `yarn` instead of `npm`? I don't think that it's an important change, because npm is a default package manager which is installed on all systems with node.js

ovr (approved):
LGTM excluding yarn -> npm replace.

packages/cubejs-api-gateway/package.json
dmitrika: No strong opinion from my side

Only issue with it, I have warning in the console, since we run yarn test from root directory and then yarn calls npm. Like:
```
❯ yarn test
yarn run v1.22.17
$ npm run unit
npm WARN lifecycle The node binary used for scripts is /var/folders/21/cwdmycgd6jl_l5ngd53h6ktc0000gn/T/yarn--1647596852752-0.47937816308405234/node but npm is using /Users/dmitrika/.nvm/versions/node/v14.19.0/bin/node itself. Use the `--scripts-prepend-node-path` option to include the path for the node binary npm was executed with.
```

I'll revert it to avoid confusion 👍

PR #4253, opened 2022-03-22
feat(schema-compiler): Allow filtering with or/and in FILTER_PARAMS.

tchell:
The BaseGroupFilter needs to implement the `filterParams` and
`conditionSql` methods order to support using or/and filters with
FILTER_PARAMS.

**Check List**

If `FILTER_PARAMS` were used in a cube's `sql` it would not allow filtering with `or`/`and` filters and would crash at query time because `filtersProxy` was not a method on `BaseGroupFilter`. Now when resolving the FILTER_PARAMS filters any OR/AND filters are able to be added correctly.


paveltiunov:
@tchell Hey Tanner! Thanks for contributing that!

PR #4295, opened 2022-03-31
fix: adding a query parameter in Playground requires one extra click to type in search field

MrFlashAccount:
**Check List**

this PR add focusing in input on opening the menu and on typing within the list


packages/cubejs-playground/src/QueryBuilder/MemberDropdown.tsx
```diff
@@ -117,8 +118,25 @@ export default function MemberMenu({
...
+      onClick={() => {
+        // we need to delay focusing since React needs to render <Menu /> first :)
+        setTimeout(() => {
+          searchInputRef.current?.focus();
+        });
+      }}
       overlay={
         <Menu
+          onKeyDown={(e) => {
+            if (['ArrowDown', 'ArrowUp'].includes(e.key)) {
```
tenphi, line 134: Looks weird. We do focus on every key except these two? It should break keyboard navigation hard.

MrFlashAccount: I'm not sure it's weird, though. When we navigate the menu and can't find the needed item, we can start typing, and even if the input isn't focused, the text will be typed in the input. This is a nice pattern :)

https://user-images.githubusercontent.com/61194245/161109725-f9780a93-7382-4f79-bf83-8532e0e68cfd.mov

MrFlashAccount: Although, there is actually no keyboard navigation except navigation within the list. I've tested.

MrFlashAccount:
I've found a bug, it doesn't work with filters.

packages/cubejs-playground/src/QueryBuilder/MemberDropdown.tsx
tenphi: The pattern is good. But keyboard navigation includes: Space, Enter, PageUp, PageDown, Up, Down, Right, Left (yeah, horizontal, even in vertical layout). Maybe we should test an event for a character and only in that case do the focus.

tenphi (approved):

PR #4310, opened 2022-04-03
docs(cube-cloud): add new page for Alerts functionality

hassankhan:


docs/content/Cube-Cloud/Developer-Tools/Alerts.mdx
```diff
@@ -0,0 +1,98 @@
...
+  />
+</div>
+
+## Alert Types
+
+Cube Cloud lets you set up alerts so you can be notified when there are issues
+with any of the deployments in your account. There are three different types of
+alerts:
+
+- API outages: These are triggered if the REST/GraphQL/SQL APIs become
```
duskpoet, line 50: were you able to verify that? I have a feeling it actually might work for REST endpoint only, or REST and GraphQL

duskpoet (commented):
I think it's worth mentioning, that we send an email only once per specific alert, and when it resolves we send the resolve email

docs/content/Cube-Cloud/Developer-Tools/Alerts.mdx
hassankhan: I wasn't able to verify, was hoping you'd confirm 😀

duskpoet (approved):

docs/content/Cube-Cloud/Developer-Tools/Alerts.mdx
```diff
@@ -0,0 +1,106 @@
...
+Cube Cloud has three different types of alerts:
+
+<InfoBox>
+
+Cube Cloud only sends an outage email **once** per configured alert; when the
+issue resolves, a resolution email is also sent.
+
+</InfoBox>
+
+- API outages: These are triggered if the deployment's API becomes unavailable
```
rpaik, line 56: Would it make sense to replace "if" with "when" in this list?

hassankhan:
Note before merging, the image URLs need to be updated first

docs/content/Cube-Cloud/Developer-Tools/Alerts.mdx
igorlukanin: Since the outages are the events that shouldn't happen on a regular basis, IMO, putting "if" is okay. However, I don't have any hard opinion on this.

Issue #4369, opened 2022-04-12
Use tsd for type testing


reify-thomas-smith:
Issues #4160 (PR #4366) and #4202 involve changing some of the declared types in the API. To minimize the risk of regressions (i.e. type errors in code that previously did not produce them), it would be helpful to explicitly test API types. I'm testing out [tsd](https://github.com/SamVerschueren/tsd), and it seems to work quite well, but unfortunately I ran into dependency conflicts, and someone more familiar with the codebase would probably be in a better position to resolve them.

reify-thomas-smith:
I am interested in working on this issue. However, it is probably blocked until some more work is done on the build system, starting with #4541.

Issue #4202, opened 2022-03-10
TypeScript type inference for query results


reify-thomas-smith:
**Is your feature request related to a problem? Please describe.**
Motivating example (which actually won't work due to #4160):

```ts
const results = await api.load({
  measures: ["Foo.count"],
  dimensions: ["Bar.id", "Bar.name"],
} as const);
```

We can infer that the type of `results` is:

```ts
ResultSet<{
  ["Foo.count"]: string | number | boolean,
  ["Bar.id"]: string | number | boolean,
  ["Bar.name"]: string | number | boolean,
}>
```

But `api.load()` is declared as returning `ResultSet<any>`.

**Describe the solution you'd like**

In the above example, `api.load` should return the more specific type. This would require some moderately clever type definitions. If there is interest in this, I can take a crack at it.

**Describe alternatives you've considered**
Not inferring the type, manually defining the correct return type, and casting the result.

paveltiunov:
Hey @reify-thomas-smith ! Do you mean some type of code generation that can provide result set types based on Cube Data Schema?

reify-thomas-smith:
I mean declaring a more complex type for e.g. `useCubeQuery` so that TypeScript does all the work. Here's a slightly stripped-down version of the wrapper I'm currently using:

```ts
// The cube.js functions won't accept readonly queries.
// See https://github.com/cube-js/cube.js/issues/4160
type DeeplyReadonly<T> = {
  readonly [K in keyof T]: DeeplyReadonly<T[K]>;
};

// As Query, but supporting type inference for dimension and measure names.
type TypedQuery<AttributeNames extends string> = DeeplyReadonly<
  Omit<Query, "measures" | "dimensions">
> & {
  readonly measures?: readonly AttributeNames[];
  readonly dimensions?: readonly AttributeNames[];
};

type CubeRecord<AttributeNames extends string> = {
  [k in AttributeNames]: string | number | boolean;
};

export function useTypedCubeQuery<AttributeNames extends string>(
  query: TypedQuery<AttributeNames>,
  options?: UseCubeQueryOptions
): UseCubeQueryResult<CubeRecord<AttributeNames>> {
  return useCubeQuery<CubeRecord<AttributeNames>>(query as Query, options);
}

export function loadTyped<AttributeNames extends string>(
  cubejsApi: CubejsApi,
  query: TypedQuery<AttributeNames>,
  options?: LoadMethodOptions
[...]

PR #4375, opened 2022-04-13
[WIP] Detailed query types

reify-thomas-smith:
**Check List**
- [x] Tests for the changes have been added if not covered yet (see below)

Implement #4202.

This PR incorporates #4366.

1. The `Query` type is now generic. `Query<'foo' | 'bar'>` is a query with dimension, measures, and time dimensions `foo` and `bar`.
2. `load`, `subscribe`, and `useCubeQuery` will now “look into” the type of the passed query/queries and return the narrowest possible type.

The upshot is that if you write a const query, and pass it into e.g. `load`, then you will get a result whose type has exactly the dimensions and measures that you put in the query.

These changes should in principle be backward-compatible:

- `Query` with no type parameter is equivalent to `Query<string>`, which preserves the prior typing. (The same is true for the time dimension types that needed to be modified.)
- `load`, `subscribe`, and `useCubeQuery` will still accept any input they did before and return exactly the same value. The new declared return types are narrower than the old ones.
- The behavior of `useCubeQuery` with an explicit first type parameter works as before, *unless* you pass an explicit type parameter of `unknown`, which makes no sense.

However, any time a general type (such as `any`) is replaced by a more specific type, it is possible that some code that is not safely typed but did compile before will require a cast to compile afterward.

Notes:

- I improved the types in `@cubejs-client/core` and `@cubejs-client/react`, but not elsewhere (e.g. not in `@cubejs-client/vue`). I have not used those other packages, but I'd be happy to work with someone more familiar with them to incorporate these improvements.
- This PR should not change any actual behavior, so there is no need to update the existing tests. However, it would be nice to add some new type-oriented tests. I have been testing this locally using [`tsd`](https://github.com/SamVerschueren/tsd), and I think that might be a good thing to add to the existing test infrastructure (#4369), but there were dependency conflicts and I didn't include it in this PR.

### DIFF
--- a/packages/cubejs-client-core/index.d.ts
+++ b/packages/cubejs-client-core/index.d.ts
@@ -1029,7 +1029,7 @@ declare module '@cubejs-client/core' {
     load<T extends DeeplyReadonly<Query | Query[]>>(
       query: T,
       options?: LoadMethodOptions,
-      callback?: LoadMethodCallback<ResultSet>,
+      callback?: LoadMethodCallback<ResultSet<QueryResultType<T>>>,
     ): void;
     load<T extends DeeplyReadonly<Query | Query[]>>(
       query: T,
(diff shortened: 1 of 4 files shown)


reify-thomas-smith:
Marking as WIP because I need to revisit how it handles time dimensions.

reify-thomas-smith:
Closing in favor of #4446.

PR #4446, opened 2022-04-26
New implementation of detailed types

reify-thomas-smith:
**Check List**

#4202

A reimplementation of #4375.

This implementation does not make the `Query` type generic. Instead, it leans entirely on `QueryRecordType`. Although this makes `QueryRecordType` more complex, it's probably easier to follow on the whole, and it's certainly a less intrusive change. This implementation should also produce correct types for time dimensions (e.g. inferring `Foo.createdAt.day`), which was a challenge under the prior PR.

As in the prior PR:

- There are no functional changes, so there are no updates to the tests.
- Only `core` and `react` get the augmented types, since I'm not familiar with the other front-end packages.
- The change should be backward-compatible, *except* that when a narrower type can be inferred, it will be. When no narrower type can be inferred, `load()` and company will fall back to `any`.

### DIFF
--- a/packages/cubejs-client-core/index.d.ts
+++ b/packages/cubejs-client-core/index.d.ts
@@ -808,6 +808,36 @@ declare module '@cubejs-client/core' {
     total?: boolean;
   }
 
+  export type QueryRecordType<T extends DeeplyReadonly<Query | Query[]>> =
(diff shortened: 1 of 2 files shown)


paveltiunov:
Looks good to me! Thanks for contributing it @reify-thomas-smith !

PR #4459, opened 2022-04-29
Examples update frontend hf

igorlukanin:
After installing the wrapper, the app breaks for some reason:

Screenshot 2022-04-29 at 11 53 34


revuwem (approved):
There was a bug with setting favicon. Please, update `cube-example-wrapper` package to latest version.

PR #4470, opened 2022-04-29
feat(cubesql): Support pg_type_is_visible postgres udf

MazterQyou:
**Check List**

This PR adds support for `pg_catalog.pg_type_is_visible` UDF used by Superset.
Additionally, pg_type definitions were refactored in order to use a macro which reduces errors and gets rid of duplicate code.


rust/cubesql/cubesql/src/sql/types.rs
```diff
@@ -43,15 +43,15 @@ impl ColumnType {
             ColumnType::Timestamp => PgTypeId::TIMESTAMP,
             ColumnType::Double => PgTypeId::NUMERIC,
             ColumnType::List(field) => match field.data_type() {
-                DataType::Binary => PgTypeId::ArrayBytea,
```
ovr, line 46: https://doc.rust-lang.org/1.0.0/style/style/naming/README.html

Enum variants should use CamelCase for naming

rust/cubesql/cubesql/src/compile/snapshots/cubesql__compile__tests__pg_type_is_visible.snap
```diff
@@ -0,0 +1,61 @@
+---
+source: cubesql/src/compile/mod.rs
+expression: "execute_query(\"\n                SELECT t.oid, pg_catalog.pg_type_is_visible(t.oid) isv\n                FROM pg_catalog.pg_type t\n                ORDER BY t.oid ASC;\n                \".to_string(),\n            DatabaseProtocol::PostgreSQL).await?"
+---
++-------+-------+
```
ovr, line 5: Could you please add the type name to this query? It will be much easier to understand the test logic from testing fixtures. Thanks

ovr (changes requested):

rust/cubesql/cubesql/src/sql/types.rs
MazterQyou: The reason behind the change is that all out PgTypeId values are uppercase and the newly added macro gets rid of duplicates in definitions and functions (see [rust/cubesql/pg-srv/src/pg_type.rs](https://github.com/cube-js/cube.js/pull/4470#diff-25a4e4e7200cb58df7b3b8203a0a9070195395a93e7c524dd13135de88c81a56)).

If I keep things as is, the compiler will complain about the names of the defined constants, so while keeping things like this may not be the best way to follow naming style (although most of our enums already were uppercase so I imagine this isn't a big deal), it unifies our PgTypeId enum style and gets the job done.

rust/cubesql/cubesql/src/compile/snapshots/cubesql__compile__tests__pg_type_is_visible.snap
MazterQyou: 👍

Issue #1091, opened 2020-09-14
Moving to TypeScript


ovr:
Hey!

In 2020 TypeScript started to be a standard (exclude flow in fb internals 😄 ) for every project, It helps use to:

- Speedup development (autocompletion, easy to know how to implement drivers by interfaces/abstract classes)
- Protect errors
- Generate true compatibility code for matrix of compatibility versions of Node.JS platform

## High level iteration plan:

Due to the fact, that cube.js is not a small project, it's expected to have a plan, How to move it to TypeScript.

1. Generate declarations for every package automatically (1-2 base packages, it's needed to write driver on TypeScript)
2. Introduce first driver fully on top of TS
3. Move packages from 1st step to TS
4. Move all packages to TS

## Base packages

- [ ] cubejs-query-orchestrator
- [ ] cubejs-schema-compiler
- [ ] cubejs-server-core

## All scope

// todo

## Non code

- [ ] ESLint (support typescript)
- [ ] Docs (a little bit of examples, how to debug?)
- [ ] Tests

Thanks

PR #4475, opened 2022-05-02
[WIP] Convert cubejs-client-react to TypeScript

reify-thomas-smith:
**Check List**

#1091


vasilev-alex:
Hi @reify-thomas-smith! Thanks for your PR.

However, we can't accept it at this moment as there's a blocker on our side. We use TypeDoc for docs generation, and we use the manually maintained `index.d.ts` for this purpose. In case of migration that would be impossible.

reify-thomas-smith:
Closing as this will require substantially more work, which in turn depends on other things.

PR #4507, opened 2022-05-08
chore(cubestore):fix workers config for cluster tests

waralex:
**Check List**

When testing sql queries in cluster mode, the `select_workers` parameter was missing in the workers config. Because of this, all requests were executed on one worker, which did not allow catching errors that occur when executing requests in parallel.
Now the `cubesql::cluster::unsorted_merge_assertion` test does not pass, and this is normal (the bug was that it passed with the actually error exists). This error is fixed here  https://github.com/cube-js/arrow-datafusion/pull/27. 
Accordingly, this PR should be merged after datafusion PR


paveltiunov (approved):

cristipp (commented):
LGTM

PR #4509, opened 2022-05-09
Add iam_role arn to Redshift Unload

### LINKED ISSUE #4505: Redshift Driver Requires aws api key and secret
**Describe the bug**
The Redshift driver forces the population of the `CUBEJS_DB_EXPORT_BUCKET_AWS_KEY` and `CUBEJS_DB_EXPORT_BUCKET_AWS_SECRET`.

This is useful in development, however, during deploy that information is available to the sdk under the hood.  I didnt see the same behavior with the `CUBESTORE_S3_BUCKET` and `CUBESTORE_S3_REGION` settings, those work as expected.

I think the relevant lines are here:

https://github.com/cube-js/cube.js/blob/bef623f1d76f934c9469ffca0523ecb32988ec18/packages/cubejs-redshift-driver/src/RedshiftDriver.ts#L53-L59

**To Reproduce**
1. set `CUBEJS_DB_EXPORT_BUCKET_TYPE=s3`
2. set `CUBEJS_DB_EXPORT_BUCKET_AWS_REGION=<region>`
3. set `CUBEJS_DB_EXPORT_BUCKET=<bucket_name>`
4. Grant `read/write` to the execution role of whatever is running cubjs for the s3 bucket
5. Deploy without the `CUBEJS_DB_EXPORT_BUCKET_AWS_KEY` and `CUBEJS_DB_EXPORT_BUCKET_AWS_SECRET` vars
6. Logs produce the following:

```
Error: Unsupported configuration exportBucket, some configuration keys are empty: keyId,secretKey
```

**Expected behavior**
The `aws sdk` should pick up the permissions from the assigned task role, and this should work.

magno32:
This makes the keys `keyId` and `secretKey` optional when `unloadArn` is set (with env: `CUBEJS_DB_REDSHIFT_UNLOAD_ARN`).  `secretKey` and `keyId` are still required when `unloadArn` is absent since Redshift needs to use credentials to perform `UNLOAD`.  If the redshift `iam_role` is used, keys should not be required for aws api operations if the proper task-roles are defined.

This also adds the `CUBEJS_DB_REDSHIFT_UNLOAD_ARN` environment variable, referenced in `env.ts` by `redshiftBucketUnloadArn`

This change will _prefer_ using the `iam_role` ARN over provided credentials for the unload operation.

**Check List**
  - Tested against our development AWS stack
  - Tested using the Redshift Smoke test that was added in #4545 

Fixes #4505


cristipp:
Thanks for the contribution! I've been swamped with some other work, so only a quick comment: we'll need to add a test for this feature. I might get some time tomorrow to add such a test for current Redshift driver, similar with the Athena / BigQuery tests we've recently added. Then you can tweak that to complete this PR with proof that it actually works.

Beware of bugs in the above code; I have only proved it correct, not tried it. -- Donald Knuth

magno32:
Thanks, I'll watch for changes referencing this, and take care of that.

packages/cubejs-redshift-driver/src/RedshiftDriver.ts
```diff
@@ -5,14 +5,24 @@ import crypto from 'crypto';
...
-  keyId: string,
-  secretKey: string,
   region: string,
 }
 
+interface RedshiftDriverExportArnAWS extends RedshiftDriverExportRequiredAWS{
+  unloadArn?: string,
+}
+
+interface RedshiftDriverExportKeySecretAWS extends RedshiftDriverExportRequiredAWS{
```
cristipp, line 18: Hmm, looks to me like we always need key/secret, because that's what Cube needs to use to download the files. How about:

```
interface RedshiftDriverExportAWS {
  bucketType: 's3',
  bucketName: string,
  keyId: string,
  secretKey: string,
  region: string,
  unloadArn?: string,
}
```

cristipp (commented):
Good start. A few questions:

* Why is ARN needed? Not an AWS expert, just curious :)

* How can Cube download S3 files without credentials.

* To test this, please add a test to https://github.com/cube-js/cube.js/blob/master/packages/cubejs-testing/test/smoke-redshift.test.ts

The way I run the tests is 
```
$ env $(cat ~/.env.redshift | xargs) yarn smoke:redshift
<snip>
 PASS  dist/test/smoke-redshift.test.js (14.563 s)
  redshift
    ✓ query measure (2035 ms)

Test Suites: 1 passed, 1 total
Tests:       1 passed, 1 total
Snapshots:   1 passed, 1 total
Time:        14.638 s
```

where

```
$ cat ~/.env.redshift
CUBEJS_DB_HOST=<EIP>
CUBEJS_DB_PORT=<PORT>
CUBEJS_DB_NAME=<DB>
CUBEJS_DB_USER=<USER>
CUBEJS_DB_PASS=<PASS>
$ 
```

` % env $(cat ~/.env.redshift | xargs) yarn smoke:redshift`, wh

packages/cubejs-redshift-driver/src/RedshiftDriver.ts
magno32: Same as the above comment, if the task/container/lambda/etc are assigned a role with permissions, the underlying SDK will be able to make the request.

These keys are helpful in development, where the dev machine probably does not have an `iam_role` with permissions.

packages/cubejs-redshift-driver/src/RedshiftDriver.ts
```diff
@@ -122,24 +150,27 @@ export class RedshiftDriver extends PostgresDriver<RedshiftDriverConfiguration>
         }
       });
 
+      const baseQuery = `UNLOAD ('SELECT * FROM ${table}') TO '${bucketType}://${bucketName}/${exportPathName}/'`;
+      const credentialQuery = unloadArn ?
```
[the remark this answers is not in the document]
magno32: I didnt particularly like this either, but try to use `const` whenever I can.  Suggestions?

packages/cubejs-redshift-driver/src/RedshiftDriver.ts
```diff
@@ -122,24 +150,27 @@ export class RedshiftDriver extends PostgresDriver<RedshiftDriverConfiguration>
...
+
       if (unloadTotalRows === 0) {
         return {
           csvFile: [],
         };
       }
 
       const client = new S3({
-        credentials: {
+        credentials: (keyId && secretKey) ? {
```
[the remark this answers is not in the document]
magno32: Another note,  this behavior can be tested by launching an EC2 instance with an attached `iam_role` that has access to a single S3 bucket.

You can install the AWS CLI, and without configuration (setting keys/secrets) you should only be able to access that assigned bucket.

packages/cubejs-redshift-driver/src/RedshiftDriver.ts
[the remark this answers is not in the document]
cristipp: `const` is fine, this is just a formatting nit for `? ...` and `: ...` being in 2 separate lines for readability.

packages/cubejs-redshift-driver/src/RedshiftDriver.ts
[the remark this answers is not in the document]
cristipp: Cool! Thanks for the clarification. Can you please add the comment(s) content as documentation for `RedshiftDriverExportArnAWS.unloadArn`. For example, but feel free to provide a better blurb :) 

```
 interface RedshiftDriverExportArnAWS extends RedshiftDriverExportRequiredAWS {
   // ARN used to access S3 data from e.g. EC2 instances, instead of explicit key/secret credentials.
   unloadArn?: string,
 }
```

packages/cubejs-redshift-driver/src/RedshiftDriver.ts
```diff
@@ -5,14 +5,24 @@ import crypto from 'crypto';
...
+interface RedshiftDriverExportArnAWS extends RedshiftDriverExportRequiredAWS{
+  unloadArn?: string,
+}
+
+interface RedshiftDriverExportKeySecretAWS extends RedshiftDriverExportRequiredAWS{
+  keyId?: string,
+  secretKey?: string,
+}
+
+interface RedshiftDriverExportAWS extends RedshiftDriverExportArnAWS, RedshiftDriverExportKeySecretAWS {
```
cristipp, line 23: Can we have *both* key/secret and ARN? Which one takes precedence?

If they are mutually exclusive, then can you please use discriminated unions (and no optional fields) along the lines of:

```
interface RedshiftDriverExportArnAWS extends RedshiftDriverExportRequiredAWS {
  type: 'arn',
  unloadArn: string,
}

interface RedshiftDriverExportKeySecretAWS extends RedshiftDriverExportRequiredAWS {
  keyId: string,
  secretKey: string,
}

type RedshiftDriverExportAWS = RedshiftDriverExportArnAWS | RedshiftDriverExportKeySecretAWS
```

packages/cubejs-testing/test/smoke-redshift.test.ts
```diff
@@ -28,7 +33,30 @@ describe('redshift', () => {
...
+};
+
+// Base environment vars that are required if S3 export is to be enabled.
+const baseS3Env: { [ key: string ]: string } = {
+  CUBEJS_DB_EXPORT_BUCKET_TYPE: 's3',
+  CUBEJS_DB_EXPORT_BUCKET: 'foo-bucket',
+  CUBEJS_DB_EXPORT_BUCKET_AWS_REGION: 'bar-region'
+};
+// No export/unload configured
+describe('redshift', reusableSuite());
```
cristipp, line 51: Perhaps I'm missing something obvious, but how can one override the actual values of the env variables when running the tests?

magno32: They can't.  I didn't see that you were actually testing the export/unload process, so I did not think the tests needed to do this (unless I missed something). The specific issue I had originally, was the connection process failing, simply because key/secret were not provided in the config, so these tests pass by virtue of the proper configuration being present.

Configuring the test in this manner, runs the 3 separate cases of configuration.  Per your note above, there are 4 cases (when the ARN and Keys are provided), but I will need to add those conditions in.

packages/cubejs-redshift-driver/src/RedshiftDriver.ts
magno32: I spent quite a bit figuring out how to do this... and failed!  Thanks for the pointer, would definitely make it easier to read.  Should be a similar option for when both are present as well.

magno32: As for your question, it turns out to be a two parter.

- For the `unload` operation in redshift, this proposed solution _prefers_ the `unloadArn` if both are present.
- It is my assumption, that upstream in the codebase, there is a configuration block for `S3Client` that loads the key/secret into its configuration.  If they are `undefined`, the underlying `S3Client` SDK will use the `iam_role` that is assigned to whatever is running it.  If they are **both** defined it will use the provided key/secret pair.  I am uncertain as to what will happen if only one of the API Key parameters is present.

cristipp:
LGTM

PR #4539, opened 2022-05-16
feat(extensions): Adds minio s3 region env to cubestore cloud storage

olejeglejeg:
**Check List**

Currently, we migrating to minio usage for cubestore pre-aggregations storage on s3 and we got a problem `X-Amz-Credential parameter is malformed`. The reason of this error is that our s3 vendor has API similar to AWS S3 with `AWS Signature Version 4`, where the `X-Amz-Credential` request header must be in format `AWS4-HMAC-SHA256.Credential=key/date/region/s3/aws4_request`. But cubestore generates header in format  `AWS4-HMAC-SHA256.Credential=key/date//s3/aws4_request`

Slack [thread](https://cube-js.slack.com/archives/CC0403RRR/p1652691255839329)

Added passing of new env that is responsible for s3 region for minio usage in cubestore cloud storage case.
New env name: `CUBESTORE_MINIO_REGION`. 
The behaviour is simple, for minio case the region is either taken from env or if not present it sets to empty string as it was previously


alikhil (approved):

waralex (approved):
Looks good.
Also I have tested  this PR with local minIO - all works with and without region in minIO server configuration.
Thanks a lot!

PR #4541, opened 2022-05-16
Refactor Rollup config, fix warnings, and support more file extensions

reify-thomas-smith:
**Check List**
- [x] Tests has been run in packages where changes made if available (n/a)
- [x] Linter has been run for changed code (I'm not sure if the linter actually check the rollup config, but it doesn't produce any errors)
- [x] Tests for the changes have been added if not covered yet (no functional changes)
- [x] Docs have been added / updated if required (no functional changes)

When trying to convert `cubejs-client-react` to TypeScript (#4475), I ran into a major problem — non-`.js` file extensions wouldn't resolve, but TypeScript will complain loudly if you explicitly import a `.ts` file. This PR fixes that problem, and it's no longer necessary to specify the file extension to import JSX, TS, or TSX files.

In the course of fixing the above, I ran into a lot of warnings, which I've also fixed. The build no longer produces warnings.

Some notes on specific changes:

- I added `plugin-node-resolve` to every configuration. This is what allows it to resolve non-JS extensions. Previously, the UMD configurations were set to resolve `.ts`, but it didn't matter because the other configurations were not.
- I removed `.json` from the list of extensions to resolve. It wasn't being used (and it wouldn't have worked anyway), and it probably makes sense to specify the extension when importing a JSON file for clarity's sake. It can be added back in if we want to do it the other way.
- I added `plugin-commonjs` to every configuration, but restricted it to only look at `node_modules`. This is apparently a common idiom.
- I changed `plugin-babel`'s exclusion list. It was previously set to ignore `node_modules`, and also ignore `core-js` only for UMD builds. But `core-js` is in `node_modules`; the reason it wasn't excluded properly is that `node_modules/**` doesn't exclude `packages/…/node_modules`. I fixed the `node_modules` exclusion and removed the now-redundant `core-js` exclusion.
- Since that was the only difference between the various Babel plugin configs, I consolidated all of the common plugins, which should make `bundle()` a bit easier to read.
- The UMD and CJS builds had `exports` set to `auto`, but for files that have both named and default exports, this behaves the same as `named` but emits a warning. I changed these to `exports: 'named'`.
- Some configurations didn't use the passed `globals`, which caused warnings. All configurations now use `globals`.
- The `cubejs-client-react` configurations didn't pass a `globals` with React, which caused a warning. It now passes globals.
- Finally, I reformatted the exported list to use array spread rather than a bunch of concats.

This should all work just as before, but without the warnings; however, I literally learned Rollup last weekend, so probably someone more familiar with the old rollup config should take a look at this.

It looks like the client package tests run on the package sources, not on the bundles, so it's not clear that there are any relevant tests to run. This definitely should be sanity-checked in some way before merging.


reify-thomas-smith:
@vasilev-alex Sorry to ping, just wanted to highlight that this is holding up future work on TypeScript conversion. If there's anything I can do to facilitate the review process, please let me know.

vasilev-alex:
@reify-thomas-smith thanks! 

I've tried to build the `client-core` library and the new `.esm.js` build file looks like a commonjs 

Also, hope you could see my comment on migration to TypeScript https://github.com/cube-js/cube.js/pull/4475#issuecomment-1138201359

PR #4550, opened 2022-05-17
feat: logging cubesql queries errors

gandronchik:


rust/cubesql/cubesql/src/sql/statement.rs
```diff
@@ -647,6 +646,36 @@ impl<'a> Visitor<'a> for UdfWildcardArgReplacer {
     }
 }
 
+#[derive(Debug)]
+pub struct SensitiveDataSanitizer {}
```
ovr, line 650: We need at least one test for it.

rust/cubesql/cubesql/src/sql/statement.rs
```diff
@@ -657,6 +656,39 @@ impl<'a> Visitor<'a> for UdfWildcardArgReplacer {
...
+impl<'ast> Visitor<'ast> for SensitiveDataSanitizer {
+    fn visit_value(&mut self, val: &mut ast::Value) {
+        match val {
+            ast::Value::SingleQuotedString(str)
+            | ast::Value::DoubleQuotedString(str)
+            | ast::Value::NationalStringLiteral(str) => {
+                if vec!["f", "false", "t", "true"].contains(&str.as_str()) {
+                    return;
+                }
+                *str = "[REPLACED]".to_string();
```
MazterQyou, line 685: Minor: `[REDACTED]` sounds like being used just for that everywhere, should we change the string?

PR #4560, opened 2022-05-19
feat(cubesql): support DISCARD [ALL | PLANS | SEQUENCES | TEMPORARY |…

gandronchik:


rust/cubesql/pg-srv/src/protocol.rs
```diff
@@ -274,6 +274,25 @@ impl Serialize for CommandComplete {
     }
 }
 
+impl PartialEq for CommandComplete {
```
ovr, line 277: Probably, it should work with `#[derive(PartialEq)]`

PR #4582, opened 2022-05-23
draft

buntarb:
**Check List**

[For example #12]

[Description goes here]

### DIFF
--- a/packages/cubejs-schema-compiler/src/adapter/PreAggregations.js
+++ b/packages/cubejs-schema-compiler/src/adapter/PreAggregations.js
@@ -362,81 +362,176 @@ export class PreAggregations {
   }
 
   static canUsePreAggregationForTransformedQueryFn(transformedQuery, refs) {
+    /**
+     * Returns an array of 2-elements arrays with the dimension and
+     * granularity strings sorted by the concatenated dimension +
+     * granularity key.
+     * @param {Array<{dimension: string, granularity: string}>} timeDimensions
+     * @returns {Array<Array<string>>}
+     */
+    const sortTimeDimensions = (timeDimensions) => (
+      timeDimensions &&
+      R.sortBy(
+        d => d.join('.'),
+        timeDimensions.map(
+          d => [
+            d.dimension,
+            d.granularity || 'day', // TODO granularity shouldn't be null?
+          ]
+        ),
+      ) || []
+    );
+
+    /**
+     * @type {Set<string}
+     */
     const filterDimensionsSingleValueEqual =
       transformedQuery.filterDimensionsSingleValueEqual instanceof Set
         ? transformedQuery.filterDimensionsSingleValueEqual
         : new Set(
-          Object.keys(transformedQuery.filterDimensionsSingleValueEqual || {})
+          Object.keys(
+            transformedQuery.filterDimensionsSingleValueEqual || {},
+          )
         );
-      
-    function sortTimeDimensions(timeDimensions) {
-      return timeDimensions && R.sortBy(
-        d => d.join('.'),
-        timeDimensions.map(d => [d.dimension, d.granularity || 'day']) // TODO granularity shouldn't be null?
-      ) || [];
-    }
-    const expandGranularity = (granularity) => transformedQuery.granularityHierarchies[granularity] || [granularity];
-    // TimeDimension :: [Dimension, Granularity]
-    // TimeDimension -> [TimeDimension]
-    function expandTimeDimension(timeDimension) {
-      const [dimension, granularity] = timeDimension;
-      const makeTimeDimension = newGranularity => [dimension, newGranularity];
-      return expandGranularity(granularity).map(makeTimeDimension);
-    }
-    // [[TimeDimension]]
-    const queryTimeDimensionsList = transformedQuery.sortedTimeDimensions.map(expandTimeDimension);
 
+    /**
+     * Determine whether pre-aggregation can be used or not.
+     * @param {*} references
+     * @returns {boolean}
+     */
+    const canUsePreAggregationNotAdditive = (references) => {
+      const sortedTimeDimensions =
+        references.sortedTimeDimensions ||
+        sortTimeDimensions(references.timeDimensions);
+
+      return (
+        transformedQuery.hasNoTimeDimensionsWithoutGranularity &&
+        !transformedQuery.hasCumulativeMeasures &&
+        (
+          references.allowNonStrictDateRangeMatch ||
+          R.equals(transformedQuery.sortedTimeDimensions, sortedTimeDimensions)
+        ) &&
+        (
+          transformedQuery.isAdditive ||
+          R.equals(transformedQuery.timeDimensions, sortedTimeDimensions)
+        ) &&
+        (
+          references.dimensions.length === filterDimensionsSingleValueEqual.size &&
+          R.all(d => filterDimensionsSingleValueEqual.has(d), references.dimensions)
+        ) &&
+        (
+          R.all(m => references.measures.indexOf(m) !== -1, transformedQuery.measures) ||
+          R.all(m => references.measures.indexOf(m) !== -1, transformedQuery.leafMeasures)
+        )
+      );
+    };
+    
+    /**
+     * Wrap granularity string into an array.
+     * @param {string} granularity
+     * @returns {Array<string>}
+     */
+    const expandGranularity = (granularity) => (
+      transformedQuery.granularityHierarchies[granularity] ||
+        [granularity]
+    );
+
+    /**
+     * Determine whether time dimensions match to the window
+     * granularity or not.
+     * @param {*} references
+     * @returns {boolean}
+     */
     const windowGranularityMatches = (references) => {
       if (!transformedQuery.windowGranularity) {
         return true;
       }
+      const sortedTimeDimensions =
+        references.sortedTimeDimensions ||
+        sortTimeDimensions(references.timeDimensions);
 
-      const sortedTimeDimensions = references.sortedTimeDimensions || sortTimeDimensions(references.timeDimensions);
-      return expandGranularity(transformedQuery.windowGranularity).map(windowGranularity => R.all(
-        td => td[1] === windowGranularity,
-        sortedTimeDimensions,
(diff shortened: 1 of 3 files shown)


packages/cubejs-schema-compiler/src/adapter/PreAggregations.js
```diff
@@ -362,82 +363,175 @@ export class PreAggregations {
...
   }
 
+  /**
+   * Returns function to determine whether pre-aggregation can be used or not
+   * for specified query, or its value for `refs` if specified.
+   * @param {Object} transformedQuery transformed query
+   * @param {Object?} refs pre-aggs reference
+   * @returns {function(preagg: Object): boolean}
+   */
   static canUsePreAggregationForTransformedQueryFn(transformedQuery, refs) {
```
cristipp, line 373: A bit unclear what a "transformedQuery" is. Could this function be called `canUsePreAggregationFn`? I also see above `PreAggregations.transformQueryToCanUseForm(query)`. Sounds like `canUseForm` is a specific query format. Would be awesome if it were explicitly documented in a docstring. For example:

```
/**
 * Returns function to determine whether pre-aggregation can be used or not
 * for specified query, or its value for `refs` if specified.
 * @param {Object} canUseQuery query in canUseForm, i.e. <has property A and B>
 * @param {Object?} refs pre-aggs reference
 * @returns {function(preagg: Object): boolean}
 */
static canUsePreAggregationFn(canUseQuery, refs) {
    ...
}
```

packages/cubejs-schema-compiler/src/adapter/PreAggregations.js
```diff
@@ -362,82 +363,175 @@ export class PreAggregations {
     }
   }
 
+  /**
+   * Returns function to determine whether pre-aggregation can be used or not
```
cristipp, line 367: Does this function need to return a predicate function, of could we simply call it directly, as in:

```
    static canUsePreAggregation(canUseQuery, refs, preagg): boolean {
        ...
    }
```

Alternatively, consider extracting a mini-class, especially since the code defines a ton of closures. Something along the lines of:

```
    class CanUsePreaggregation {
        constructor(private canUseQuery, private refs) {}

        canUse(preaggr) {
            ...
        }

        sortTimeDimensions(timeDimensions) {
           ...
        }
       
        <etc>
    }
```

packages/cubejs-schema-compiler/src/adapter/PreAggregations.js
```diff
@@ -362,82 +363,175 @@ export class PreAggregations {
...
+    /**
+     * Returns an array of 2-elements arrays with the dimension and granularity
+     * sorted by the concatenated dimension + granularity key.
+     * @param {Array<{dimension: string, granularity: string}>} timeDimensions
+     * @returns {Array<Array<string>>}
+     */
+    const sortTimeDimensions = (timeDimensions) => (
+      timeDimensions &&
+      R.sortBy(
+        d => d.join('.'),
```
cristipp, line 383: There are a number of codecov warnings along the lines of `Added line #L383 was not covered by tests`. These can be potential sources of bugs, especially since this is vanilla JS and there is little typing guardrails. Consider adding some tests to increase test coverage.

packages/cubejs-schema-compiler/src/adapter/PreAggregations.js
```diff
@@ -362,82 +363,175 @@ export class PreAggregations {
...
         ? transformedQuery.filterDimensionsSingleValueEqual
         : new Set(
-          Object.keys(transformedQuery.filterDimensionsSingleValueEqual || {})
+          Object.keys(
+            transformedQuery.filterDimensionsSingleValueEqual || {},
+          )
         );
+
+    /**
+     * Determine whether pre-aggregation can be used or not.
```
cristipp, line 406: This comment could be better by stating the criteria that make a pre-aggregation usable or not. That could help establishing intent, and identifying bugs if the code diverges from the stated intent.

packages/cubejs-schema-compiler/src/adapter/PreAggregations.js
```diff
@@ -362,82 +363,175 @@ export class PreAggregations {
...
+        R.all(m => references.measures.indexOf(m) !== -1, transformedQuery.leafMeasures)
+      ));
+    };
+    
+    /**
+     * Wrap granularity string into an array.
+     * @param {string} granularity
+     * @returns {Array<string>}
+     */
+    const expandGranularity = (granularity) => (
```
cristipp, line 442: A unit test would be great, my instinct is 'what does it mean to "wrap granularity into an array"?'. If this were a class, unit testing would be very simple :)

packages/cubejs-schema-compiler/src/adapter/PreAggregations.js
```diff
@@ -362,82 +363,175 @@ export class PreAggregations {
...
+      transformedQuery.granularityHierarchies[granularity] ||
+        [granularity]
+    );
 
+    /**
+     * Determine whether time dimensions match to the window granularity or not.
```
cristipp, line 448: What is the intended criteria?

packages/cubejs-schema-compiler/src/adapter/PreAggregations.js
```diff
@@ -362,82 +363,175 @@ export class PreAggregations {
...
+     */
     const windowGranularityMatches = (references) => {
       if (!transformedQuery.windowGranularity) {
         return true;
       }
+      const sortedTimeDimensions =
+        references.sortedTimeDimensions ||
+        sortTimeDimensions(references.timeDimensions);
+
+      return expandGranularity(transformedQuery.windowGranularity)
```
cristipp, line 460: An example (unit test) would help comprehension!

packages/cubejs-schema-compiler/src/adapter/PreAggregations.js
```diff
@@ -362,82 +363,175 @@ export class PreAggregations {
...
+        .length > 0;
+    };
 
-      const sortedTimeDimensions = references.sortedTimeDimensions || sortTimeDimensions(references.timeDimensions);
-      return expandGranularity(transformedQuery.windowGranularity).map(windowGranularity => R.all(
-        td => td[1] === windowGranularity,
-        sortedTimeDimensions,
-      )).filter(x => !!x).length > 0;
+    /**
+     * Returns an array of 2-element arrays with dimension and granularity.
```
cristipp, line 472: Example (unit test) pls? 🙏

packages/cubejs-schema-compiler/src/adapter/PreAggregations.js
```diff
@@ -452,13 +546,24 @@ export class PreAggregations {
...
 
   // eslint-disable-next-line no-unused-vars
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   getCubeLattice(cube, preAggregationName, preAggregation) {
     throw new UserError('Auto rollups supported only in Enterprise version');
   }
 
+  /**
+   * Returns pre-agg which determined as applicable for the query (the first one
+   * from the list of potencially applicable pre-aggs). The order of the incoming
```
cristipp, line 556: Nit: s/potencially/potentially/

packages/cubejs-schema-compiler/src/adapter/PreAggregations.js
```diff
@@ -452,13 +546,24 @@ export class PreAggregations {
...
+   * list of potencially applicable pre-aggs is equal to the order of these pre-
```
cristipp, line 557: Nit: "The order of the potentially applicable pre-aggs is the same as the order in which these pre-aggs appear in the schema file"

Generally speaking, I'm trying to help with the English formulation here (heh, not native Engslish speaker myself either). If this feels too nit-picky, I'm happy to stop :)

packages/cubejs-schema-compiler/src/adapter/PreAggregations.js
```diff
@@ -492,19 +598,35 @@ export class PreAggregations {
     return [];
   }
 
+  /**
+   * Returns an array of potencially applicable for the query preaggs in the
+   * same order they have in the cube schema js-file.
```
cristipp, line 603: Nit: "same order they appear in the schema file"

cristipp:
Looks great! Dropped in a few comments/notes. Hope it helps!

packages/cubejs-schema-compiler/src/adapter/PreAggregations.js
buntarb: By the way, `transformedQuery` is a transformed for the processing query and some parts of it differ from what we send from the client.

packages/cubejs-schema-compiler/src/adapter/PreAggregations.js
buntarb: Yeah, it should return a predicate function that closures the initial query. The predicates will be executed with every pre-aggregation. As far as I know, a simple call is used only for unit testing. The root cause of this design - is performance.

packages/cubejs-schema-compiler/src/adapter/PreAggregations.js
buntarb: Here we have unit tests and a bunch of integration tests, which, unfortunately, are not analyzed by the coverage engine. Generally speaking, this area is tested very well. We can improve coverage, for sure, but in the scope of some other tasks))

packages/cubejs-schema-compiler/src/adapter/PreAggregations.js
buntarb: Well, this is a good point. To be honest, I don't have a better comment for this function right now because I don't know all criteria we have. I need to invest a lot of time in the research to understand it. And as always, I don't have it ))

packages/cubejs-schema-compiler/src/adapter/PreAggregations.js
buntarb: It is impossible to unit test this function certainly because it's closed in the scope inside the parent and does not export from the file. But we have a lot of unit tests for the parent function, where this one was also tested. In general, I agreed that this part looks complicated, but refactoring it is very risky.

PR #4723, opened 2022-06-07
Temporarily support both `order` variants + better code!

omid:
How should we have a better Rust code for the clients?

I would like to have such PR, but it will avoid us to run the openapi generator again!
And as you may know, openapi generators cannot generate a good code, most of the time.


rust/cubesql/cubeclient/src/models/v1_load_request_query.rs
```diff
@@ -7,20 +9,21 @@ pub struct V1LoadRequestQuery {
     #[serde(rename = "segments", skip_serializing_if = "Option::is_none")]
     pub segments: Option<Vec<String>>,
     #[serde(rename = "timeDimensions", skip_serializing_if = "Option::is_none")]
-    pub time_dimensions: Option<Vec<crate::models::V1LoadRequestQueryTimeDimension>>,
+    pub time_dimensions: Option<Vec<super::V1LoadRequestQueryTimeDimension>>,
     #[serde(rename = "order", skip_serializing_if = "Option::is_none")]
-    pub order: Option<Vec<Vec<String>>>,
+    pub order: Option<Value>,
```
omid, line 14: This is the main change here, to temporarily support both variants of `order`.
All other changes are to have a better idiomatic Rust code!

ovr:
Hello @omid ,

Thank you for submitting the PR!

> And as you may know, openapi generators cannot generate a good code, most of the time.

It's true. But the biggest problem is that merging anything which improves code quality with hands will lead to blocking of regeneration.

I will think about what we can do. Maybe adding our own template for the open spec generator can help.

Thanks

Issue #4570, opened 2022-05-20
Unsopprted react 18 strictmode 


julio-burgos:
**Describe the bug**
Hooks that use useEffect such as useCubeQuery, does not update correctly isMounted.current when using strictmode.

**To Reproduce**
1. Use latest version of "@cubejs-client/react"
2. Wrap the app with  React.StrictMode

**Expected behavior**

**Minimally reproducible Cube Schema**
In case your bug report is data modelling related please put your minimally reproducible Cube Schema here.
You can use selects without tables in order to achieve that as follows.

```javascript
   const { resultSet, error, isLoading } = useCubeQuery({
    measures: ["Table.count"],
    total: true,
  });
```
isLoading does not became false   allthough the query seems to be executed

**Version:**
[0.29.54, 0.30.4]

**Additional context**
The docs for the "@cubejs-client/react" also dont show up on the docs

PR #4740, opened 2022-06-09
fix(client-react): update useIsMounted hook to be compatible with React 18 StrictMode

alexbarrett:
**Check List**

#4570

Starting in React 18, StrictMode intentionally double-invokes effects (`mount` -> `unmount` -> `mount`) for newly mounted components during development. Because the `isMounted` flag inside the `useIsMounted` hook is only set to true during initialisation, the flag gets set to false immediately during the 'unmount' phase and never gets restored to true. Setting the flag to true in the 'mount' phase restores the correct behaviour.

See: https://github.com/reactwg/react-18/discussions/19


rpaik:
@vasilev-alex can you help with the review?

PR #4820, opened 2022-06-28
fix(athena-driver): column order swapping for Athena

zinychz:
**Check List**

[For example #12]

[Description goes here]

### DIFF
--- a/packages/cubejs-athena-driver/src/AthenaDriver.ts
+++ b/packages/cubejs-athena-driver/src/AthenaDriver.ts
@@ -64,16 +64,14 @@ export class AthenaDriver extends BaseDriver implements DriverInterface {
   }
 
   public async tableColumnTypes(table: string) {
-    const [schema, name] = table.split('.');
+    const [schema, name] = table.split('.', 2);
 
     const columns = (await this.query(
       `SELECT columns.column_name as ${this.quoteIdentifier('column_name')},
-             columns.table_name as ${this.quoteIdentifier('table_name')},
-             columns.table_schema as ${this.quoteIdentifier('table_schema')},
-             columns.data_type  as ${this.quoteIdentifier('data_type')}
+              columns.data_type  as ${this.quoteIdentifier('data_type')}
       FROM information_schema.columns
       WHERE table_name = ${this.param(0)} AND table_schema = ${this.param(1)}
-      ORDER BY 1, 2, 3, 4`,
+      ORDER BY ordinal_position`,
       [name, schema]
       // eslint-disable-next-line camelcase
     )) as {column_name: string, data_type: string}[];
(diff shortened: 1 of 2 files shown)


packages/cubejs-athena-driver/src/AthenaDriver.ts
```diff
@@ -63,6 +63,24 @@ export class AthenaDriver extends BaseDriver implements DriverInterface {
...
+    const [schema, name] = table.split('.');
+
+    const columns = (await this.query(
+      `SELECT columns.column_name as ${this.quoteIdentifier('column_name')},
+             columns.table_name as ${this.quoteIdentifier('table_name')},
+             columns.table_schema as ${this.quoteIdentifier('table_schema')},
+             columns.data_type  as ${this.quoteIdentifier('data_type')}
+      FROM information_schema.columns
+      WHERE table_name = ${this.param(0)} AND table_schema = ${this.param(1)}
+      ORDER BY 1, 2, 3, 4`,
```
cristipp, line 76: Nit: `ORDER BY 1` suffices, the other columns are redundant.

packages/cubejs-athena-driver/src/AthenaDriver.ts
```diff
@@ -63,6 +63,24 @@ export class AthenaDriver extends BaseDriver implements DriverInterface {
     this.athena = new Athena(this.config);
   }
 
+  public async tableColumnTypes(table: string) {
+    const [schema, name] = table.split('.');
```
cristipp, line 67: Nit: `table.split('.', 2)` so we are sure there are no surprises.

packages/cubejs-athena-driver/src/AthenaDriver.ts
```diff
@@ -63,6 +63,24 @@ export class AthenaDriver extends BaseDriver implements DriverInterface {
     this.athena = new Athena(this.config);
   }
 
+  public async tableColumnTypes(table: string) {
```
cristipp, line 66: Not 100% why/where the order of columns returned by the source DB matters. Shouldn't Cube column order be defined by the order of measures/dimensions/timeDimensions? Can you please add a small jsdoc comment line to explain /why/ the `ORDER BY columns.column_name` clause is necessary?

cristipp:
If I understand correctly, this PR stabilizes the column order. The fix is done only for AthenaDriver. Are other drivers potentially subject to the same issue?

packages/cubejs-athena-driver/src/AthenaDriver.ts
zinychz: You are right, agree

cristipp:
Generally speaking, a bug fix should have a test associated with it, which should fail without the fix and pass with the fix.

PR #4863, opened 2022-07-07
fix(cubejs): Default value for options.queueOptions

cristipp:
**Check List**

N/A

Adds a default value for `options.queueOptions`. Prevents "error": "TypeError: this.options.queueOptions is not a function".

### DIFF
--- a/packages/cubejs-query-orchestrator/src/orchestrator/PreAggregations.ts
+++ b/packages/cubejs-query-orchestrator/src/orchestrator/PreAggregations.ts
@@ -1634,7 +1634,7 @@ export class PreAggregations {
           redisPool: this.options.redisPool,
           // Centralized continueWaitTimeout that can be overridden in queueOptions
           continueWaitTimeout: this.options.continueWaitTimeout,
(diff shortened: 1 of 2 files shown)


packages/cubejs-query-orchestrator/src/orchestrator/PreAggregations.ts
```diff
@@ -1634,7 +1634,7 @@ export class PreAggregations {
           redisPool: this.options.redisPool,
           // Centralized continueWaitTimeout that can be overridden in queueOptions
           continueWaitTimeout: this.options.continueWaitTimeout,
-          ...(await this.options.queueOptions(dataSource)),
+          ...(this.options.queueOptions ? await this.options.queueOptions(dataSource) : {}),
```
buntarb, line 1637: @cristipp after my changes `this.options.queueOptions` does not require defaults because it is always set and a function. You can find its configuration [here](https://github.com/cube-js/cube.js/blob/master/packages/cubejs-server-core/src/core/OptsHandler.ts#L523).

cristipp:
Already fixed!

PR #4908, opened 2022-07-14
fix(cubesql): Binary operations with dates and intervals

gandronchik:
- fix binary operations with negative hour/minute/second intervals
- handle February
- handle leap year


ovr (approved):
LGTM

Issue #4889, opened 2022-07-11
Support multiple CubeProviders


reify-thomas-smith:
**Is your feature request related to a problem? Please describe.**

I have a React application that needs to query two different cube services. However, the `<CubeProvider>` component only supports specifying a single Cube service, and if `<CubeProvider>`s are nested, then only the innermost client object will be available.

**Describe the solution you'd like**

Add an optional `name` prop to `<CubeProvider>`, with a default value of “default”. The internal CubeContext, rather than holding a single client object, would hold an object mapping names to clients.

Hooks that presently accept an optional client argument would also accept a string value for that argument. If a string is provided, the hook would use the client with that name from the context, if it exists, or throw an error otherwise. If no value is passed, it would use the “default” client (matching current behavior).

**Describe alternatives you've considered**

As a workaround, I've written my own `<CubeProvider>` implementation that allows multiple clients. To use this with the existing hooks requires manually extracting the client instance from the context. A built-in implementation would be better.

As an alternative to the existing hooks accepting strings, we could directly expose a hook that gets a client by name, then expect the user to call this hook and pass the client object to the existing hook. This would arguably be a smaller change.

As an alternative to using the string “default” as a default name, we could treat an un-named client specially. This would arguably be cleaner in a way, but it would also be more complicated, and I don't see the need.

I am willing to work on this.

paveltiunov:
Hey @reify-thomas-smith ! Could you please elaborate on your use case? Why do you need to query multiple Cube services?

paveltiunov:
@reify-thomas-smith That makes sense! Do you have any proposal on how API should look like in this case?

reify-thomas-smith:
Yeah, just add an optional `name` prop to `CubeProvider`, and have `useCubeQuery` and such accept a string for the `client` option. I have a PR I've been sitting on because I haven't had time to look at updating docs (or testing it at all), but it should clarify the proposal: #4947.

PR #4947, opened 2022-07-19
React client multiple cubes

reify-thomas-smith:
**Check List**
- [x] Tests has been run in packages where changes made if available (n/a)

#4889

### DIFF
--- a/packages/cubejs-client-react/src/CubeProvider.jsx
+++ b/packages/cubejs-client-react/src/CubeProvider.jsx
@@ -1,6 +1,16 @@
-import React from 'react';
+import React, { useContext, useMemo } from 'react';
 import CubeContext from './CubeContext';
 
-export default function CubeProvider({ cubejsApi, children }) {
-  return <CubeContext.Provider value={{ cubejsApi }}>{children}</CubeContext.Provider>;
+export default function CubeProvider({
+  name,
+  cubejsApi,
+  children,
+}) {
+  const oldContext = useContext(CubeContext);
+  const newContext = useMemo(() => ({
+    ...oldContext,
+    [name ?? 'default']: cubejsApi,
+  }), []);
+
+  return <CubeContext.Provider value={newContext}>{children}</CubeContext.Provider>;
 }
--- a/packages/cubejs-client-react/src/hooks/cube-api.js
+++ b/packages/cubejs-client-react/src/hooks/cube-api.js
@@ -1,0 +1,7 @@
+import { useContext } from 'react';
+import CubeContext from '../CubeContext';
+
+export function useCubeApi(explicitApi) {
+  const context = useContext(CubeContext);
+  return explicitApi || context?.cubejsApi;
+}
--- a/packages/cubejs-client-react/src/hooks/cube-api.js
+++ b/packages/cubejs-client-react/src/hooks/cube-api.js
@@ -1,7 +1,11 @@
 import { useContext } from 'react';
 import CubeContext from '../CubeContext';
 
-export function useCubeApi(explicitApi) {
+export function useCubeApi(explicitApiOrName) {
   const context = useContext(CubeContext);
-  return explicitApi || context?.cubejsApi;
+  if (typeof explicitApiOrName === 'string') {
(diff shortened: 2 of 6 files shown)


packages/cubejs-client-react/src/hooks/cube-query.js
```diff
@@ -12,15 +12,14 @@ export function useCubeQuery(query, options = {}) {
   const [resultSet, setResultSet] = useState(null);
   const [progress, setProgress] = useState(null);
   const [error, setError] = useState(null);
-  const context = useContext(CubeContext);
+  const cubejsApi = useCubeApi(options.cubejsApi);
```
vasilev-alex, line 15: This change makes it throw an error on line 68

packages/cubejs-client-react/src/CubeProvider.jsx
```diff
@@ -1,6 +1,16 @@
...
+export default function CubeProvider({
+  name,
+  cubejsApi,
+  children,
+}) {
+  const oldContext = useContext(CubeContext);
+  const newContext = useMemo(() => ({
+    ...oldContext,
+    [name ?? 'default']: cubejsApi,
+  }), []);
```
vasilev-alex, line 13: It won't work without `[oldContext]`

Also, I'm not sure about this approach as it seems a bit unnatural having multiple `CubeProvider`s override the context value.

I would rather make the `value` an object, e.g.

```jsx
<CubeProvider = {{
  myApi: cubeApi1,
  myOtherApi: cubeApi2
  // ...
```

packages/cubejs-client-react/src/hooks/cube-api.js
```diff
@@ -0,0 +1,11 @@
+import { useContext } from 'react';
+import CubeContext from '../CubeContext';
+
+export function useCubeApi(explicitApiOrName) {
```
vasilev-alex, line 4: This should only accept `name` as a param and default to the `default` api

vasilev-alex (changes requested):

packages/cubejs-client-react/src/hooks/cube-api.js
reify-thomas-smith: This is just a helper function to avoid code duplication, not new API. But I can change it if you like; it will just mean a bit more complexity on the caller side.

packages/cubejs-client-react/src/CubeProvider.jsx
reify-thomas-smith: I'm not sure I understand the example.

As currently implemented, you would write:

```tsx
<CubeProvider name="foo" cubeJsApi={fooApi}>
  <CubeProvider name="bar" cubeJsApi={barApi}>
    <CubeProvider cubeJsApi={defaultApi}> {/* name is "default" */}
      {…}
    </CubeProvider>
  </CubeProvider>
</CubeProvider>
```

Internally, the APIs are kept in a shared context, but they wouldn't override each other as long as they have different names. The fact that all of the APIs are stored in a single object in the context is an implementation detail.

We could do the same thing with a different API:

```tsx
<CubeProvider cubeJsApi={{
  foo: fooApi,
  bar: barApi,
  default: defaultApi,
}}>
  {…}
</CubeProvider>
```

I like the `name` prop myself, but this API would be fine. The behavior should be identical and any performance differences negligible.

Also, `...oldContext` should work fine; it's object spread, not array spread, so it will work even if `oldContext` is null. I might be misunderstanding what you're saying here.

vasilev-alex: I've updated the example where I left out the `cubejsApi=` part.

While the approach with `name` works just fine, it's more error-prone (e.g. duplicate, missing names). Also, it's feels a bit off because technically you can access a `cubeApi` outside of its own provider (because we have multiple providers and a single context)

> Also, ...oldContext should work fine; it's object spread, not array spread, so it will work even if oldContext is null. I might be misunderstanding what you're saying here.

I meant you needed to include it as the useEffect's dependency otherwise, it won't update

PR #4949, opened 2022-07-20
docs(sql): How to change security context for SQL API

ovr:
Hello!

Thanks

### DIFF
--- a/docs/content/API-Reference/SQL-API.mdx
+++ b/docs/content/API-Reference/SQL-API.mdx
@@ -6,8 +6,8 @@ menuOrder: 3
 ---
 
 <InfoBox>
-  <b>Connecting your Business Intelligence Tools to Cube</b> workshop on July 27, 2022.<br/> 
-  Building on our <a href="https://cube.dev/events/sql-api">SQL API workshop</a>, we'll continue our discussion on making Cube available to your favorite Business Intelligence applications.<br /> 
+  <b>Connecting your Business Intelligence Tools to Cube</b> workshop on July 27, 2022.<br/>
+  Building on our <a href="https://cube.dev/events/sql-api">SQL API workshop</a>, we'll continue our discussion on making Cube available to your favorite Business Intelligence applications.<br />
   Check out the agenda and resigter for the workshop today on the <a href="https://cube.dev/events/adv-sql-api/">event page</a> 👈
 </InfoBox>
 
@@ -390,6 +390,29 @@ module.exports = {
 };
 ```
 
+## Security Context (Row-Level Security)
+
+Cube's SQL API can also use the Security Context for
+[Dynamic Schema Creation][ref-dynamic-schemas] or [`queryRewrite`][ref-config-queryrewrite] property in your [`cube.js`
+configuration file][ref-config-js].
+
+By default, the SQL API uses the current user's Security Context, but this behaviour can be modified so that certain users are allowed to switch. To do this, we must first define which user is allowed to change Security Context:
+
+First, you need to define what user is allowed to change security context:
+
+```
+CUBEJS_SQL_SUPER_USER=admin
+```
+
+If it's not enough for your case, you define your logic for check with `canSwitchSqlUser` property in your [`cube.js`
+configuration file][ref-config-js].
+
+You can change security context for specific query via virtual filter on:
+
+```sql
+SELECT * FROM Orders WHERE __user = 'anotheruser';
+```
+
 ## Limitations
 
 ### <--{"id" : "Limitations"}--> Projection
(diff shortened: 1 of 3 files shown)


docs/content/API-Reference/SQL-API.mdx
```diff
@@ -369,6 +369,29 @@ module.exports = {
 };
 ```
 
+## Security Context (Row-Level Security)
+
+Cube.js provides security context functionality, which can be used in the
```
hassankhan, line 374:
```suggestion
Cube's SQL API can also use the Security Context for
```

docs/content/API-Reference/SQL-API.mdx
```diff
@@ -369,6 +369,29 @@ module.exports = {
...
+[Dynamic Schema Creation][ref-dynamic-schemas] or [`queryRewrite`][ref-config-queryrewrite] property in your [`cube.js`
+configuration file][ref-config-js].
+
+By default, SQL API uses current user's security context, but there is way how to change it.
```
hassankhan, line 378:
```suggestion
By default, the SQL API uses the current user's Security Context, but this behaviour can be modified so that certain users are allowed to switch. To do this, we must first define which user is allowed to change Security Context:
```

docs/content/API-Reference/SQL-API.mdx
```diff
@@ -369,6 +369,29 @@ module.exports = {
...
+
+First, you need to define what user is allowed to change security context:
+
+```
+CUBEJS_SQL_SUPERUSER=admin
```
hassankhan, line 383: We should add `CUBEJS_SQL_SUPERUSER` to https://cube.dev/docs/reference/environment-variables

docs/content/Configuration/Config.mdx
```diff
@@ -314,6 +315,31 @@ module.exports = {
 };
 ```
 
+### <--{"id" : "Options Reference"}--> canSwitchSqlUser
+
+Used in [SQL API][ref-sql-api], and can be an `async` function. Default
```
hassankhan, line 320: `[ref-sql-api]` isn't linked properly

docs/content/Configuration/Config.mdx
```diff
@@ -314,6 +315,31 @@ module.exports = {
...
+implementation depends on `CUBEJS_SQL_SUPERUSER` and return `true` when it's equal to session's user.
+
+Called on each change request from Cube SQL API.
+
+For example, you can use `canSwitchSqlUser` to define your custom logic.
```
hassankhan, line 325:
```suggestion
For example, you can use `canSwitchSqlUser` to define your custom logic:
```

docs/content/Configuration/Config.mdx
```diff
@@ -314,6 +315,31 @@ module.exports = {
...
+        return true;
+    }
+
+    if (current === 'service') {
+        return username !== 'admin';
+    }
+
+    return false;
+  },
+};
```
hassankhan, line 340: Can we use an `async` function here and return `Promise.resolve`d values instead?

hassankhan (changes requested):
Minor feedback, thanks for the write-up @ovr <3

docs/content/Configuration/Config.mdx
ovr: Yes, user is allowed to use async function here.

docs/content/Configuration/Config.mdx
ovr: Hm, as I can see, it's defined at the end of `Config.mdx`
image

docs/content/API-Reference/SQL-API.mdx
hassankhan: > I didn't find a section for it. Should I create a new one?

That's a great point, could you create a new 'SQL API' section? Thank you 😀

docs/content/Configuration/Config.mdx
hassankhan: Weird, when I tried it in the Netlify Deploy Preview the link didn't seem to work

docs/content/Configuration/Config.mdx
hassankhan: Sorry I meant can we change the example here to be a Promise-returning one instead?

docs/content/Configuration/Config.mdx
hassankhan: Just checked, the correct URL should be `/backend/sql` (instead of `/sql-api`)

ovr: ✅

docs/content/API-Reference/SQL-API.mdx
ovr: ✅

docs/content/Configuration/Config.mdx
ovr: Move to async fn.

docs/content/API-Reference/SQL-API.mdx
hassankhan: Amazing, thank you

PR #4951, opened 2022-07-20
docs(config): document default concurrency settings for all drivers

hassankhan:
**Check List**


docs/content/Configuration/Overview.mdx
```diff
@@ -57,6 +57,24 @@ module.exports = {
...
+
+</InfoBox>
+
+For increased performance, Cube uses multiple concurrent connections to configured
+data sources.
+
+Internally, Cube uses two queues that manage database requests; one
+for API queries and one for pre-aggregations. The number of concurrent database
+connections for each queue can be set by the `CUBEJS_CONCURRENCY` environment variable.
+To limit the maximum number of concurrent connections to the database, the `CUBEJS_DB_MAX_POOL`
```
buntarb, line 75: Not sure do we need to highlight it, but not every DB have max pool option.

docs/content/Configuration/Overview.mdx
```diff
@@ -57,6 +57,24 @@ module.exports = {
...
+The following information is included as a reference.
+
+</InfoBox>
+
+For increased performance, Cube uses multiple concurrent connections to configured
+data sources.
+
+Internally, Cube uses two queues that manage database requests; one
+for API queries and one for pre-aggregations. The number of concurrent database
+connections for each queue can be set by the `CUBEJS_CONCURRENCY` environment variable.
```
buntarb, line 74: Generally speaking, CUBEJS_CONCURRENCY is a base concurrency configuration parameter that determines queues concurrencies, refresh scheduler concurrency and max pool size. All in one. For example, if it equals 5, queues concurrencies will be 5, refresh scheduler concurrency will be 5, and max pool will be (5 + 5) x 2 = 20. There is some math behind the scene, but it's our internal stuff. The main point is that you can play with this variable instead of manually configuring queues, scheduler, and driver. The default depends on the used driver (getDefaultConcurrency static method in every driver we have) if the variable is not specified.

docs/content/Configuration/Overview.mdx
hassankhan: Good point, I've updated the text

docs/content/Configuration/Overview.mdx
hassankhan: > There is some math behind the scene, but it's our internal stuff

That's the reason why I didn't go into too much detail here and left the information to be more general here. I've rewritten it now, hopefully it reads better now

buntarb (approved):

Issue #4900, opened 2022-07-12
[Firebolt] The Firebolt driver should start the Firebolt engine if it isn't running


mikulskibartosz:
When the Firebolt engine isn't running, Cube returns error 404 and cannot perform any queries. The driver should start the engine automatically when Cube performs a query.

**Describe the solution you'd like**
Cube Firebolt driver should start the Firebolt engine automatically before sending a query.

**Additional context**

Steps to reproduce the issue:

1. Configure a connection to Firebolt.
2. Send a query. Make sure that it works.
3. Stop the engine.
4. Send the same query again.

Actual behavior:

Cube logs error 404 and the Cube application fails.

Expected behavior:

Cube starts the configured Firebolt engine automatically and successfully runs the query.

PR #5005, opened 2022-07-27
feature (firebolt-driver) engine autostart

Jeseeq:
**Check List**

Resolves issue [#4900]


packages/cubejs-testing/src/REQUIRED_ENV_VARS.ts
```diff
@@ -19,7 +19,7 @@ export const REQUIRED_ENV_VARS: {[key: string]: string[]} = {
     'CUBEJS_DB_USER',
     'CUBEJS_DB_PASS',
     'CUBEJS_DB_NAME',
-    'CUBEJS_FIREBOLT_ENGINE_ENDPOINT'
+    'CUBEJS_FIREBOLT_ENGINE_NAME'
```
cristipp, line 22: Add `CUBEJS_FIREBOLT_ACCOUNT` here too?

Jeseeq: `CUBEJS_FIREBOLT_ACCOUNT` is not really required. Firebolt sdk resolves yours default account if account is not specified. In Firebolt your user may be a member of > 1 account, only then you need to specify account name.

cristipp: Fair enough. That is a good point!

For background, the naming here is not that fortunate, the goal of these variables is two fold:
1. Ensure that the auth is correctly set.
2. Copy auth vars into Docker environment, if using Docker-ized cubejs.
I will make a separate PR to automatize 2. based on env var prefixes, so it gets less confusing.

cristipp:
LGTM. Thanks for the contribution!

PR #5047, opened 2022-08-01
feat(cubejs-api-gateway): change ContinueWait response status code

rawpixel-vincent:
**Check List**

Change the http status code of the Continue Wait API response to 202.

This would allow setting a CDN in front of the cube client and add a rule to exclude the incomplete response from the cache.

AFAICT this is all that is needed, but I can't tell if there is infrastructure that relies on the status code 200.

I couldn't find a straightforward way to add a header to the response, adding a custom http header to Continue Wait api response (like X-Cube-Continue-Wait: true) could also resolve this feature.


rpaik:
@ovr can you help with the review?

rawpixel-vincent:
Hi @buntarb @rpaik any chance to get some feedback on this? I'm happy to help makes the changes needed to be able to identify a fulfilled query response apart from a pending one at a CDN level.

buntarb:
@rawpixel-vincent, thanks for your contribution. These changes look significant. But I'm afraid it can cause backward compatibility issues for many customers who don't use our client. Please add the `CUBEJS_CONTINUE_WAIT_ERROR_STATUS` environment variable with the `200` default and supported values `200` and `202`, and use it to determine the status in runtime.

PR #5081, opened 2022-08-09
feat(client-core): Add loadResponse and loadResponses to ResultSet

tchell:
The loadResponse(s) that are set on the ResultSet allow determining some
useful metadata on a query. These attributes should be included in the
typing so that users writing in Typescript can access their information
without running into errors from the missing attributes.

**Check List**


vasilev-alex:
hi @tchell! We didn't expose it intentionally as it's not a part of public API. Though you can still use the `serialize` method

tchell:
Understood. That is not super obvious in the documentation.

PR #5094, opened 2022-08-11
refactor(cubesql): logging sanitized query

gandronchik:
- refactor logging sanitized query
- add sanitized query to log message


ovr (approved):
LGTM

PR #5131, opened 2022-08-21
refactor(cubesql): Rewrite Postgres server/connection

ovr:


borodark (commented):
Any chance this patch would make Elixir Ecto working with Cube? I have issues with Ecto Repo failing to start because of expectations of Ecto from types returned by server.

PR #5132, opened 2022-08-22
Rename .d.ts files to .ts files in order to expose the types in the build

ojhaujjwal:
**Check List**

I imported `QueryFilter` from the api-gateway package but the typing didn't work. For example:
image

The type exists for CreateOptions, so TS is giving type errors but not for QueryFilter.

I looked in the imported package folder in my project and non of the .d.ts files from source were there.

image

Only the `enum.d.ts` was there because it's a `.ts` file in the source already.

This [StackOverflow](https://stackoverflow.com/a/56440335/2508053) answers why those types don't exist.

This should help the TS experience.


ojhaujjwal:
This is a very small change. 

Can someone please review or merge this if all's okay? Not a fan of using `any` types in typescript.

PR #5148, opened 2022-08-23
update(driver-materialize) compatibility with newer versions

joacoc:
**Check List**

There is no reference issue.

Newer versions of Materialize have a different way to query the schema.
This PR maintains backward compatibility with older versions too.

Added `semver` lib to calculate the versions.
Added `ramda` lib to keep the same mechanism to return the schema as BaseDriver in `tablesSchema`.


buntarb:
Hi, @joacoc! Thanks for your contribution. I have reviewed this PR, and it looks straightforward enough. But unfortunately, there were some significant changes (the `BaseDriver` class was split into an independent package) a while ago, and it contains some conflicts and stuff to fix. So, could you please rebase and fix it?

PR #5152, opened 2022-08-24
fix(databricks-jdbc): databricks multiple downloads of jar

frolovdev:
**Check List**

[For example #12]

[Description goes here]


packages/cubejs-backend-shared/src/env.ts
```diff
@@ -204,7 +204,7 @@ const variables: Record<string, (...args: any) => any> = {
   databrickToken: () => get('CUBEJS_DB_DATABRICKS_TOKEN')
     .asString(),
   databrickAcceptPolicy: () => get('CUBEJS_DB_DATABRICKS_ACCEPT_POLICY')
-    .asString(),
+    .asBoolStrict(),
```
frolovdev, line 207: 📓 looks like a typo

packages/cubejs-databricks-jdbc-driver/src/DatabricksDriver.ts
```diff
@@ -64,18 +64,13 @@ const DatabricksToGenericType: Record<string, string> = {
   'decimal(10,0)': 'bigint',
 };
 
-const jdbcDriverResolver: Promise<string> | null = null;
-
 async function resolveJDBCDriver(): Promise<string> {
-  if (jdbcDriverResolver) {
-    return jdbcDriverResolver;
-  }
```
frolovdev, line 72: 📓 dead code

packages/cubejs-databricks-jdbc-driver/src/installer.ts
```diff
@@ -16,33 +15,8 @@ function acceptedByEnv() {
   return acceptStatus;
 }
 
-async function cliAcceptVerify() {
```
frolovdev, line 19: 📓 now can only be configured with env variable

packages/cubejs-databricks-jdbc-driver/src/post-install.ts
```diff
@@ -2,11 +2,15 @@ import 'source-map-support/register';
...
 import { displayCLIError } from '@cubejs-backend/shared';
 
+import fs from 'fs';
+import path from 'path';
 import { downloadJDBCDriver } from './installer';
 
 (async () => {
   try {
-    await downloadJDBCDriver(true);
+    if (!fs.existsSync(path.join(__dirname, '..', 'download', 'SparkJDBC42.jar'))) {
```
frolovdev, line 11: 📓 don't reload on every yarn install

VinThin:
@frolovdev - we are trying to use this driver but simple not able to work, my js skills are super rusty this is my issue

packages/cubejs-databricks-jdbc-driver/src/post-install.ts
buntarb: This block could be called more than onece, which could be a cause for a second download. You should check not only the file existing, but also a promise of the download function. Isn't it?

frolovdev: Didn't get the point, in which case we can call it more than once? Two parallel yarn installs looks like a redundant case or do you mean smth else?

buntarb (approved):

PR #5168, opened 2022-08-26
chore: plain error messages

vasilev-alex:
Introduce `plainError` as an alternative to the error containing code frames


ovr (approved):
LGTM

Right now, we return formated errors from schema-compiler via `@babel/core-frame`. To do that, we uses large package which formats code-frames in the cube client. This PR allow to make this as optional.

PR #5205, opened 2022-08-31
fix: add logger on driver level

frolovdev:
**Check List**

[For example #12]

Added an option to use a driver context and logger in driver


packages/cubejs-server-core/src/core/server.ts
```diff
@@ -718,11 +719,11 @@ export class CubejsServerCore {
    * Returns driver instance by a given context
    */
   public async getDriver(
```
frolovdev, line 721: 📓 because it's a public method which can be called only on instantiated class it looks much better to exclude logger from a supertype

PR #5235, opened 2022-09-05
fix(cubesql): select column with the same name as table

gandronchik:
**example:**
`select table.column as column from (select 1 column, 2 table union all select 3 column, 4 table) table;`


ovr (approved):
LGTM

PR #5236, opened 2022-09-05
feat(cubesql): Support yearly granularity (ThoughtSpot)

ovr:
image

### DIFF
--- a/packages/cubejs-backend-native/Cargo.lock
+++ b/packages/cubejs-backend-native/Cargo.lock
@@ -79,7 +79,7 @@ checksum = "8da52d66c7071e2e3fa2a1e5c6d088fec47b593032b254f5e980de8ea54454d6"
 [[package]]
 name = "arrow"
 version = "11.1.0"
-source = "git+https://github.com/cube-js/arrow-rs.git?rev=bf19359d1a5cf6e2a96dec805a0bd4da1c30df5e#bf19359d1a5cf6e2a96dec805a0bd4da1c30df5e"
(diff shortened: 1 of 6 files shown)


rust/cubesql/cubesql/src/compile/snapshots/cubesql__compile__tests__test_thought_spot_yearly_plan.snap
```diff
@@ -0,0 +1,20 @@
+---
+source: cubesql/src/compile/mod.rs
+assertion_line: 6442
+expression: "format!(\"{:?}\", logical_plan)"
+---
+Projection: #CAST(CAST(datepart(Utf8("YEAR"),ta_1.order_date) * Int64(100) + Int64(1) * Int64(100) + Int64(1) AS Utf8) AS Date32) AS ca_1, CASE WHEN #SUM(ta_1.count) IS NOT NULL THEN #SUM(ta_1.count) ELSE Int64(0) END AS ca_2
```
ovr, line 6: This assertation is important, because without this fix SQL API can reassembly this query by splitting/pushing all expression for outer_aggregate, but it's doesn't work.

>Execution error: Arrow error: Cast error: Cannot cast string '20190101' to value of arrow::datatypes::types::Date32Type type

ovr: Looks like Postgresql supports different delimiters for casting:

```
select cast('2019|01|01' as date);
select cast('2019\01\01' as date);
select cast('2019/01/01' as date);
```

PR #5253, opened 2022-09-08
Handle async errors by passing the error to next middleware

ojhaujjwal:
**Check List**

If express route handlers/middleware return async errors/rejected promise, they aren't passed on to next automatically but it's passed on for sync errors/uncaught exceptions.

I believe this is changing as a part of express 5 but I think this is a worthwhile addition until express 5 final release is out and we migrate to express 5.


paveltiunov:
Hey @ojhaujjwal ! Thanks for contributing to it! Could you please elaborate on what problem you are trying to solve with this change?

PR #5254, opened 2022-09-08
feat(cubesql): Support interval multiplication

MazterQyou:
**Check List**

This PR adds support for interval multiplication (e.g. `-1 * 5 * INTERVAL '1 month'`). Such operations are extensively used by QuickSight. It also adds related tests.


rust/cubesql/cubesql/src/compile/engine/udf.rs
```diff
@@ -1198,6 +1198,95 @@ fn last_day_of_month(y: i32, m: u32) -> u32 {
...
+                Ok(Arc::new(result))
+            }
+            _ => Err(DataFusionError::Execution(
+                "unsupported interval type".to_string(),
+            )),
+        }
+    });
+
+    let return_type: ReturnTypeFunction = Arc::new(move |arg_types| {
+        if arg_types.len() != 2 {
```
ovr, line 1252: Simple `assert` should be enough. It looks useless because DF will check return_type after validating signature where we define 2 arguments. Any thoughts?

ovr: Oh, I see. Let's keep it here. It looks like we need to do the same behavior for all return_type functions, but it's a separate task.

ovr (approved):
LGTM

PR #5255, opened 2022-09-08
chore: Added driverType to readiness/liveness probe failed events

RusovDmitriy:


packages/cubejs-server-core/src/core/OrchestratorApi.ts
```diff
@@ -166,16 +167,17 @@ export class OrchestratorApi {
    * mode, to the datasources.
    */
   public async testConnection() {
+    console.log('testConnection', this.options);
```
buntarb, line 170: Hey, @RusovDmitriy. Could you please remove this?

packages/cubejs-query-orchestrator/src/orchestrator/QueryOrchestrator.ts
```diff
@@ -10,6 +10,12 @@ import { LocalQueueEventsBus } from './LocalQueueEventsBus';
 
 export type CacheAndQueryDriverType = 'redis' | 'memory';
 
+export enum DriverType {
+  external = 'external',
+  internal = 'internal',
+  cache = 'cache',
+}
```
buntarb, line 17: Usually, enum's keys are capital letters or started from capital letters. Maybe you can update keys?

packages/cubejs-api-gateway/src/gateway.ts
```diff
@@ -1600,27 +1605,22 @@ class ApiGateway {
   protected liveness: RequestHandler = async (req, res) => {
     let health: 'HEALTH' | 'DOWN' = 'HEALTH';
 
+    const handleError = (e: any) => {
```
buntarb, line 1608: Why do you need this wrapper function instead of using the existing method?

packages/cubejs-api-gateway/src/gateway.ts
```diff
@@ -1563,34 +1563,39 @@ class ApiGateway {
...
+    return 'DOWN';
+  }
+
   protected readiness: RequestHandler = async (req, res) => {
     let health: 'HEALTH' | 'DOWN' = 'HEALTH';
 
     if (this.standalone) {
       const orchestratorApi = await this.adapterApi({});
 
+      const handleError = (e: any) => {
```
buntarb, line 1582: Why do you need this wrapper function instead of using the existing method?

packages/cubejs-api-gateway/src/gateway.ts
RusovDmitriy: Just DRY. What existing method do you mean?

buntarb: Well, you created a `handleProbeError` that does all stuff. And instead of using it and just reassigning the `health` variable, you create another function, close this variable inside, and call this function to assign the variable. In my opinion, this makes code less readable and more buggy. Usually, we try to avoid closures if it is possible. And here, I don't see any significant reason for this. Does it make sense to you?

buntarb (approved):

PR #5265, opened 2022-09-08
feat(cubestore): Initial support for kv Cache

ovr:
Hello!

This PR introduces support for CACHE (key, value) storage in Cubestore + driver for Cube.

Supported commands:

```
CACHE GET
CACHE SET [NX] [TTL ?number] key value
CACHE REMOVE
CACHE TRUNCATE
CACHE KEYS
```

refs https://github.com/cube-js/cube.js/issues/5042

Thanks

### DIFF
--- a/rust/cubestore/cubestore/src/metastore/mod.rs
+++ b/rust/cubestore/cubestore/src/metastore/mod.rs
@@ -1,4 +1,6 @@
+pub mod cache;
 pub mod chunks;
+pub mod compaction;
 pub mod index;
 pub mod job;
 pub mod listener;
@@ -14,8 +16,8 @@ use async_trait::async_trait;
 use byteorder::{BigEndian, ReadBytesExt, WriteBytesExt};
 use log::info;
 use rocksdb::{
-    DBIterator, Direction, IteratorMode, MergeOperands, Options, ReadOptions, Snapshot, WriteBatch,
-    WriteBatchIterator, DB,
+    ColumnFamily, ColumnFamilyDescriptor, DBIterator, Direction, IteratorMode, MergeOperands,
+    Options, ReadOptions, Snapshot, WriteBatch, WriteBatchIterator, DB, DEFAULT_COLUMN_FAMILY_NAME,
 };
 use serde::{Deserialize, Deserializer, Serialize};
 use std::hash::{Hash, Hasher};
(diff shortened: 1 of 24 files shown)


packages/cubejs-query-orchestrator/src/orchestrator/PreAggregations.ts
```diff
@@ -1804,15 +1797,15 @@ export class PreAggregations {
 
   protected tablesUsedRedisKey(tableName) {
     // TODO add dataSource?
-    return `SQL_PRE_AGGREGATIONS_${this.redisPrefix}_TABLES_USED_${tableName}`;
+    return this.queryCache.getKey('SQL_PRE_AGGREGATIONS_TABLES_USED', tableName);
```
ovr, line 1800: *Notice*: Key change. I know

rust/cubestore/cubestore/src/metastore/mod.rs
```diff
@@ -137,16 +149,28 @@ macro_rules! base_rocks_secondary_index {
...
+                RocksSecondaryIndex::get_expire(self, row)
+            }
         }
     };
 }
 
 #[macro_export]
 macro_rules! rocks_table_impl {
-    ($table: ty, $rocks_table: ident, $table_id: expr, $indexes: block) => {
+    ($table: ty, $rocks_table: ident, $table_id: expr, $indexes: block, $cfn: expr) => {
```
waralexrom, line 173: May be `$cfn` -> `$column_family` ?  It's not obvious what `cfn` means then you're looking at the signature

ovr: Sure! Renamed.

rust/cubestore/cubestore/src/metastore/mod.rs
```diff
@@ -1632,18 +2052,24 @@ trait RocksTable: Debug + Send + Sync {
...
-                    .into_iter()
-                    .find(|i| i.get_id() == BaseRocksSecondaryIndex::get_id(secondary_index))
-                    .unwrap();
+                if RocksSecondaryIndex::is_ttl(secondary_index) {
+                    trace!(
+                        "Row exists in secondary index (with TTL) however missing in {:?} table: {}. Compaction problem?",
+                        self, id
+                    );
+
+                    continue;
```
waralexrom, line 2067: It's data consistency problem, not compaction (because of get_row_ids_by_index filter out expired ids). I think we should rebuild index at this case, or, at least put `TODO` about it here

rust/cubestore/cubestore/src/metastore/mod.rs
```diff
@@ -3383,6 +4039,129 @@ impl MetaStore for RocksMetaStore {
...
+            let rows =
+                cache_schema.get_rows_by_index(&index_key, &CacheItemRocksIndex::ByPrefix)?;
+
+            Ok(rows)
+        })
+        .await
+    }
+
+    async fn cache_set(&self, item: CacheItem, nx: bool) -> Result<bool, CubeError> {
+        self.write_operation_cache(move |db_ref, batch_pipe| {
```
waralexrom, line 4062: Please rename `nx` to something more understandable

PR #5325, opened 2022-09-20
feat(cubesql): Holistics - support range of charts

gandronchik:
**supported charts:**
- pivot table
- area chart
- line chart
- column chart
- bar chart
- combine chart

### DIFF
--- a/rust/cubesql/cubesql/src/compile/mod.rs
+++ b/rust/cubesql/cubesql/src/compile/mod.rs
@@ -1575,8 +1575,8 @@ mod tests {
         query.unwrap()
     }
 
-    fn find_cube_scan_deep_search(parent: Arc<LogicalPlan>) -> CubeScanNode {
-        pub struct FindCubeScanNodeVisitor(Option<CubeScanNode>);
+    fn find_cube_scans_deep_search(parent: Arc<LogicalPlan>) -> Vec<CubeScanNode> {
+        pub struct FindCubeScanNodeVisitor(Vec<CubeScanNode>);
 
         impl PlanVisitor for FindCubeScanNodeVisitor {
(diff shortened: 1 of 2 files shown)


MazterQyou (approved):

rust/cubesql/cubesql/src/compile/mod.rs
```diff
@@ -1631,7 +1638,11 @@ mod tests {
         }
 
         fn find_cube_scan(&self) -> CubeScanNode {
-            find_cube_scan_deep_search(Arc::new(self.clone()))
+            find_cube_scans_deep_search(Arc::new(self.clone()))[0].clone()
```
ovr, line 1641: We need to write down an assertation here that there is only one cube scan.

rust/cubesql/cubesql/src/compile/rewrite/rules/split.rs
```diff
@@ -1879,4 +1981,65 @@ impl SplitRules {
...
+            for fun in var_iter!(egraph[subst[fun_expr_var]], AggregateFunctionExprFun) {
+                match fun {
+                    AggregateFunction::Count | AggregateFunction::Sum => (),
+                    _ => return true,
+                };
+
+                for expr in var_iter!(egraph[subst[expr_val]], LiteralExprValue) {
+                    match expr {
+                        ScalarValue::Utf8(None) | ScalarValue::LargeUtf8(None) => return true,
+                        ScalarValue::Utf8(value) | ScalarValue::LargeUtf8(value) => match value {
```
MazterQyou, line 2002: I have concerns regarding this match, but let's go with this solution for the time being.

MazterQyou (approved):

PR #5477, opened 2022-10-18
fix(cubesql): Use real database name in Postgres meta layer

MazterQyou:
**Check List**

This PR fixes an issue with PostgreSQL meta layer using a hardcoded database name of `db`, resulting in `information_schema` and `pg_catalog` system tables to be inconsistent with the `current_database()` function. A related test is included.
Additionally the PR fixes a minor issue where `pg_catalog.pg_tables.tableowner` column would contain the database name instead of the user name.


ovr (approved):
LGTM + don't forget to check this PR with two different BIs that we didn't break anything!

PR #5545, opened 2022-10-31
docs(config): document multiple data source functionality

hassankhan:


rpaik (approved):

docs/content/Configuration/Advanced/Multiple-Data-Sources.mdx
```diff
@@ -0,0 +1,104 @@
...
+  dataSource: 'datasource1',
+});
+```
+
+Next, Cube must be configured so that it knows which database to use for each
+data source. This can be done either with the
+[`driverFactory`][ref-config-ref-driverfactory] property in the `cube.js`
+configuration file or through environment variables.
+
+## Configuring data sources with cube.js
```
keydunov, line 36: I think we want to focus only on env based management

hassankhan: I thought it would be good to contrast the approaches, especially since there's content in the existing Multitenancy page which really belongs here too.

PR #5585, opened 2022-11-08
feat(cubesql): Join Cubes

gandronchik:


rust/cubesql/cubeclient/src/apis/default_api.rs
```diff
@@ -115,7 +115,10 @@ pub async fn meta_v1(
 
     let local_var_client = &local_var_configuration.client;
 
-    let local_var_uri_str = format!("{}/v1/meta", local_var_configuration.base_path);
+    let local_var_uri_str = format!(
```
ovr, line 118: extended should be defined as a parameter in swagger + passed in our client

PR #5656, opened 2022-11-18
fix(cubejs-client-core): Update types to work with TS v4.8 and above

sacummings91:
**Check List**

[**Issue Reference this PR resolves**]

(https://github.com/cube-js/cube.js/issues/5446)

Small type changes to make tsc pass

### DIFF
--- a/packages/cubejs-client-core/index.d.ts
+++ b/packages/cubejs-client-core/index.d.ts
@@ -816,7 +816,7 @@ declare module '@cubejs-client/core' {
     never;
 
   type QueryArrayRecordType<T extends DeeplyReadonly<Query[]>> =
-    T extends readonly [infer First, ...infer Rest]
+    T extends readonly [infer First extends DeeplyReadonly<Query>, ...infer Rest]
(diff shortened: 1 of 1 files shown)


paveltiunov:
@sacummings91 Thanks for contributing it! cc @reify-thomas-smith

paveltiunov:
@sacummings91 Could you please fix all failing tests?

packages/cubejs-client-core/index.d.ts
```diff
@@ -816,7 +816,7 @@ declare module '@cubejs-client/core' {
     never;
 
   type QueryArrayRecordType<T extends DeeplyReadonly<Query[]>> =
-    T extends readonly [infer First, ...infer Rest]
+    T extends readonly [infer First extends DeeplyReadonly<Query>, ...infer Rest]
```
reify-thomas-smith, line 819: Won't this break compatibility with older versions of TypeScript?

sacummings91: I think you might be right about that, we can probably close this. I'm just going to set `"skipLibCheck: true"` in `tsconfig` instead as a work around

PR #5742, opened 2022-12-07
Rate limit bigquery dataset loading

paco-valdez:
**Check List**

TBD

When connecting to a bigquery db if a dataset contains a lot of tables, Cube will hit the rate limit for concurrent requests.

### DIFF
--- a/packages/cubejs-bigquery-driver/src/BigQueryDriver.ts
+++ b/packages/cubejs-bigquery-driver/src/BigQueryDriver.ts
@@ -194,10 +194,11 @@ export class BigQueryDriver extends BaseDriver implements DriverInterface {
 
   public async tablesSchema() {
     const dataSets = await this.bigquery.getDatasets();
-    const dataSetsColumns = await Promise.all(
-      dataSets[0].map((dataSet) => this.loadTablesForDataset(dataSet))
-    );
-
+    let dataSetsColumns = [];
+    while (dataSets[0].length) {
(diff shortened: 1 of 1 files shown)


pacofvf:
I set the rate limit to 10, although I'm not sure what is the ideal limit.

jediwarpraptor:
> @jediwarpraptor this is the change I had in mind, it looks like I need to add type hints. (I'm new to TS). The other proposed solution from Pavel was to add an env variable to only read from that dataset.

@pacofvf - I absolutely think this would fix the current issue I'm seeing in my environment.  Because I use BQ datasets liberally as the primary means of organization in the warehouse to separate/segregate based on both information domain (subscriptions, retail, wholesale, etc.) as well as ELT transformation stage (ingestion/source, staging, final, etc.), for my use-case (and others like mine), issuing hundreds of API calls through one big `.map()` immediately exceeds Google's rate limits.

re: the tweaking of that parameter, my gut tells me that `10` is a reasonable default. I don't know how easy it is within your codebase to abstract that to a config value which might be more performant in other conditions.  

My gut tells me that since this kind of metadata discovery will not need to be run often (at least during regular usage); combined with the fact that anyone who has fewer BigQuery datasets are unlikely to notice/perceive any performance slowdown related to this change.

tl;dr; Looks good to me! 
👍

jediwarpraptor:
Also apologies for not responding sooner, I put this project on hold until the beginning of the year and am only just coming back to it now.

paco-valdez:
We added a map-reduce logic on this logic, so it should be fixed in latest release.

PR #5789, opened 2022-12-12
feat(api-gateway, server-core): Added dataSources method

wolfeeeg:
**Check List**

[For example #12]

[Description goes here]

### DIFF
--- a/packages/cubejs-api-gateway/src/gateway.ts
+++ b/packages/cubejs-api-gateway/src/gateway.ts
@@ -1023,10 +1023,7 @@ class ApiGateway {
 
     try {
       const orchestratorApi = await this.getAdapterApi(context);
-      const { dataSources } = await this.getCompilerApi(context).dataSources({
-        requestId: context?.requestId,
-        skipQueryFactory: true,
-      }, orchestratorApi);
+      const { dataSources } = await this.getCompilerApi(context).dataSources(orchestratorApi);
 
       res({ dataSources });
     } catch (e) {
--- a/packages/cubejs-api-gateway/src/gateway.ts
+++ b/packages/cubejs-api-gateway/src/gateway.ts
@@ -366,6 +366,18 @@ class ApiGateway {
           });
         }
       );
+      
+      app.get(
+        '/cubejs-system/v1/data-sources',
+        jsonParser,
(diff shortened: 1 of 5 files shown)


packages/cubejs-server-core/src/core/CompilerApi.js
```diff
@@ -33,7 +33,7 @@ export class CompilerApi {
     return this.graphqlSchema;
   }
 
-  async getCompilers({ requestId } = {}) {
+  async getCompilers({ requestId, skipQueryFactory = false } = {}) {
```
wolfeeeg, line 36: we should skip queryFactory because inside it run driverFactory for all dataSources. 
And we need to check it one by one for security context

packages/cubejs-api-gateway/src/gateway.ts
```diff
@@ -990,6 +1002,27 @@ class ApiGateway {
       });
     }
   }
+  
+  protected async dataSources({ context, res }: { context: RequestContext, res: ResponseResultFn }) {
+    const requestStarted = new Date();
+
```
RusovDmitriy, line 1008: use `?` for context and check it inside function

MikeNitsenko (approved):

packages/cubejs-server-core/src/core/CompilerApi.js
```diff
@@ -187,6 +189,35 @@ export class CompilerApi {
     };
   }
 
+  async dataSources(options, orchestratorApi) {
```
buntarb, line 192: I would prefer if you'll keep `getCompilers` method in its original state, without changing its signature. Instead, you can partially duplicate its logic in the `dataSource` method. It is just a ~25 lines of code and I don't like this new `if` statement in the `getCompilers` method. You already added `dataSource` for your functionality, so lets split these two.

buntarb (changes requested):
See comments.

packages/cubejs-server-core/src/core/CompilerApi.js
wolfeeeg: done

buntarb (approved):

buntarb (approved):

PR #5834, opened 2022-12-15
fix(oracle-driver): wrong config setup

fech-dev:

This PR fixes the Oracle DB connection problem with the ` ORA-24415 Missing or null username` error.


paveltiunov:
Hey @fech-dev ! Nice catch! I guess it'd be better to close the connection in `finally` block. Also would love it if you could add a smoke test for the Oracle driver. Here's an example: https://github.com/cube-js/cube.js/blob/master/packages/cubejs-testing/test/smoke-materialize.test.ts.

ovr:
Hello @fetch-dev,

I am going to close this PR in favour of https://github.com/cube-js/cube.js/pull/5469.

Fixed & released in [v0.31.65](https://github.com/cube-js/cube.js/releases/tag/v0.31.65)

Thanks

Issue #673, opened 2020-05-26
Unable to use shortTitle in .series()


CMD-V:
In my ChartRenderer.js I was able to use `shortTitle` as seen below to remove the database prefix off of my table (antd) column titles.

```
  table: ({ resultSet }) => (
    <Table
      pagination={false}
      columns={resultSet.tableColumns().map(c => ({ title: c.shortTitle, key: c.key, dataIndex: c.key }))}
      dataSource={resultSet.tablePivot()}
    />
  )
```
When attempting to use this on my bar graph (chartjs) I was unable to get a similar result.

`  bar: ({ resultSet }) => {
    const data = {
      labels: resultSet.categories().map(c => c.category),
      datasets: resultSet.series().map((s, index) => ({
        label: s.shortTitle,
        data: s.series.map(r => r.value),
        backgroundColor: COLORS_SERIES[index],
        fill: false
      }))
    };
    const options = {
      scales: {
        xAxes: [
          {
            stacked: true
          }
        ]
      }
    };
    return <Bar data={data} options={options} />;`

here is where I am hoping to expose shortTitle:
https://github.com/cube-js/cube.js/blob/master/packages/cubejs-client-core/src/ResultSet.js#L74-L80

keydunov:
Hey @vasilev-alex @paveltiunov, just to give you some context here. We had a chat with @CMD-V  in Slack and it seems that `series()` method doesn't expose the `shortTitle` property, while `tableColumns()` does.

paveltiunov:
Yeah. Agree.

paveltiunov:
Contributions are welcomed here!

PR #5836, opened 2022-12-15
feat(@cubejs-client/core): expose shortTitle in series()

euljr:
**Check List**

Unable to use shortTitle in .series() #673

Hello Cube.JS team 👋 

I'm using Cube.JS client at Primer and we use the `.seriesNames()` and `.series()` methods. Recently, it was asked for us to use the `shortTitle` instead of `title`, and I found out that it isn't exposed in these methods (the same as the issue above).

This PR adds a few lines to expose `shortTitle` in `.seriesNames()` so we can have it in `.series()` as well.

Please forgive me for the naive implementation, as I don't understand very well what's happening here, I just did the same as it's done for the `title`.

I believe I can move the `pivotConfig.y.find(d => d === 'measures')` check to a boolean variable in order to avoid the extra `find` call, not sure what to name the variable though 🤔 

Let me know if it makes sense or if there is any other better way I'm not aware of, feel free to reach out if you need anything.

Regards,

Euler


paveltiunov:
Hey @euljr ! Looks good to me! Thanks for contributing!

PR #5838, opened 2022-12-15
fix(cubesql): Improve Thoughtspot compatibility

MazterQyou:
**Check List**

This PR improves compatibility with Thoughtspot by allowing certain expressions actively used by Thoughtspot.
List of supported expressions:

- Cast inside aggregate function (`SELECT AVG(CAST(expr AS FLOAT8))`)
- String concatenation with Cube dimension (`SELECT (str_dimension || 'literal')`)
- Equality check with extract in projection (`SELECT (EXTRACT(granularity FROM expr) = 'literal')`)
- Extracting month of quarter (`SELECT (((EXTRACT(MONTH FROM expr) - 1) % 3) + 1)`)
- Comparison check with date parts of two time dimensions in projection (`SELECT (EXTRACT(MONTH FROM expr1) < (EXTRACT(MONTH FROM expr2) + 1.0))`)
- Equality check in projection (`SELECT (LOWER(expr) = 'literal1' OR LOWER(expr) = 'literal2'`)
- Result of comparison of a dimension value with literal in projection (`SELECT (expr > 10.0)`)
- Dropping of doubling DateTrunc functions (`SELECT DATE_TRUNC('granularity', DATE_TRUNC('granularity', expr))`)
- Thoughtspot year DateTrunc expression (`SELECT DATE_TRUNC('month', CAST(CAST(((((EXTRACT(YEAR FROM expr) * 100) + 1) * 100) + 1) AS CHARACTER VARYING) AS timestamp))`)

Related tests for all the mentioned cases are included.


rust/cubesql/cubesql/src/compile/mod.rs
```diff
@@ -2384,6 +2384,7 @@ mod tests {
                 .dimensions,
             Some(vec![
                 "KibanaSampleDataEcommerce.order_date".to_string(),
```
gandronchik, line 2386: last_mod? Does it mean modified at?

rust/cubesql/cubesql/src/compile/rewrite/rules/split.rs
```diff
@@ -1544,6 +1680,77 @@ impl RewriteRules for SplitRules {
...
+                    agg_fun_expr("?fun", vec![column_expr("?column")], "?distinct"),
+                    "?alias_to_cube",
+                ),
+            ),
+            transforming_chain_rewrite(
+                "split-push-down-aggr-fun-cast-arg-float64-outer-replacer",
+                outer_projection_split_replacer("?expr", "?alias_to_cube"),
+                vec![(
+                    "?expr",
+                    agg_fun_expr(
```
gandronchik, line 1707: I think outer_projection_split_replacer can not contains agg_fun_expr

rust/cubesql/cubesql/src/compile/mod.rs
MazterQyou: Yeah, that was the idea. Think that's good enough or do you propose renaming it?

rust/cubesql/cubesql/src/compile/rewrite/rules/split.rs
MazterQyou: Lines L421-437 and L1614-1623 didn't have the problem with agg_fun_expr inside outer_projection. We're moving the cast here and it only affects Float64, so this should be bijective.

gandronchik (approved):

PR #5883, opened 2022-12-26
chore(cubestore): Improve safety on compaction

ovr:
Hello!

Any kind of migration doesn't remove old values from the database; it's why compaction should remove them.

Thanks


waralexrom (approved):
LGTM

PR #5933, opened 2023-01-06
feat(cubestore): Support lazy initialization for CacheStore

ovr:


rust/cubestore/cubestore/src/cachestore/lazy.rs
```diff
@@ -0,0 +1,168 @@
...
+        let mut guard = self.state.write().await;
+        match &*guard {
+            LazyRocksCacheStoreState::FromRemote {
+                path,
+                metastore_fs,
+                config,
+                // receiver will be closed on drop
+                init_flag: _,
+            } => {
+                let store =
```
ovr, line 78: I tried to use move logic with initializing state & std::mem::replace, but I realized that load_from_remote may fail.

rust/cubestore/cubestore/src/cachestore/lazy.rs
```diff
@@ -0,0 +1,168 @@
...
+        self.get_initialized_store()
+            .await
+            .unwrap()
+            .wait_upload_loop()
+            .await;
+    }
+
+    pub async fn stop_processing_loops(&self) {
+        if let Ok(store) = self.get_initialized_store().await {
+            store.stop_processing_loops().await;
```
ovr, line 121: Should I introduce a new state (`Closed`) to protect a problem when someone calls `CacheStore->init` after`stop_processing_loops`?

waralexrom (approved):
LGTM 👍

PR #6017, opened 2023-01-17
Sql runner test

wolfeeeg:
**Check List**

[For example #12]

[Description goes here]

### DIFF
--- a/packages/cubejs-api-gateway/src/gateway.ts
+++ b/packages/cubejs-api-gateway/src/gateway.ts
@@ -1161,15 +1161,18 @@ class ApiGateway {
 
       const orchestratorApi = await this.getAdapterApi(context);
       const schema = await orchestratorApi.fetchSchema(query.dataSource, context?.securityContext, query.renew);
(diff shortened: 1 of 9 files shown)


packages/cubejs-api-gateway/src/gateway.ts
```diff
@@ -1058,17 +1058,25 @@ class ApiGateway {
...
+        );
+      }
+
       if (query.resultFilter?.objectTypes && !Array.isArray(query.resultFilter.objectTypes)) {
         throw new UserError(
           'A query.resultFilter.objectTypes must be an array of strings'
         );
       }
+
+      query.query = `${query.query.replace(';', '')} LIMIT ${query.limit};`;
```
ovr, line 1079: Maybe it's better to do `SELECT * FROM (${query.query}) LIMIT ${query.limit}` instead of hacking?)

wolfeeeg: yeah we think about it, but will it work on all db types?

PR #6018, opened 2023-01-17
feat(athena-driver): add session-token and database env vars

paulorsbrito:
**Check List**

Add env vars config for Athena session token and use `CUBEJS_DB_NAME` for Athena default database.


paveltiunov:
Hey @paulorsbrito ! Thanks for contributing! We already have `CUBEJS_DB_NAME` supported by the Athena driver. It is just not provided as `Database` connection property. Could you please elaborate on why do you want to provide `Database` property? Is it to avoid using fully qualified names?

darapuk:
Instead of this, cant you just push your credentials into AWS default environment variables? The AWS sdk should handle the above case automatically.

As an example, the below would enable authentication to work as intended:
```
docker run -p 4000:4000 -p 5432:5432 \
  -v ${PWD}:/cube/conf \
  -e CUBEJS_DEV_MODE=true \
  -e AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID \
  -e AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY \
  -e AWS_SESSION_TOKEN=$AWS_SESSION_TOKEN \
  cubejs/cube
```

paulorsbrito:
That really worked, @darapuk ! Thanks! 

As far as I could see in the docs, there's no such env var for default database, so I'll keep that specific changes

josezambrana:
I couldn't find a way to configure a database different than `default`  for athena, but I find this PR can solve this that problems. Is there some way to promote this to be integrated?

PR #6035, opened 2023-01-20
feat(api-gateway, server-core): Added limit for sql-runner method

wolfeeeg:
**Check List**

[For example #12]

[Description goes here]


packages/cubejs-api-gateway/src/gateway.ts
```diff
@@ -1020,17 +1020,25 @@ class ApiGateway {
...
+        );
+      }
+
       if (query.resultFilter?.objectTypes && !Array.isArray(query.resultFilter.objectTypes)) {
         throw new UserError(
           'A query.resultFilter.objectTypes must be an array of strings'
         );
       }
+
+      query.query = `${query.query.replace(';', '')} LIMIT ${query.limit};`;
```
wolfeeeg, line 1041: image

paveltiunov, line 1041: There could be DDL queries also.

MikeNitsenko: @paveltiunov do you mean that we should:
- prohibit the use of DDL queries
- allow the use of DDL queries and support them respectively

Issue #6023, opened 2023-01-17
Redshift changes order of preaggregation table columns causing union all to fail when filtering over date ranges that extend through multiple tables


jskarda829:
**Describe the bug**
When querying data preaggregated from Redshift on a monthly scale, the tables can have their columns ordered differently causing the union all between the tables of the generated query to fail because of the different schema. 

**To Reproduce**
1. Have multiple years of data in Redshift to be preaggregated.
2. Preaggregate based on the following, using S3 Export Bucket(full cube setup further down)
```preAggregations: {
    action_count: {
      measures: [Actions.count],
      dimensions: [
            PropertyActions1.property_id, 
            PropertyActions2.property_id,
            PropertyActions3.property_id, 
            PropertyActions4.property_id, 
            PropertyActions5.property_id,
            PropertyActions6.property_id, 
            Actions.action_type
      ],
      indexes: {
        property1Index: {
          columns: [PropertyActions1.property_id]
        },
        property2Index: {
          columns: [PropertyActions2.property_id]
        },
        property3Index: {
          columns: [PropertyActions3.property_id]
        },
        property4Index: {
          columns: [PropertyActions4.property_id]
        },
        property5Index: {
          columns: [PropertyActions5.property_id]
        },
        property6Index: {
          columns: [PropertyActions6.property_id]
        }
      },
      timeDimension: Actions.timestamp,
      granularity: `day`,
      partitionGranularity: `month`,
      refreshKey: {
        every: `1 day`,
        incremental: true,
        updateWindow: `7 day`
      }
    },
  }
  ```
3. Query for all dimensions for all time and you may receive a UNION ALL error.
4. Using the MySQL CLI client, run ```select * from system.tables```. The column orders will be different.

**Expected behavior**
Expected behavior is that querying on preaggregations across all time works.
Actual behavior is that if the preaggregations have differently ordered columns the query fails because the table schema is different between them.

**Error messages and output**
```cube_1                 | Error while querying: e8d6320f-86ed-43bf-b906-b83494081d06-span-1 (12ms)
cube_1                 | {
cube_1                 |   "queueSize": 0,
cube_1                 |   "queryKey": [
[...]

PR #6068, opened 2023-01-27
fix(redshift-driver): fixes column order

rdwoodring:
**Check List**

#6023

Referenced changes made to Athena driver in [PR 4183](https://github.com/cube-js/cube.js/pull/4183) and updated the unload query in the Redshift driver to match.

### DIFF
--- a/packages/cubejs-redshift-driver/src/RedshiftDriver.ts
+++ b/packages/cubejs-redshift-driver/src/RedshiftDriver.ts
@@ -146,10 +146,13 @@ export class RedshiftDriver extends PostgresDriver<RedshiftDriverConfiguration> 
     return false;
   }
 
-  public async unload(table: string, options: UnloadOptions): Promise<DownloadTableCSVData> {
+  public async unload(tableName: string, options: UnloadOptions): Promise<DownloadTableCSVData> {
     if (!this.config.exportBucket) {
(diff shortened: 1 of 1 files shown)


rdwoodring:
@paveltiunov I think I brought over the changes correctly from the Athena fix into the Redshift driver. It looks like the Redshift driver is lacking tests and I wasn't sure enough of myself to try and abstract anything away, so it's not the most elegant change.

paveltiunov:
@rdwoodring Hey Ryan! Thanks for contributing! I believe `types` should be returned in all success path of unload method otherwise columns would still be mismatched.

rdwoodring:
@paveltiunov ok, I think I have that fixed up. What about the `csvNoHeaders: true` in the [AthenaDriver](https://github.com/cube-js/cube.js/blob/eedd12a18645d2d6faa68a43963a6cb98560b5a6/packages/cubejs-athena-driver/src/AthenaDriver.ts#L246)? Should that also be included in the last return statement in the RedshiftDriver ([here-ish](https://github.com/cube-js/cube.js/pull/6068/files#diff-9d3f29fa5e7b2077845929ff3a008263c957108b143f1b882827d7d4788695b6R248))?

paveltiunov:
@rdwoodring Yep. That's correct! Looks good to me!

PR #6125, opened 2023-02-06
test(query-orchestrator): Temporary disable orphaned test for Cube Store

ovr:
Hello!

Thanks

### DIFF
--- a/packages/cubejs-query-orchestrator/test/unit/QueryQueue.abstract.ts
+++ b/packages/cubejs-query-orchestrator/test/unit/QueryQueue.abstract.ts
@@ -21,8 +21,8 @@ export const QueryQueueTest = (name: string, options: QueryQueueTestOptions = {}
 
     let delayCount = 0;
     let streamCount = 0;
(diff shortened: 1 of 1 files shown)


packages/cubejs-query-orchestrator/test/unit/QueryQueue.abstract.ts
```diff
@@ -63,11 +63,9 @@ export const QueryQueueTest = (name: string, options: QueryQueueTestOptions = {}
     async function awaitProcessing() {
       // process query can call reconcileQueue
       while (await queue.shutdown() || processMessagePromises.length || processCancelPromises.length) {
-        await Promise.all(processMessagePromises);
-        processMessagePromises = [];
-
-        await Promise.all(processCancelPromises);
-        processCancelPromises = [];
+        await Promise.all(processMessagePromises.splice(0).concat(
```
ovr, line 66: `= []` can lead to a leak of promises, it's better to use splice as atomic way

PR #6142, opened 2023-02-09
feat(server-core, query-orchestrator): sql-runner - supp…

wolfeeeg:
…ort running queries against extDb

**Check List**

[For example #12]

[Description goes here]

### DIFF
--- a/packages/cubejs-cubestore-driver/src/CubeStoreDriver.ts
+++ b/packages/cubejs-cubestore-driver/src/CubeStoreDriver.ts
@@ -81,7 +81,15 @@ export class CubeStoreDriver extends BaseDriver implements DriverInterface {
   }
 
   public informationSchemaQuery() {
-    return `${super.informationSchemaQuery()} AND columns.table_schema = '${this.config.database}'`;
+    return `
+      SELECT columns.column_name as ${this.quoteIdentifier('column_name')},
+             columns.table_name as ${this.quoteIdentifier('table_name')},
+             columns.table_schema as ${this.quoteIdentifier('table_schema')},
+             columns.data_type as ${this.quoteIdentifier('data_type')}
+      FROM information_schema.columns as columns
+      WHERE columns.table_schema NOT IN ('information_schema', 'mysql', 'performance_schema', 'sys', 'INFORMATION_SCHEMA')
+      AND columns.table_schema = '${this.config.database}'
+   `;
   }
 
   public createTableSqlWithOptions(tableName, columns, options: CreateTableOptions) {
--- a/packages/cubejs-query-orchestrator/src/orchestrator/QueryCache.ts
+++ b/packages/cubejs-query-orchestrator/src/orchestrator/QueryCache.ts
@@ -429,6 +429,7 @@ export class QueryCache {
       useCsvQuery,
       persistent,
       aliasNameToMember,
+      tablesSchema,
     }: {
       cacheKey: CacheKey,
       dataSource: string,
@@ -439,6 +440,7 @@ export class QueryCache {
       useCsvQuery?: boolean,
       persistent?: boolean,
       aliasNameToMember?: { [alias: string]: string },
+      tablesSchema?: boolean,
     }
   ) {
     const queue = external
@@ -452,6 +454,7 @@ export class QueryCache {
       requestId,
       inlineTables,
       useCsvQuery,
+      tablesSchema,
     };
 
     const opt = {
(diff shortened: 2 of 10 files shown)


packages/cubejs-server-core/src/core/OrchestratorApi.ts
```diff
@@ -129,7 +129,10 @@ export class OrchestratorApi {
         return res;
       }
 
-      data.dbType = await extractDbType(data);
+      if (!query.external) {
```
wolfeeeg, line 132: Please pay attention on this line, `extractDbType` method use accessing the driver.
When we make request to external driver without `dataSource`, it will call regular driver that can lead to an exception because `dataSource` is empty.

packages/cubejs-cubestore-driver/src/CubeStoreDriver.ts
```diff
@@ -81,7 +81,15 @@ export class CubeStoreDriver extends BaseDriver implements DriverInterface {
...
 
   public informationSchemaQuery() {
-    return `${super.informationSchemaQuery()} AND columns.table_schema = '${this.config.database}'`;
+    return `
+      SELECT columns.column_name as ${this.quoteIdentifier('column_name')},
+             columns.table_name as ${this.quoteIdentifier('table_name')},
+             columns.table_schema as ${this.quoteIdentifier('table_schema')},
+             columns.data_type as ${this.quoteIdentifier('data_type')}
+      FROM information_schema.columns as columns
+      WHERE columns.table_schema NOT IN ('information_schema', 'mysql', 'performance_schema', 'sys', 'INFORMATION_SCHEMA')
```
wolfeeeg, line 90: this line may be redundant

buntarb (changes requested):
See comments.

packages/cubejs-cubestore-driver/src/CubeStoreDriver.ts
wolfeeeg: nope, it doesn't. The difference in this line: 
`FROM information_schema.columns as columns`
The cubestore doesn't make an alias automatically.

packages/cubejs-server-core/src/core/CompilerApi.js
```diff
@@ -240,6 +240,14 @@ export class CompilerApi {
       })
     );
 
+    if (orchestratorApi.options?.externalDriverFactory) {
```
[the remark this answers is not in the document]
wolfeeeg: We need to get a special dataSource for `extDb`.  So if the user setup externalDriverFactory inside cube.js we will add one more dataSource called `externalDb`

packages/cubejs-server-core/src/core/OrchestratorApi.ts
wolfeeeg: I'm not sure that is my specific case. 
If you will use the `executeQuery` method with `query.external = true` and not specify `dataSource` before then it will throw an Error inside the `extractDbType`. 
And as understand we do not need `dataSource` and dbType in the external case at all.

packages/cubejs-query-orchestrator/src/orchestrator/QueryCache.ts
```diff
@@ -988,8 +992,8 @@ export class QueryCache {
     return this.cacheDriver.testConnection();
   }
   
-  public async fetchSchema(dataSource: string) {
-    const queue = await this.getQueue(dataSource);
+  public async fetchSchema(dataSource?: string, isExternalQuery?: boolean) {
```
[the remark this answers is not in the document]
wolfeeeg: done

packages/cubejs-server-core/src/core/OrchestratorApi.ts
buntarb: As far as I remember, `query.external` flag could be set by the `QueryOrchestrator.getPreAggregationPreview` method only and only in a case when `PreAggregation` is configured as an `external` (it stores in an external data source not the Cubestore). The `OrchestrationApi.executeQuery` is called before this moment, so the `external` flag couldn't be true here.
On the other hand, I found what concerns me here: you call this method from the `ApiGateway.sqlRunner` method. However, we already have the `ApiGateway.getSqlResponseInternal` method designed for this. So, instead of producing a new pipeline for the query execution, I prefer to utilize the existing one. Please be able to change the `ApiGateway.sqlRunner` method in the following way: prepare the query object according to the `ApiGateway.getSqlResponseInternal` method, call it, and process the result according to your needs. This will keep the single query execution pipeline and disambiguate it.

packages/cubejs-server-core/src/core/CompilerApi.js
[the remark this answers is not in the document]
buntarb: What will you get if the user creates a more complicated `externalDriverFactory` depending on the `context`?

buntarb (changes requested):
See comments.

packages/cubejs-cubestore-driver/src/CubeStoreDriver.ts
wolfeeeg: Without this fixes we have an error 
Screenshot 2023-03-20 at 14 37 43

buntarb (approved):

PR #6143, opened 2023-02-09
Clickhouse string ordering and string filtering by UTF8 instead of bytes

casab:
Clickhouse defaults to using bytes to order by and string manipulation functions such as `lower`, `upper` uses ascii. To overcome this limitation they have `COLLATE` keyword, and `lowerUTF8`, `upperUTF8` functions.

**Check List**

- Replaced `CONCAT` SQL function with js template literal to prevent unnecessary DB function call
- Used `lowerUTF8` instead of `lower` to support utf8 compatible search
- Added `COLLATE ‘en’` when ordering by strings to order incasesensitive. Clickhouse orders by bytes on default.

### DIFF
--- a/packages/cubejs-schema-compiler/src/adapter/ClickHouseQuery.js
+++ b/packages/cubejs-schema-compiler/src/adapter/ClickHouseQuery.js
@@ -16,7 +16,7 @@ class ClickHouseFilter extends BaseFilter {
   likeIgnoreCase(column, not, param, type) {
     const p = (!type || type === 'contains' || type === 'ends') ? '%' : '';
     const s = (!type || type === 'contains' || type === 'starts') ? '%' : '';
-    return `lower(${column}) ${not ? 'NOT' : ''} LIKE CONCAT('${p}', lower(${this.allocateParam(param)}), '${s}')`;
+    return `lowerUTF8(${column}) ${not ? 'NOT' : ''} LIKE lowerUTF8(${p}${this.allocateParam(param)}${s})`;
   }
 
   castParameter() {
@@ -147,6 +147,28 @@ export class ClickHouseQuery extends BaseQuery {
 
     const direction = hash.desc ? 'DESC' : 'ASC';
     return `${fieldAlias} ${direction}`;
+  }
+
+  orderBy() {
+    //
+    // ClickHouse orders string by bytes, so we need to use COLLATE 'en' to order by string
+    //
+    if (R.isEmpty(this.order)) {
+      return '';
+    }
+
+    const orderByString = R.pipe(
+      R.map(this.orderHashToString),
+      R.reject(R.isNil),
+      R.join(' COLLATE \'en\''),
+      R.join(', ')
+    )(this.order);
+
+    if (!orderByString) {
+      return '';
+    }
+
+    return ` ORDER BY ${orderByString}`;
   }
 
   groupByClause() {


casab:
> Hey @casab ! Thanks for contributing! Could you please add an integration test for it?

Hey Pavel, I have finally added an integration test but I am not sure that it runs. I see integration test runners in .github/actions/integration/clickhouse.sh but don't really know much Javascript or lerna(?)

igorlukanin:
@casab May I kindly ask you to test with the latest release and rebase your changes on top of it? We have migrated to a new ClickHouse client library recently. Thanks in advance!

casab:
@igorlukanin Can you approve running the workflows?

KSDaemon:
Hi @casab Could you please sync with the latest master and resolve conflicts? Ping me whenever you need to approve running workflows!

KSDaemon:
@casab Hey! Thnx for the updates! Unfortunately, some lint errors still exists:
```
  /home/runner/work/cube/cube/packages/cubejs-schema-compiler/src/adapter/ClickHouseQuery.ts
    178:1   error    Expected indentation of 6 spaces but found 8  indent
    178:16  warning  Strings must use singlequote                  quotes
    182:1   error    Expected indentation of 6 spaces but found 8  indent
    182:26  error    Unexpected string concatenation               prefer-template
    182:58  warning  Strings must use singlequote                  quotes
    183:1   error    Expected indentation of 6 spaces but found 8  indent
    184:1   error    Expected indentation of 6 spaces but found 8  indent
    184:16  warning  Strings must use singlequote                  quotes
    188:1   error    Expected indentation of 6 spaces but found 8  indent
    188:16  warning  Strings must use singlequote                  quotes
    192:1   error    Expected indentation of 2 spaces but found 0  indent
```

Can you fix them, please?

casab:
@KSDaemon Yeah sorry about that. I have fixed the linter issues now. It's just that this PR is haunting me for 2 years...

KSDaemon:
Ah.. Sorry to hear that! But... Don't worry! We'll nail it down!

casab:
Let me check these integration test results and I'll make the necessary changes.

casab:
@KSDaemon I think it's alright now. Apparently I wasn't checking if the order field's type is string or not. I also modified the other integration tests for the new test case I have added.

KSDaemon:
@casab Yeah, great addition, that was one of my questions, I want to ask! :)
What about the collation: what if the underlying values are in a different language other than English? Current code hardcode the `collation 'en'`.

casab:
@KSDaemon I thought about it. When I created this PR, we were having problems with Turkish ordering in the company I used to work. But the biggest problem is not even the Turkish characters. Ordering is not case insensitive. 'en' would solve huge portion of that. I thought about adding a parameter to the query. Or maybe fetch it from cube definition.

What do you suggest? Where can we get the selected collation from?

packages/cubejs-schema-compiler/test/integration/clickhouse/ClickHouseDbRunner.ts
```diff
@@ -65,7 +65,8 @@ export class ClickHouseDbRunner extends BaseDbRunner {
       (3, 300, '2017-01-05 16:00:00', '2017-01-19 16:00:00', 2, 'google', 120.120, 70.60),
       (4, 400, '2017-01-06 16:00:00', '2017-01-24 16:00:00', 2, null, 120.120, 10.60),
       (5, 500, '2017-01-06 16:00:00', '2017-01-24 16:00:00', 2, null, 120.120, 58.10),
-      (6, 500, '2016-09-06 16:00:00', '2016-09-06 16:00:00', 2, null, 120.120, 58.10)
+      (6, 500, '2016-09-06 16:00:00', '2016-09-06 16:00:00', 2, null, 120.120, 58.10),
+      (7, 300, '2017-01-07 16:00:00', '2017-01-25 16:00:00', 2, 'Google', 120.120, 59.60)
```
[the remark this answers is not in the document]
casab: since Clickhouse orders by bytes. When ordered `Google` would always be after `some`. So I had to add some other entry that starts with uppercase G.

packages/cubejs-schema-compiler/test/integration/clickhouse/clickhouse-dataschema-compiler.test.ts
```diff
@@ -299,6 +300,7 @@ describe('ClickHouse DataSchemaCompiler', () => {
         expect(res).toEqual(
           [
             { visitors__enabled_source: 'google', visitors__visitor_count: '1' },
+            { visitors__enabled_source: 'Google', visitors__visitor_count: '1' },
```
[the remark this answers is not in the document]
casab: You are right about this. The reason that I added an entry that starts with G is explained in the other comment. But I need to change `Google` to something else that starts with G.

packages/cubejs-schema-compiler/src/adapter/ClickHouseQuery.ts
```diff
@@ -168,6 +184,33 @@ export class ClickHouseQuery extends BaseQuery {
...
+    //
+    if (R.isEmpty(this.order)) {
+      return '';
+    }
+
+    const orderByString = R.pipe(
+      R.map((order) => {
+        let orderString = this.orderHashToString(order);
+        if (this.getFieldType(order) === 'string') {
+          orderString = `${orderString} COLLATE 'en'`;
```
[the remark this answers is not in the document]
casab: This might work. But to be honest, being able to send the collation per api request or per ordering field would be better. But this might be the easiest sollution. So I am going to add this, then maybe we can figure out more granular collation setting.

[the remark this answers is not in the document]
KSDaemon: Don't forget that there are more than just a REST API: SQL API and GraphQL API, and they all should work in the same way. And I think it would be tricky to preserve the same behavior across all APIs... So, let's start with the simple and obvious solution...

packages/cubejs-schema-compiler/src/adapter/ClickHouseQuery.ts
```diff
@@ -168,6 +184,35 @@ export class ClickHouseQuery extends BaseQuery {
...
+      R.map((order) => {
+        let orderString = this.orderHashToString(order);
+        if (this.getFieldType(order) === 'string') {
+          orderString = `${orderString} COLLATE '${collation}'`;
+        }
+        return orderString;
+      }),
+      R.reject(R.isNil),
+      R.join(', ')
+    )(this.order);
```
KSDaemon, line 207: What if someone uses ascii and/or doesn't need any collation sorting at all? Also with this flow, we preserve the backward compatibility because, in other cases, someone might be surprised with the new results if their data is in non-en locale...

```suggestion
    const collation = getEnv('clickhouseSortCollation');

    const orderByString = R.pipe(
      R.map((order) => {
        let orderString = this.orderHashToString(order);
        if (this.getFieldType(order) === 'string') && collation !== '' {
          orderString = `${orderString} COLLATE '${collation}'`;
        }
        return orderString;
      }),
      R.reject(R.isNil),
      R.join(', ')
    )(this.order);
```

packages/cubejs-schema-compiler/src/adapter/ClickHouseQuery.ts
```diff
@@ -168,6 +184,35 @@ export class ClickHouseQuery extends BaseQuery {
...
+    if (R.isEmpty(this.order)) {
+      return '';
+    }
+
+    const collation = getEnv('clickhouseSortCollation');
+
+    const orderByString = R.pipe(
+      R.map((order) => {
+        let orderString = this.orderHashToString(order);
+        if (this.getFieldType(order) === 'string') && collation !== '' {
```
KSDaemon, line 200: Sorry for the mess, after thinking for a while, I agree with your original suggestion to fall back to `en` locale as it will work more correctly even for data in other locales...

mcheshkov: As written before it didn't allow removing `COLLATE` at all, and I think it's important to have a way to disable it. At the very least, to keep old behaviour for users, who for any reason does not expect this change.
I think we can enable it by default, but way to opt-out from using collations is a must: without opt-out there's no way to make ordering case-sensitive, and even if we want to force it, users would have only one way to fix this during migration - revert back to previous release and keep waiting.

I, personally, would not expect it to use collation by default, but I also find default collations in PostgreSQL or MySQL strange, so maybe it's just me.

mcheshkov: Something I wrote when reviewing previous version:

_Any_ collation in order by [makes ordering case-insensitive](https://clickhouse.com/docs/en/sql-reference/statements/select/order-by#collation-support) in ClickHouse, and it would not match `orderBy` for any other DB adapter in Cube, at least from ops perspective.

Consider query like this:
```
SELECT
    s
FROM (
    SELECT
        'a' AS s
    UNION ALL
    SELECT
        'A' AS s
    UNION ALL
    SELECT
        'a' AS s
    UNION ALL
    SELECT
        'b' AS s
    UNION ALL
    SELECT
        'B' AS s
    UNION ALL
    SELECT
        'b' AS s
    ) t
ORDER BY
    s COLLATE 'en'
;
```

With `COLLATE 'en'` ClickHouse would return all `a` and `A`, and then all `b` and `B`.
But for DBMS like PostgreSQL this would depend on type of column (and encoding, for that matter), so admin of warehouse can decide how to order those, not admin of Cube, and they can decide whether it should or shouldn't be case sensitive.

Also keep in mind that Cube can issue ungrouped queries with order and limit, that would need to sort very large set of rows, so performance impact from enabling it can (I think) be huge in some specific cases, and specifically it can be unexpected for ClickHouse

packages/cubejs-schema-compiler/src/adapter/ClickHouseQuery.ts
```diff
@@ -18,7 +20,7 @@ class ClickHouseFilter extends BaseFilter {
   public likeIgnoreCase(column, not, param, type) {
     const p = (!type || type === 'contains' || type === 'ends') ? '%' : '';
     const s = (!type || type === 'contains' || type === 'starts') ? '%' : '';
-    return `lower(${column}) ${not ? 'NOT' : ''} LIKE CONCAT('${p}', lower(${this.allocateParam(param)}), '${s}')`;
+    return `lowerUTF8(toValidUTF8(${column})) ${not ? 'NOT' : ''} LIKE CONCAT('${p}', lowerUTF8(toValidUTF8(${this.allocateParam(param)})), '${s}')`;
```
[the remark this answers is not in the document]
casab: I am not sure if ILIKE is UTF-8 compatible. It might just be calling lower(). Let me ask this to the Clickhouse folk first.

packages/cubejs-schema-compiler/src/adapter/ClickHouseQuery.ts
casab: I agree on being able to disable collation behavior. We might document it to set the env variable to a specific value. Or I might add an another env variable.

The problem with Clickhouse is that it only allows collation in the order by clause. Nothing else, so there is no database level, table level, column level, session level collation setting anywhere. Otherwise I would agree.

packages/cubejs-schema-compiler/src/adapter/ClickHouseQuery.ts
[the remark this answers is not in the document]
mcheshkov: It is, for UTF-8 strings `_` has a "single code point" meaning

https://clickhouse.com/docs/en/sql-reference/functions/string-search-functions#ilike

casab:
@KSDaemon @mcheshkov Hey, I have fixed all the errors. And implemented all the required parts. I would appreciate it if you can review it. Especially the rust part.

packages/cubejs-schema-compiler/src/adapter/ClickHouseQuery.ts
```diff
@@ -168,6 +188,43 @@ export class ClickHouseQuery extends BaseQuery {
...
+      R.map((order) => {
+        let orderString = this.orderHashToString(order);
+        if (collation && this.getFieldType(order) === 'string') {
+          orderString = `${orderString} COLLATE '${collation}'`;
+        }
+        return orderString;
+      }),
+      R.reject(R.isNil),
+      R.join(', ')
+    )(this.order);
```
KSDaemon, line 219: Don't you mind rewriting this in plain JS instead of using Ramda?
Smth like this:
```suggestion
if (this.order.length === 0) {
  return '';
}

const collation = this.getCollation();

const orderByString = this.order
  .map((order) => {
    let orderString = this.orderHashToString(order);
    if (collation && this.getFieldType(order) === 'string') {
      orderString = `${orderString} COLLATE '${collation}'`;
    }
    return orderString;
  })
  .filter(Boolean) // Analogue `R.reject(R.isNil)`
  .join(', ');
```

packages/cubejs-schema-compiler/src/adapter/ClickHouseQuery.ts
```diff
@@ -281,6 +338,22 @@ export class ClickHouseQuery extends BaseQuery {
...
+    if (collation) {
+      templates.expressions.sort = `${templates.expressions.sort}{% if data_type and data_type == 'string' %} COLLATE '${collation}'{% endif %}`;
+      templates.expressions.order_by = `${templates.expressions.order_by}{% if data_type and data_type == 'string' %} COLLATE '${collation}'{% endif %}`;
+
+      const oldOrderBy = '{% if order_by %}\nORDER BY {{ order_by | map(attribute=\'expr\') | join(\', \') }}{% endif %}';
+
+      const newOrderBy =
+        '{% if order_by %}\nORDER BY {% for item in order_by %}{{ item.expr }}' +
+        `{%- if item.data_type and item.data_type == 'string' %} COLLATE '${collation}'{% endif %}` +
+        '{%- if not loop.last %}, {% endif %}{% endfor %}{% endif %}';
```
MazterQyou, line 353: I dislike the idea of replacing strings by putting the original `ORDER BY` expression here. The `statements/select` template might change over time in BaseQuery, and if `ORDER BY` condition is touched, this code would break.

I would suggest moving **whole** `statements/select` template here, then deciding the `ORDER BY` part based on the provided condition. This way if BaseQuery' `statements/select` is modified, this dialect would be untouched and would continue working regardless. We already use similar approach with overriding whole `statements/select` template in `PinotQuery` and `PrestodbQuery`.

PR #6154, opened 2023-02-13
fix: Correct the syntax error in the ClickHouse cast method

fxx-real:
**Check List**

Hello，

I am using cube-js (version 0.31.60) with a data source of ClickHouse (version 23.1.2.9) and I have discovered that when there are two primaryKeys in the count method, the following branch logic is executed:

`this.concatStringsSql(primaryKeys.map((pk) => this.castToString(this.primaryKeySql(pk, cubeName))))`

The castToString(sql) method contains a sql syntax error and the generated SQL will result in an error when executed in ClickHouse:

> “Code: 50. DB::Exception: Unknown data type family: STRING. Maybe you meant: ['String','Ring']: While processing concat(toString(CAST(id, 'STRING')), toString(CAST(name, 'STRING'))). (UNKNOWN_TYPE) (version 23.1.2.9 (official build))”

In my test case, the generated SQL is as follows:

`SELECT count(toString(CAST(`test`.id as STRING)) || toString(CAST(`test`.name as STRING))) `test__count` FROM my_test.test AS `test``

In the current version of ClickHouse, the correct parameter for the cast method should be 'String'.
https://clickhouse.com/docs/en/sql-reference/functions/type-conversion-functions/#castx-t

After changing 'STRING' to 'String' in the ClickHouseQuery.castToString(sql) method, the program can run successfully. The generated SQL is as follows：

`SELECT count(toString(CAST(`test`.id as String)) || toString(CAST(`test`.name as String))) `test__count` FROM my_test.test AS `test``


fxx-real:
This is my schema file.
Pasted Graphic 2

Pasted Graphic

paveltiunov:
@fxx-real Thanks for contributing! Could you please add an integration test for that in `cubejs-testing`?

fxx-real:
> @fxx-real Thanks for contributing! Could you please add an integration test for that in `cubejs-testing`?

Sure, I'd like to, but I'm new to both TS and JS. Could you provide me with a sample test code for reference? If it's related to ClickHouse, that would be even better.

fxx-real:
@paveltiunov  Sure, I'd like to, but I'm new to both TS and JS. Could you provide me with a sample test code for reference? If it's related to ClickHouse, that would be even better.

paveltiunov:
@fxx-real https://github.com/cube-js/cube.js/blob/master/packages/cubejs-testing/test/smoke-materialize.test.ts would be a great starting point.

PR #6190, opened 2023-02-22
fix(server-core): Fix dbType for baseDrivers & generic driverFactory

wolfeeeg:
**Check List**

[For example #12]

When we setup the custom driverFactory handler in `cube.js` file we can get many db connections. 
An example: 
```
 driverFactory: ({ securityContext, dataSource }) => {
    if (dataSource === "datasource2" && securityContext.tenant === "local") {
      return new PostgresDriver({
        database: "database2",
      });
    }

    if (
      dataSource === "datasource3" &&
      securityContext.tenant === "clickhouse"
    ) {
      return new ClickhouseDriver({
        host: "",
      });
    }

    throw new Error("Unknown tenant in Security Context");
  },
```

To determine dbType for this case we get dbType from driverClassName

### DIFF
--- a/packages/cubejs-server-core/src/core/OptsHandler.ts
+++ b/packages/cubejs-server-core/src/core/OptsHandler.ts
@@ -247,11 +247,10 @@ export class OptsHandler {
         if (!this.driverFactoryType) {
           val = await driverFactory(ctx);
         }
-        if (
-          this.driverFactoryType === 'BaseDriver' &&
-          process.env.CUBEJS_DB_TYPE
-        ) {
-          type = <DatabaseType>process.env.CUBEJS_DB_TYPE;
+        if (this.driverFactoryType === 'BaseDriver') {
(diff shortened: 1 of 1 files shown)


packages/cubejs-server-core/src/core/OptsHandler.ts
```diff
@@ -247,11 +247,10 @@ export class OptsHandler {
         if (!this.driverFactoryType) {
           val = await driverFactory(ctx);
         }
-        if (
-          this.driverFactoryType === 'BaseDriver' &&
-          process.env.CUBEJS_DB_TYPE
-        ) {
-          type = <DatabaseType>process.env.CUBEJS_DB_TYPE;
+        if (this.driverFactoryType === 'BaseDriver') {
+          type = <DatabaseType>(process.env.CUBEJS_DB_TYPE || (await driverFactory(ctx)).constructor.name
```
wolfeeeg, line 251: Maybe we should add a result type check for values in DatabaseType

PR #6277, opened 2023-03-10
Implement Druid support for countDistinctApprox

tquachbot:
**Check List**

[For example #12]

[Description goes here]

### DIFF
--- a/packages/cubejs-druid-driver/src/DruidQuery.ts
+++ b/packages/cubejs-druid-driver/src/DruidQuery.ts
@@ -52,4 +52,8 @@ export class DruidQuery extends BaseQuery {
   public nowTimestampSql(): string {
     return `CURRENT_TIMESTAMP`;
   }
+
+  public countDistinctApprox(sql) {
+    return `APPROX_COUNT_DISTINCT_DS_THETA(${sql})`
+  }
 }
--- a/packages/cubejs-druid-driver/test/druid-query.test.ts
+++ b/packages/cubejs-druid-driver/test/druid-query.test.ts
@@ -19,6 +19,9 @@ describe('DruidQuery', () => {
       measures: {
         count: {
           type: 'count'
+        },
+        unique: {
+          type: 'countDistinctApprox'
         }
       },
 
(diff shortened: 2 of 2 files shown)


paveltiunov:
@tquachbot Thanks for contributing! Could you please fix the linter errors?

packages/cubejs-druid-driver/src/DruidQuery.ts
```diff
@@ -52,4 +52,8 @@ export class DruidQuery extends BaseQuery {
   public nowTimestampSql(): string {
     return `CURRENT_TIMESTAMP`;
   }
+
+  public countDistinctApprox(sql) {
+    return `APPROX_COUNT_DISTINCT_DS_THETA(${sql})`
```
ovr, line 57:
```suggestion
    return `APPROX_COUNT_DISTINCT_DS_THETA(${sql})`;
```

packages/cubejs-druid-driver/src/DruidQuery.ts
```diff
@@ -52,4 +52,8 @@ export class DruidQuery extends BaseQuery {
   public nowTimestampSql(): string {
     return `CURRENT_TIMESTAMP`;
   }
+
+  public countDistinctApprox(sql) {
+    return `APPROX_COUNT_DISTINCT_DS_THETA(${sql})`;
```
ovr, line 57: Interesting to see that Druid supports 3 different implementations of HyperLogLog

-`APPROX_COUNT_DISTINCT` - Counts distinct values of a regular column or a prebuilt sketch column.
-`APPROX_COUNT_DISTINCT_DS_HLL` - Counts distinct values of an HLL sketch column or a regular column.
-`APPROX_COUNT_DISTINCT_DS_THETA` - Counts distinct values of a Theta sketch column or a regular column.

ovr: https://druid.apache.org/docs/latest/querying/sql-aggregations.html

> The specific algorithm depends on the value of [druid.sql.approxCountDistinct.function](https://druid.apache.org/docs/latest/configuration/index.html#sql). By default, this is APPROX_COUNT_DISTINCT_BUILTIN. If the [DataSketches extension](https://druid.apache.org/docs/latest/development/extensions-core/datasketches-extension.html) is loaded, you can set it to APPROX_COUNT_DISTINCT_DS_HLL or APPROX_COUNT_DISTINCT_DS_THETA.

It looks like we need to use `APPROX_COUNT_DISTINCT`.

@tquachbot Do you have any reason why you specified directly `APPROX_COUNT_DISTINCT_DS_THETA` instead of using config?

mormigil: I work with Tim, but we personally use DS_THETA for increased accuracy of theta intersect operations. If there's an easy way to choose the approx distinct method through config that'd be ideal.

paveltiunov:
@tquachbot We need also to fix test here.

packages/cubejs-druid-driver/src/DruidQuery.ts
mormigil: Specifically how would we use config to specify different types of count distinct? Is it possible to change the count distinct definitions through some schema flag instead of having a single hardcoded option?

PR #6293, opened 2023-03-14
feat(cubestore): RocksStore - panic protection for RW loop

ovr:
Hello!

Issue: Single panic in the rw loop can break the whole processing

## Example:

image

```
mysql> cache get “1”;
ERROR 1815 (HY000): channel closed

# another operation without panic will not be processed
mysql> cache set “1" “1”;
ERROR 1815 (HY000): channel closed
```

Thanks

### DIFF
--- a/rust/cubestore/cubestore/src/metastore/rocks_store.rs
+++ b/rust/cubestore/cubestore/src/metastore/rocks_store.rs
@@ -563,8 +563,16 @@ impl RocksStore {
 
         let join_handle = cube_ext::spawn_blocking(move || loop {
             if let Some(fun) = rw_loop_rx.blocking_recv() {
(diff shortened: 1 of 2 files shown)


rust/cubestore/cubestore/src/metastore/rocks_store.rs
```diff
@@ -888,10 +908,17 @@ impl RocksStore {
             Ok(())
         }));
         if let Err(e) = res.await {
-            log::error!("Error during read write loop send: {}", e);
+            log::error!("Error during scheduling read task in loop: {}", e);
+
+            return Err(CubeError::internal(format!(
```
ovr, line 913: fast exit

waralexrom (approved):

PR #6339, opened 2023-03-23
feat(cubestore): Use new XXH3 checksum alg for RocksDB

ovr:
Hello!

Thanks


rust/cubestore/cubestore/src/metastore/rocks_store.rs
```diff
@@ -523,15 +523,15 @@ impl RocksStoreConfig {
         Self {
             // Supported since RocksDB 6.27
             checksum_type: RocksStoreChecksumType::XXH3,
-            cache_capacity: 1024 * 8,
+            cache_capacity: 8 * 1024 * 1024,
```
ovr, line 526: image

copy/pasted from test :(

PR #6367, opened 2023-03-29
feat(schema-compiler): Support new aliases for join. relationship

ovr:
Hello!

Thanks

### DIFF
--- a/packages/cubejs-schema-compiler/src/compiler/CubeValidator.js
+++ b/packages/cubejs-schema-compiler/src/compiler/CubeValidator.js
@@ -471,7 +471,11 @@ const baseSchema = {
   public: Joi.boolean().strict(),
   joins: Joi.object().pattern(identifierRegex, Joi.object().keys({
     sql: Joi.func().required(),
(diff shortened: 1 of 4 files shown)


packages/cubejs-schema-compiler/src/compiler/CubeValidator.js
```diff
@@ -470,7 +470,7 @@ const baseSchema = {
   public: Joi.boolean().strict(),
   joins: Joi.object().pattern(identifierRegex, Joi.object().keys({
     sql: Joi.func().required(),
-    relationship: Joi.any().valid('hasMany', 'belongsTo', 'hasOne').required()
+    relationship: Joi.any().valid('belongsTo', 'belongs_to', 'many_to_one', 'hasMany', 'has_many', 'one_to_many', 'hasOne', 'has_one', 'one_to_one').required()
```
keydunov, line 473: We should support new relationship names in camel case too.

ovr: Ok, no problem 👍

PR #6396, opened 2023-04-05
feat(schema-compiler): Support snake_case aliases for fields

ovr:
Hello!

Thanks


splatcollision:
I was wondering why some meta object property names in my cube schemas changed all of the sudden, this PR is the culprit. Had to debug front end code for a bit before realizing something had changed in the latest cube code.

I had meta objects in my schemas with snake case property names, a python habit I suppose, but IMO cube shouldn't care about what's in those objects. User expectation is that schema item `meta` fields should remain unchanged. 

https://cube.dev/docs/schema/reference/dimensions#parameters-meta

"Custom metadata. Can be used to pass any information to the frontend."

Cheers and thanks for the great product though!

ovr:
@splatcollision 

I am sorry to hear, I've prepared PR with a fix for that - https://github.com/cube-js/cube/pull/6444

Good catch 👍

splatcollision:
OK Thanks, reverting my front-end changes, thanks!

PR #6420, opened 2023-04-12
docs(workspace): update Query Inspection page

hassankhan:


igorlukanin:
@hassankhan I've elaborated in #6442, so I'll be high-level here:
* let's try to not only describe the UI but also provide actionable advice on using the data, e.g., "check that all queries are sub-second," "check that top queries are accelerated," "check there are no failing queries"; also, let's provide advice on what is "normal" or "recommended" to be seen there and what is not, e.g., query latency; it really looks like we're giving a tool but not giving advice on what's the best use for that tool, so let's interview Pavel, Paco, etc. to add that advice
* let's cross-link with Cube Core sections

docs/content/Workspace/Inspecting-Queries.mdx
```diff
@@ -5,55 +5,155 @@ category: Workspace
...
+
+If no pre-aggregations were used for this query, you should see the following
+screen:
+
+<Screenshot src="https://ucarecdn.com/b5cadb89-fbda-404a-806c-da99dc046845/" />
+
+Clicking <Btn>Accelerate</Btn> takes you to the Rollup Designer, where you can
+add a pre-aggregation to accelerate similar queries in the future.
+
+### <--{"id" : "API Queries"}--> Queue Graph
```
igorlukanin, line 121: Super unclear when to use this one and when to use the next one.

docs/content/Workspace/Inspecting-Queries.mdx
```diff
@@ -5,55 +5,155 @@ category: Workspace
...
+### <--{"id" : "API Queries"}--> Errors
+
+If the query failed, the <Btn>Errors</Btn> tab will show you the error message
+and stacktrace:
+
+<Screenshot src="https://ucarecdn.com/858633a0-4ee3-4dec-adb6-1d91ca30919f/" />
+
+### <--{"id" : "API Queries"}--> SQL
+
+The <Btn>SQL</Btn> tab shows the SQL query sent by Cube to your data source:
```
igorlukanin, line 102: "...or Cube Store if the query was accelerated with a pre-aggregation", right?

docs/content/Workspace/Inspecting-Queries.mdx
```diff
@@ -5,55 +5,155 @@ category: Workspace
...
+  highlight="inset(1% 2% 93% 68% round 10px)"
+  src="https://ucarecdn.com/e3c00115-49f9-48a6-8f2b-949c53523c9f/"
+/>
+
+To drill down on a specific query, click it to see more information.
+
+### <--{"id" : "API Queries"}--> Query
+
+The <Btn>Query</Btn> tab shows the raw JSON query sent to the Cube Cloud
+deployment.
```
igorlukanin, line 89: Does it also show a JSON query for a request through the SQL API? 🤨

docs/content/Workspace/Inspecting-Queries.mdx
```diff
@@ -5,55 +5,155 @@ category: Workspace
...
+
+<Screenshot src="https://ucarecdn.com/3c0490be-757d-46cd-ba2d-d7c93a7acf4e/" />
```
igorlukanin, line 91: I wonder what "Client: API" means and what are other possible values. Maybe not worth adding to docs but still interested.

docs/content/Workspace/Inspecting-Queries.mdx
```diff
@@ -5,55 +5,155 @@ category: Workspace
...
 
-You can inspect Cube queries in a web-based interface and improve their performance with pre-aggregations in Cube Cloud.
+The Queries page in Cube Cloud is a one-stop shop for all performance and
+diagnostic information about queries issued for a deployment. It shows you the
+most recent queries, the slowest queries, and the queries that are not using
+pre-aggregations, along with handy visualizations to help you understand where
+you might be losing performance.
 
-<SuccessBox>
+<Screenshot src="https://ucarecdn.com/44bb6d8d-5943-4978-8b15-73e1e43fd988/" />
```
igorlukanin, line 14: Should we say that the data is updated live? (It is, right?)

docs/content/Workspace/Inspecting-Queries.mdx
hassankhan: Interestingly, I'm not seeing any new items in the Queries page when making SQL API requests, could be a bug

hassankhan: It seems like SQL API queries go directly to the "Database" queries tab instead of showing up in the "API" section 😕

hassankhan: Confirmed with @pacofvf, SQL API queries show up as a JSON query too

igorlukanin (approved):

PR #6442, opened 2023-04-15
docs(workspace): document the "Pre-Aggregations" page in Cube Cloud

hassankhan:


docs/content/Workspace/Inspecting-Pre-aggregations.mdx
```diff
@@ -0,0 +1,83 @@
...
+title: Pre-aggregation inspection
+permalink: /cloud/inspecting-pre-aggregations
+category: Workspace
+menuOrder: 5
+---
+
+The Pre-Aggregations page in Cube Cloud allows you to inspect all
+pre-aggregations in your deployment at a glance. You can see which
+pre-aggregations are accelerating queries, which are not, and which are not
+being refreshed, along with the last 24 hours of build history.
```
igorlukanin, line 11: Should we have links from "pre-aggregations" and maybe "refreshed" so the Cube Cloud features are interlinked with relevant Cube Core features?

docs/content/Workspace/Inspecting-Pre-aggregations.mdx
```diff
@@ -0,0 +1,83 @@
...
+### <--{"id" : "Pre-aggregations"}--> Rebuilding
+
+If the pre-aggregation requires rebuilding (usually due to the underlying data
+changing in a significant way), click <Btn>Build All</Btn> to rebuild all
+partitions of the pre-aggregation:
+
+<Screenshot
+  highlight="inset(38% 1% 57% 91% round 10px)"
+  src="https://ucarecdn.com/d65d5910-3561-4101-9317-2c9f8366694e/"
+/>
```
igorlukanin, line 70: @hassankhan Wow, I see what you did here! Super cool! 😍

What if we, subjectively, simplify the configuration and, also subjectively, make the result more appealing by using a radial gradient (https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Images/Using_CSS_gradients#using_radial_gradients) for the highlight, providing just the coordinates of the center of the highlight, and its radius? I think it might make the screenshots more appealing but not sure.

docs/content/Workspace/Inspecting-Pre-aggregations.mdx
```diff
@@ -0,0 +1,83 @@
...
+<Screenshot src="https://ucarecdn.com/1dcef348-af33-434a-9ef7-7a4c5f30c26c/" />
+
+### <--{"id" : "Pre-aggregations"}--> Indexes
+
+The <Btn>Indexes</Btn> tab displays any configured indexes for this
+pre-aggregation:
+
+<Screenshot src="https://ucarecdn.com/f7e306e2-4245-46b4-8785-3e7fc1be518e/" />
+
+### <--{"id" : "Pre-aggregations"}--> Rebuilding
```
igorlukanin, line 61: If I'm correct, there's also a way to rebuild an individual partition (when there are many). Do you think we should specify that too? I didn't find an example deployment to check, sorry.

docs/content/Workspace/Inspecting-Pre-aggregations.mdx
```diff
@@ -0,0 +1,83 @@
...
+
+<Screenshot src="https://ucarecdn.com/eb10de99-4ff6-4a9f-abd5-736971497583/" />
+
+## Exploring pre-aggregations
+
+You can switch between [pre-aggregations](#pre-aggregations) and
+[build history](#build-history) by switching between the&nbsp;
+&ZeroWidthSpace;<Btn>Pre-Aggregations</Btn> and <Btn>Build History</Btn> tabs.
+
+## Pre-aggregations
```
igorlukanin, line 21: I don't think that docs that describe the UI are very useful; they don't add much to what's already evident at the glance. Here's what I think we can add:
* provide advice for what should be the ballpark for the partition build duration and partition size; let's ask Pavel for advice and bake this into docs (the correct way to do that will be to add to the "pre-aggs" section and then cross-link from the "workspace / pre-aggs" page, right?)

docs/content/Workspace/Inspecting-Pre-aggregations.mdx
```diff
@@ -0,0 +1,83 @@
...
+
+The <Btn>Used By</Btn> tab shows recent queries that used this pre-aggregation
+for acceleration:
+
+<Screenshot src="https://ucarecdn.com/1dcef348-af33-434a-9ef7-7a4c5f30c26c/" />
+
+### <--{"id" : "Pre-aggregations"}--> Indexes
+
+The <Btn>Indexes</Btn> tab displays any configured indexes for this
+pre-aggregation:
```
igorlukanin, line 57: A cross-link to the Cube Core page/section from "indexes"? Same for subsections above.

igorlukanin (approved):
All in all, looks great to me! Hopefully, we'll be able to provide more "scenario, use case-based" descriptions rather than "interface-based" descriptions.

docs/content/Workspace/Inspecting-Pre-aggregations.mdx
hassankhan: I tried playing around with it, but my CSS-fu is weak. Maybe we can ask @ewgenius to improve this at a later date?

docs/content/Workspace/Inspecting-Pre-aggregations.mdx
igorlukanin: Actually, there's a way!

Screenshot 2023-04-26 at 14 47 06

igorlukanin: Let's have this documented—we're definitely getting questions about that on Slack.

PR #6470, opened 2023-04-21
docs: use "Data Model" instead of "Data Schema"

hassankhan:


keydunov:
@hassankhan do you know why is preview failing? Given the scale of changes, I'd love to check the preview.

PR #6481, opened 2023-04-24
docs: replace usage of `schema/` folder with `model/`

hassankhan:


docs/content/Examples-Tutorials-Recipes/Recipes/entity-attribute-value.mdx
```diff
@@ -247,7 +247,7 @@ from an external source, e.g., the data source. Here's the code from the
 `fetch.js` file that defines the `fetchStatuses` function that would load the
 statuses from the database. Note that it uses the `pg` package (Node.js client
 for Postgres) and reuses the credentials from Cube. Let's keep this file outside
-the `schema` folder:
+the `model/` folder:
```
igorlukanin, line 250:
```suggestion
the `model` folder:
```

igorlukanin:
LGTM! Great job, @hassankhan!

However, may I ask to rebase this PR to the tip of repo? I see that we never actually elaborate on the folder structure and why we recommend to have `cube` and `view` subfolders. I think that can be added under `Data Modeling → Syntax`. I can help with that tomorrow.

hassankhan:
> I see that we never actually elaborate on the folder structure and why we recommend to have cube and view subfolders.

I thought it would be a better fit in a "Best Practices" page

igorlukanin:
> I thought it would be a better fit in a "Best Practices" page

Makes sense! However, since now the product generates these folders, it's part of product now and needs to be documented as such. I'm gonna do it right away.

igorlukanin:
I think this is ready to be merged—let's just [decide](https://cubedevinc.slack.com/archives/C04V573TYPN/p1682516404276109) on `schemaPath` vs. `modelPath`.

PR #6492, opened 2023-04-25
docs(deployment): create new page for auto-suspend functionality in Cube Cloud

hassankhan:


docs/content/Workspace/Auto-Suspend.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+<SuccessBox>
+
+Auto-suspend is available in Cube Cloud on
+[Starter and above tiers](https://cube.dev/pricing).
+
+</SuccessBox>
+
+Cube Cloud can automatically suspend infrastructure when not in use to prevent
+CCU consumption when resources are not being actively used. This is useful for
```
igorlukanin, line 16: I don't believe it's clear from the get-go what "CCU consumption" is. I think it might be better to be more high-level here and say something less specific than "CCU". Maybe "resource consumption" (then "resources are not being used" would use "infrastructure"), "overspend", "unnecessary quota use", etc.

igorlukanin: Also, having a link to the pricing page would help, right? The one where CCUs are defined.

docs/content/Workspace/Auto-Suspend.mdx
```diff
@@ -0,0 +1,62 @@
...
+deployments that are not used 24/7, such as staging deployments. Auto-suspend
+will hibernate the deployment when no API requests are received after a period
+of time, and automatically resume the deployment when API requests start coming
+in again.
+
+Development Instances are auto-suspended automatically when not in use, whereas
```
igorlukanin, line 22: Can't we link to Development Instances somewhere?

igorlukanin (approved):
@hassankhan Love this PR. Do you think we can also provide better wording in the product? Feel free to propose a change!

docs/content/Workspace/Auto-Suspend.mdx
```diff
@@ -0,0 +1,62 @@
+---
+title: Auto-suspend
+permalink: /workspace/auto-suspend
+category: Workspace
```
igorlukanin, line 4: Do you think this really belongs to Workspace? Sounds like the Deployment stuff to me :-)

hassankhan: I wasn't sure myself TBH, was actually hoping you'd have a suggestion 😀 Will move it to the Deployment category, I agree, its a better fit

keydunov:
This is great! Thank you @hassankhan.

Issue #6197, opened 2023-02-23
Databricks Connector Fails


P2W2:
**Describe the bug**
Hello,
I am in the process of trying out cube.js. I wanted to use the Databricks JDBC connector with Docker.
I tried to follow this [Documentation](https://cube.dev/docs/config/databases/databricks/jdbc).
The Container Starts up but if I try to open the schema (http://localhost:4000/#/schema) I got following failure:

```
Error: ENOENT: no such file or directory, open 'package.json'
    at DatabricksDriver.testConnection (/cube/node_modules/@cubejs-backend/jdbc-driver/src/JDBCDriver.ts:188:13)
    at CubejsServerCore.getDriver (/cube/node_modules/@cubejs-backend/server-core/src/core/server.ts:782:7)
    at /cube/node_modules/@cubejs-backend/server-core/src/core/DevServer.ts:108:22
    at /cube/node_modules/@cubejs-backend/server-core/src/core/DevServer.ts:78:9
```

**To Reproduce**
1. Follow the [Documentation](https://cube.dev/docs/config/databases/databricks/jdbc) to build the Docker Image
2. docker run -p 4000:4000 -p 3000:3000 --env-file=.env -v ${PWD}:/cube/conf -e CUBEJS_DEV_MODE=true cubejs-jdk:latest

```
Warning. There is no cube.js file. Continue with environment variables
🔥 Cube Store (0.31.64) is assigned to 3030 port.
Warning. Option apiSecret is required in dev mode. Cube.js has generated it as 5bc38414df47015c0432991a0b8f2cbe
🔓 Authentication checks are disabled in developer mode. Please use NODE_ENV=production to enable it.
🦅 Dev environment available at http://localhost:4000
🚀 Cube API server (0.31.64) is listening on 4000
2023-02-23 11:26:41,803 INFO  [cubestored] <pid:29> Cube Store version 0.31.64
2023-02-23 11:26:41,890 INFO  [cubestore::http::status] <pid:29> Serving status probes at 0.0.0.0:3031
2023-02-23 11:26:41,902 INFO  [cubestore::metastore::rocks_fs] <pid:29> Using existing metastore in /cube/conf/.cubestore/data/metastore
2023-02-23 11:26:42,315 INFO  [cubestore::mysql] <pid:29> MySQL port open on 0.0.0.0:13306
2023-02-23 11:26:42,321 INFO  [cubestore::http] <pid:29> Http Server is listening on 0.0.0.0:3030
[...]

PR #6521, opened 2023-05-01
Remove package.json dependency from DatabricksDriver.ts

carlagouveia:
This aims to solve #6197. The `package.json` file is removed during image build, so the version is never actually used. However, in case there is an exception while connecting to Databricks, the actual exception will be hidden with the error related to this instead. This makes debugging Databricks connection issues impossible.


paveltiunov:
@carlagouveia Looks good! Could we just put `CubeDev_Cube` in the user agent value instead?

carlagouveia:
> @carlagouveia Looks good! Could we just put `CubeDev_Cube` in the user agent value instead?

Done @paveltiunov.

PR #6531, opened 2023-05-02
fix (firebolt-driver) cast boolean params

Jeseeq:
**Check List**

Cube.js by default generate sql that is not supported by firebolt yet. (compare boolean column with "true", "false" string values)
This pr will fix issue with queries such as `select * from a_table where a_boolean = "true"`  by casting boolean parameters using `cast` syntax


ovr:
@Jeseeq 

Thank you for your contribution!

PR #6606, opened 2023-05-16
docs: A one-liner about how Cube is packaged

igorlukanin:
We can add a link to "Architecture" later when we have it.

### DIFF
--- a/docs/content/Getting-Started/Overview.mdx
+++ b/docs/content/Getting-Started/Overview.mdx
@@ -7,6 +7,8 @@ menuOrder: 1
 isDisableFeedbackBlock: false
 ---
 
+Cube is a cloud-native application, designed to run as a set of containers.
+
 You can get started with Cube in one of two ways.
 
 We recommend using [Cube Cloud][cube-cloud], our managed platform for Cube,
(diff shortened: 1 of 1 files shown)


docs/content/Getting-Started/Overview.mdx
```diff
@@ -7,6 +7,8 @@ menuOrder: 1
 isDisableFeedbackBlock: false
 ---
 
+Cube is a cloud-native application, designed to run as a set of containers.
```
hassankhan, line 10:
```suggestion
Cube is a cloud-native application, packaged and distributed as Docker images and designed to run in a containerized environment.
```

WDYT?

hassankhan: Could maybe add a link to https://docs.docker.com/get-started/overview/#images

igorlukanin: Sounds great!

docs/content/Getting-Started/Overview.mdx
```diff
@@ -7,6 +7,8 @@ menuOrder: 1
 isDisableFeedbackBlock: false
 ---
 
+Cube is a cloud-native application, packaged and distributed as Docker images and designed to run in a containerized environment.
```
hassankhan, line 10: Could maybe also add a link to https://docs.docker.com/get-started/overview/#images?

igorlukanin: I suggest we omit this link here but maybe add it later to the Architecture page. Would be strange if we link to Docker's website first thing in Getting started.

hassankhan: Fair enough 😀

hassankhan (approved):

PR #6607, opened 2023-05-16
docs: Add JavaScript part into the style guide

igorlukanin:


docs/content/Guides/Style-Guide.mdx
```diff
@@ -255,6 +255,51 @@ cubes:
         sub_query: true
 ```
 
+## JavaScript style guide
+
+* Indent with 2 spaces.
+* Don't use trailing semicolons.
+* Don't use trailing commas after last elements of arrays and objects.
```
igorlukanin, line 262: This is because data model should be optimized for readability as it is, not for readability in Git commits.

PR #6612, opened 2023-05-16
docs(reference): move property syntax under their respective headings

hassankhan:


hassankhan:
Will revisit in the future

PR #6622, opened 2023-05-17
force ordering by column name

jtcies:
**Check List**

[6023] - When querying `information_schema.columns` Redshift can sometimes return columns not in their ordinal position. When unloading a partitioned pre-aggregate to the export bucket, column orders may differ between partitions, causing any queries that use mismatched partitions to fail. 

Added an `order by` statement to `tableColumnTypes` query forcing columns to alway be returned in alphabetical order.


paveltiunov:
@jtcies Thanks for contributing to this! As this seems to be a fix for another fix we'd need a test that reproduces exactly this situation. You can add it to `smoke-redshift.test.ts`.

paveltiunov:
@jtcies Most likely, ordering by name will break a lot of things. It should be ordered by `ordinal_position` instead.

PR #6645, opened 2023-05-23
docs(deployment): clarify querying limits

hassankhan:


igorlukanin:
@hassankhan We have a dedicated page for limits (https://cube.dev/docs/cloud/limits). WDYT about linking to that page instead of repeating limit thresholds on several docs pages? I think that keeping specific thresholds to a single page would reduce docs maintenance burden and make less room for errors.

PR #6672, opened 2023-05-31
chore(native): Document supported platforms for python builds

ovr:
Hello!

Thanks


packages/cubejs-backend-native/README.md
```diff
@@ -16,6 +16,29 @@ Native module for Cube.js (binding to Rust codebase).
 
 [Learn more](https://github.com/cube-js/cube.js#getting-started)
 
+There are two different types of builds: with python and fallback.
```
hassankhan, line 19:
```suggestion
There are two different types of builds: with Python, and a fallback.
```

PR #6684, opened 2023-06-03
Fix incorrect index in movePivotItem

telunc:
**Check List**

Incorrect index in movePivotItem.
When moving a dimension from x to y at the second last index, dimension is inserted at third last index.
For example,
|  x | y | 
| ------------ | -------------| 
|dimension2 | dimension1 |
|| measures |

After dragging dimension2 to y index 1 (below dimension1 and above measure),
|  x | y | 
| ------------ | -------------| 
|| dimension2|
|| dimension1|
|| measures |

The expected behavior should be
|  x | y | 
| ------------ | -------------| 
|| dimension1|
|| dimension2|
|| measures |

### DIFF
--- a/packages/cubejs-client-core/src/utils.js
+++ b/packages/cubejs-client-core/src/utils.js
@@ -288,6 +288,7 @@ export function movePivotItem(
   if (id === 'measures') {
     destinationIndex = lastIndex + 1;
   } else if (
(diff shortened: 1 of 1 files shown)


paveltiunov:
@telunc Thanks for contributing! Could you please add a test for that?

PR #6704, opened 2023-06-08
feat(schema): Initial support for jinja templates

ovr:
Hello!

Thanks


packages/cubejs-schema-compiler/src/compiler/DataSchemaCompiler.js
```diff
@@ -48,25 +53,39 @@ export class DataSchemaCompiler {
     }
   }
 
+  /**
+   * @protected
+   */
+  async doCompile() {
+    if (NATIVE_IS_SUPPORTED && !NATIVE_IS_FALLBACK_BUILD) {
```
eugene1g, line 60: CubeJS does not support `native` on Apple M1/M2, but `isNativeSupported()` incorrectly returns `true` on those platforms (because there is no check for `darwin aarch64`). On these platforms, `initJinjaEngine` is executed and throws an exception.

ovr: Great catch! Thanks

I've prepared PR to detect it - https://github.com/cube-js/cube/pull/6737

It will be merged & released soon.

PR #6782, opened 2023-07-04
DuckDB dates interpreted as strings instead of timestamp

ygg:
**Check List**

CubeJS was treating DuckDB TIMESTAMPTZ data type as a string rather than a timestamp. This created problems with pre-aggregation processes, especially when trying to group data by units of time like months, weeks, etc.


paveltiunov:
@ygg Thanks for contributing! Could you please add an integration or smoke test for that?

KSDaemon:
Hi @ygg Can you elaborate when/where cube treats `date_trunc`'s as strings rather then timestamps?

PR #6886, opened 2023-07-14
feat: GraphQL GROUP BY Resolver

peterklingelhofer:
**Check List**

Related: #3433

I noticed the GraphQL API supports `orderBy`, but there isn't a built-in resolver for `groupBy` currently. Here is an initial attempt at providing at providing the equivalent to the `GROUP BY` in SQL. 

The idea here would be to be able to .groupBy option just as the .orderBy option is already available in the built-in resolver. Then, if a user wanted to group by the timestamp column for example, the data would be returned as follows instead of as a flat array:

{ 2023-07-16:00:00:00: [{ field1: value1, field2: value2}, {field1: value3, field2: value4}], 2023-07-16:01:00:00: [{ field1: value1, field2: value2}, {field1: value3, field2: value4}], ...}

This avoids the need to perform to series or chart pivot operations on the front-end and lets the easier-to-scale back-end deliver data in a format that charting libraries such as @recharts expect.

Tagging @lvauvillier as they did amazing work on the initial implementation (https://github.com/cube-js/cube/issues/3433 / https://github.com/cube-js/cube/pull/3555) if they have time to offer their thoughts on the matter.

### DIFF
--- a/packages/cubejs-api-gateway/src/graphql.ts
+++ b/packages/cubejs-api-gateway/src/graphql.ts
@@ -580,7 +580,7 @@ export function makeSchema(metaConfig: any): GraphQLSchema {
               if (memberType === MemberType.MEASURES) {
                 measures.push(key);
               } else if (memberType === MemberType.DIMENSIONS) {
-                if (groupBy.includes(memberName)) {  // only add the dimension if it's in the groupBy arg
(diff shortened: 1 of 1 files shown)


paveltiunov:
Hey @peterklingelhofer ! Could you please elaborate a little bit more on your use case? Currently, all the dimensions will go into group by as a default behavior.

PR #6942, opened 2023-07-24
docs(product): add Tableau instructions for Semantic Layer Sync

hassankhan:


docs/content/Workspace/Semantic-Layer-Sync.mdx
```diff
@@ -253,6 +253,39 @@ service account with a different email address:
...
+      name: "Tableau Sync",
+      config: {
+        database: "My Database: production-deployment",
+        site: "mytableausite",
+        region: "10ax",
+        user: "john@example.com",
+        password: "******************",
+        // Uncomment the following lines to use a personal access token instead of a username and password
+        personalAccessToken: "test",
+        personalAccessTokenSecret: "",
```
igorlukanin, line 279: Well, technically, the following lines aren't commented so one wouldn't be able to uncomment them. Should we split this code example into two variants?

docs/docs-new/pages/product/workspace/semantic-layer-sync.mdx
```diff
@@ -265,6 +265,45 @@ sending them, or use a service account with a different email address:
 
 <Screenshot src="https://ucarecdn.com/2e7ac466-73a5-4dd6-9c85-ed98d9df8979/" />
 
+### Tableau
```
igorlukanin, line 268: Very curious: why do we need PRs to both versions of docs? Do we still maintain the old one?

igorlukanin, line 268: Since we now have four tool-specific subsections on the page, should we extract them into subpages to prevent this page from being, um, too bloated?

Screenshot 2023-07-25 at 13 44 19

hassankhan: Only doing this temporarily to ensure links are migrated + redirected correctly, need some time to work on a separate PR that removes the old docs completely

hassankhan: Agreed, but that work should be part of a separate task/PR

PR #6965, opened 2023-07-31
fix(redshift-driver): resolves issue where redshift column order can …

jskarda829:
…come back in a different order, causing preaggregations across partitions to encounter a union all error

**Check List**

[#6023] - When querying information_schema.columns Redshift can sometimes return columns not in their ordinal position. When unloading a partitioned pre-aggregate to the export bucket, column orders may differ between partitions, causing any queries that use mismatched partitions to fail.

Updated table column types query in BaseDriver.ts to order by the ordinal_position ensuring that columns always come back in the expected order.

### DIFF
--- a/docs/content/Reference/Configuration/Environment-Variables-Reference.mdx
+++ b/docs/content/Reference/Configuration/Environment-Variables-Reference.mdx
@@ -447,6 +447,23 @@ The timeout value for any queries made to the database by Cube.
 | ---------------------------------------- | ---------------------- | --------------------- |
 | A number in seconds or a duration string | `10m`                  | `10m`                 |
 
+## `CUBEJS_DB_FETCH_COLUMNS_BY_ORDINAL_POSITION`
+
+Force fetching of columns by ordinal positions.  Certain data-providers (Redshift) do not guarantee columns in the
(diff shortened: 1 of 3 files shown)


paveltiunov:
@jskarda829 Thanks for contributing! While this seems to be a correct change in a `BaseDriver` it'd be potentially breaking change. So we should ship it under the env flag and then eventually switch it to true on minor changes.

paveltiunov:
@igorlukanin We'd need to release it under flag before we can do minor.

paveltiunov:
@magno32 Yep. It'll most likely break some of the pre-aggregations, and it's why we'd need a flag. Something like `CUBEJS_FETCH_COLUMNS_BY_ORD_POSITION` would work. It'd be temporary, and we'll set it to true by default eventually.

magno32:
@igorlukanin @paveltiunov 

Since my changes happened down-stream, I'll comment here.  

I added the `env` as requested: `CUBEJS_DB_FETCH_COLUMNS_BY_ORDINAL_POSITION`
This conditionally enables the change made by @jskarda829 

Documentation layout has changed a bit since I last added to it, so I hope I got the docs in the right place.

PR #6979, opened 2023-08-03
docs(product): add API quotas to Limits page

hassankhan:


keydunov:
@hassankhan Can you please add a message that customers can contact us if they need this quota to be increased.

PR #7031, opened 2023-08-14
fix(cubestore): Reduce memory usage while truncating

ovr:
Hello!

Thanks

### DIFF
--- a/rust/cubestore/cubestore/src/metastore/rocks_table.rs
+++ b/rust/cubestore/cubestore/src/metastore/rocks_table.rs
@@ -284,6 +284,7 @@ where
                 if table_id != self.table_id {
                     return None;
                 }
+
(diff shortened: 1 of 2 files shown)


rust/cubestore/cubestore/src/metastore/rocks_table.rs
```diff
@@ -879,6 +880,18 @@ pub trait RocksTable: BaseRocksTable + Debug + Send + Sync {
         Ok(IdRow::new(row_id, new_row))
     }
 
+    fn truncate(&self, batch_pipe: &mut BatchPipe) -> Result<(), CubeError> {
+        let iter = self.table_scan(self.snapshot())?;
```
ovr, line 884: Iterator instead of all rows at once.

waralexrom (approved):

Issue #6312, opened 2023-03-17
rollingWindow not implemented with Cube SQL API: dataRange filter isn't being push down to time dimension


bluestarfish-t1d:
**Describe the bug**
The specified `rollingWindow` for a measure is not implemented if called using Cube SQL API with more than the date range specified in the WHERE clause. 

The same call using Cube JSON API does implement the specified `rollingWindow` even with a value for `filters`.

**measures**
```
measures: {
  amount: {
    sql: `${CUBE}.amount`,
    type: `sum`,
  },
  last6MonthsAmount: {
    sql: `${CUBE}.amount`,
    type: `sum`,
    rollingWindow: {
      trailing: `6 month`,
      offset: `end`,
    },
    title: `Last 6 Months Amount`,
  },
}
```
 
**SQL Query**
```sql
SELECT
  amount,
  last6MonthsAmount
FROM
  AllTransactionsA
WHERE
  proxyKey = 'account'
  AND "balanceDateTime" >= TO_TIMESTAMP('2022-01-01 00:00:00.000000', 'YYYY-MM-DD HH24:MI:SS.US')
  AND "balanceDateTime" < TO_TIMESTAMP('2023-01-01 00:00:00.000000', 'YYYY-MM-DD HH24:MI:SS.US');
```

**SQL Output (Incorrect)**
 amount | last6MonthsAmount 
--|--
 153.24 |   153.24

**JSON Query**
```json
{
  "measures": [
    "AllTransactionsA.amount",
    "AllTransactionsA.last6MonthsAmount",
  ],
  "order": {
    "AllTransactionsA.balanceDateTime": "asc"
  },
  "timeDimensions": [
    {
      "dimension": "AllTransactionsA.balanceDateTime",
      "dateRange": "Last year"
    }
  ],
  "filters": [
    {
      "member": "AllTransactionsA.proxyKey",
      "operator": "equals",
      "values": [
        "account"
      ]
    }
  ]
}
```

**JSON Output (Correct)**
 amount | last6MonthsAmount
-----------------|-------------------
 153.24 |  98.50

**Version:**
0.31.69

paveltiunov:
@bluestarfish-t1d Could you please share `EXPLAIN` SQL API plan for this query?

bluestarfish-t1d:
@paveltiunov see the result from `EXPLAIN` below:

```
 plan_type   |                        plan
---------------+-----------------------------------------------------
 logical_plan  | CubeScan: request={                                +
[...]

Issue #3369, opened 2021-09-02
Additional Filter Operators


dodgex:
**Is your feature request related to a problem? Please describe.**
I'm working on an internal reporting tool and wondered about the filter operators. First of all we would like to have a `inNumberRange`/`notInNumberRange` to allow an easy filter like "invoices with a gross value between x and y" similar to the `inDateRange` variants. Also we would like to have a `afterOrEqualDate` (`>=`) as a variant of `afterDate` (`>`) that includes the given date in the range.

While we could reproduce the number range with the given operators `gt`, `gte`, `lt`, `lte` and the logic `and`/`or`, at the moment there is no easy way to have a `>=` check with dates.

**Describe the solution you'd like**
While reading the documentation about filter operators my first thought was why there are the `beforeDate` and `afterDate` operators at all. As far as i can tell those are just `lt` and `gt` with a fancy name. It would be awesome to just be able to use  `gt`, `gte`, `lt`, `lte` for dates as well as numbers, and having a new `[not]between` operator that replaces the date range operators and can be used with number and time?

paveltiunov:
Hey @dodgex ! I believe it's a great idea. Thanks for posting this one!

PR #7041, opened 2023-08-17
Fix(CubeSQL) Add ability to filter dates inclusive of date being passed in w…

darapuk:
…hen using <= or >=

**Check List**

[[#3369]](https://github.com/cube-js/cube/issues/3369) 
[[#6312]](https://github.com/cube-js/cube/issues/6312)

Example query:
```
EXPLAIN SELECT encounter_start from encounters WHERE encounter_start >= '2018'
```

What cube actually runs:
```
SELECT encounter_start from encounters WHERE encounter_start > '2018'
```

The problem with this is what the graphql endpoint receives, note the afterDate:
```
CubeScan: request={
  "measures": [],
  "dimensions": [
    "encounters.encounter_start"
  ],
  "segments": [],
  "filters": [
    {
      "member": "encounters.encounter_start",
      "operator": "afterDate",
      "values": [
        "2018"
      ]
    }
  ]
}
```

The problem with this is that it's not doing a GreaterThanOrEqual, it's doing a GreaterThan . Which means all SQL queries which rely on comparing a time using >= or <= are incorrect when using the SQL API endpoint.

This also introduces the `AfterOrOnDate` to the GraphQL endpoint so that users have a more accurate way of representing a time window.


rust/cubesql/cubesql/src/compile/mod.rs
```diff
@@ -4048,94 +4072,169 @@ ORDER BY \"COUNT(count)\" DESC"
...
+                        member: Some("KibanaSampleDataEcommerce.order_date".to_string()),
+                        operator: Some("afterOrOnDate".to_string()),
+                        values: Some(vec!["2021-08-31T00:00:00.000Z".to_string()]),
+                        or: None,
+                        and: None,
+                    },
+                    V1LoadRequestQueryFilterItem {
+                        member: Some("KibanaSampleDataEcommerce.order_date".to_string()),
+                        operator: Some("beforeDate".to_string()),
+                        values: Some(vec!["2021-09-06T23:59:59.999Z".to_string()]),
```
darapuk, line 4171: There's some code that is forcing this interval to change and I cannot seem to trace down to see where this is occuring.

paveltiunov: It's a regression. Please see further comments.

rust/cubesql/cubesql/src/compile/rewrite/rules/filters.rs
```diff
@@ -2253,6 +2253,59 @@ impl RewriteRules for FilterRules {
                     "FilterOpOp:and",
                 ),
             ),
+            transforming_rewrite(
```
paveltiunov, line 2256: We shouldn't add new rules here. If two rules like these are introduced previous two should be adjusted to use minus one millisecond logic so all existing tests are still passing.

paveltiunov (changes requested):
Thanks for contributing to it! This is a big one and, and indeed very tricky! Most of the provided changes to tests are regressions. So first of all we should avoid updating existing tests. To fix all of those minus one ms logic should be extracted to matching rule for afterDate/beforeDate during time inDateRange conversion: so if you see afterDate/beforeDate operator timestamp should be adjusted by 1 millisecond during the rewrite transformation phase.

paveltiunov:
Here's where minus one ms logic is implemented for your reference: https://github.com/cube-js/cube/pull/7041/files#diff-bb70f8319a553cbdd6e606a1e7dec4310814ca2199e35e839258a99db1757f4aL2653

igorlukanin:
@darapuk Would you also be open to updating the documentation for APIs you've changing? Thanks in advance! https://cube.dev/docs/product/apis-integrations

rust/cubesql/cubesql/src/compile/rewrite/rules/filters.rs
darapuk: Ahh, I see. To make sure I'm following, my takeaway is that the regressions are coming from the `date_range` filter not matching.  This change only matches when both sides of the SQL is <= and >=. So there would need to be additional filters for <= and > and vice versa. 

One pointer I would really appreciate is how to incorporate the minus_one logic with the filter?

I.e. I want to add the following filter:
```
transforming_rewrite(
                "filter-replacer-in-date-range-inclusive-two",
                filter_op(
                    filter_op_filters(
                        filter_member(
                            "?member",
                            "FilterMemberOp:afterOrOnDate",
                            "?date_range_start",
                        ),
                        filter_member(
                            "?member",
                            "FilterMemberOp:beforeDate",
                            "?date_range_end", <-- Here, I want to call a minus_one transformation
                        ),
                    ),
                    "FilterOpOp:and",
                ),
                filter_member("?member", "FilterMemberOp:inDateRange", "?date_range"),
                self.merge_date_range("?date_range_start", "?date_range_end", "?date_range"),
            ),
```

I'm struggling to figure out what the minus_one transformation should look like here

darapuk: The dumb solution I can think of is to add in a transformation param into `self.merge_date_range("?date_range_start", "?date_range_end", "?date_range")` and do the transformation in merge_date_range.

i.e.
`self.merge_date_range("?date_range_start", "?date_range_end", "?date_range", 'TransformMemberOp:StartMinusOne)`
`self.merge_date_range("?date_range_start", "?date_range_end", "?date_range", 'TransformMemberOp:StartPlusOne)`
`self.merge_date_range("?date_range_start", "?date_range_end", "?date_range", 'TransformMemberOp:EndMinusOne)`
`self.merge_date_range("?date_range_start", "?date_range_end", "?date_range", 'TransformMemberOp:EndPlusOne)`

paveltiunov: @darapuk I'd actually take a slightly bit different approach. In fact `filter-replacer-in-date-range-inclusive` rule should become smart enough to match either `afterDate` or `afterOrOnDate` and `beforeDate` or `beforeOrOnDate` and coerce it to `inDateRange` with minus/plus ms logic. To avoid cartesian products of all possible inversions it should also understand if `afterDate` comes first or last. So it'll be the one rule to rule them all.

darapuk: I logically understand where you're going, but need that hint for how to extract the "operation" from the SQL. In `filter-replacer-in-date-range-inclusive` the filter_member seems to match exactly on the filter type:
```
let op = match_data_node!(node_by_id, params[1], FilterMemberOp);
```

What I'm not following is how to extract which type of operator is being used then pass it into the transformer.

I.e. 
What I want to be able to write is something like:
```
            transforming_rewrite(
                "filter-replacer-in-date-range-inclusive",
                filter_op(
                    filter_op_filters(
                        filter_member(
                            "?member",
                            "?left_operator",
                            "?date_range_left",
                        ),
                        filter_member(
                            "?member",
                            "?right_operator",
                            "?date_range_right",
                        ),
                    ),
                    "FilterOpOp:and",
                ),
                filter_member("?member", "FilterMemberOp:inDateRange", "?date_range"),
                self.merge_date_range("?date_range_left", "?date_range_right", "?date_range", "?right_operator", "left_operator"),
```

is that at all close, or am I totally misunderstanding how these filters are working?

darapuk:
Thanks for the review @paveltiunov! I reverted all the tests with regressions and just updated the ones which are incorrect. I realize the code is horribly inefficient and not very DRY. But at least now the code is in a "working" state.

I'm not familiar enough with how the filters work to take it the last mile unfortunately. Some examples or direction would help me push it over the line, or if anyone else wants to contribute to the PR, open to that as well!

paveltiunov:
@darapuk Yep. You are almost there! We just need to get rid of this duplicated code. You can parameterize `FilterMemberOp` operators using variables like `?date_range_start_op` and `?date_range_end_op`.

darapuk:
Thanks for that tip @paveltiunov! I believe the code is now in a bit cleaner state for a re-review.

paveltiunov (changes requested):
We're at the last mile! We just need to fix matching issues

rust/cubesql/cubesql/src/compile/rewrite/rules/filters.rs
```diff
@@ -3810,27 +3790,92 @@ impl FilterRules {
...
+                                && !valid_filters.contains(date_range_end_op)
+                            {
+                                return false;
+                            }
+                            let mut result = Vec::new();
+                            let resolved_start_date =
+                                resolve_time_delta(&date_range_start[0], date_range_start_op);
+                            let resolved_end_date =
+                                resolve_time_delta(&date_range_end[0], date_range_end_op);
+                            if resolved_start_date <= resolved_end_date {
```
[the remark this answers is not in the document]
darapuk: OOF. Good catch! Updated the code to handle the left and right + the overlapping operator case.

rust/cubesql/cubesql/src/compile/mod.rs
```diff
@@ -17873,26 +17873,56 @@ ORDER BY \"COUNT(count)\" DESC"
     }
 
     #[tokio::test]
-    async fn test_thoughtspot_filter_date_trunc_column_with_literal() {
+    async fn test_thoughtspot_filter_date_trunc_column_with_literal2() {
```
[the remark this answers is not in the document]
darapuk: Nope, reverted

darapuk:
Hi @paveltiunov, 

when you get a sec, would love to get another review on this and push it through! Thanks!

paveltiunov:
@darapuk Looks good! Thanks for contributing this one!

siamak-haschemi:
Thank you also from my team @darapuk. We now can remove the millisecond fix during our SQL generation.

Issue #6893, opened 2023-07-18
Add WITH FILL modifier to Clickhouse driver


mrenigma:
**Is your feature request related to a problem? Please describe.**
When running date ranged queries it is often advantageous to provide data for the entire date range.

Cube already does this partially with [`fillMissingDates`](https://cube.dev/docs/@cubejs-client-core) but this only fills between two data points. It doesn't span the entire date range provided unless you have data at the very beginning and end of the date range.

This can cause confusion and misrepresentation when using graphs because the time axis doesn't make use of the full date range selected:

![image](https://github.com/cube-js/cube/assets/1901969/6cf5721a-9e6c-4779-93b0-5fe6b42fea54)

Clickhouse provides a [`WITH FILL` modifier](https://clickhouse.com/docs/en/sql-reference/statements/select/order-by#order-by-expr-with-fill-modifier) which provides this functionality out-of-the-box, but the current driver doesn't support applying the modifier.

**Describe the solution you'd like**
To support filling the entire date range provided, I would propose updating `fillMissingDates` to span the entire date range. This would most likely entail changes to https://github.com/cube-js/cube/blob/master/packages/cubejs-schema-compiler/src/adapter/ClickHouseQuery.js or https://github.com/cube-js/cube/blob/master/packages/cubejs-clickhouse-driver/src/ClickHouseDriver.ts.

This change would be Clickhouse driver specific as I'm not sure if all the supported drivers have equivalent `WITH FILL` modifiers. Using `fillMissingDates` may be the most appropriate as its existing functionality, only being augmented for Clickhouse. Documentation would be need to be updated to describe the enhancement though.

Alternatively, to maintain backwards compatibility, a new pivot option could be used for the same purpose. Although, I'm not sure if having a pivot option for one driver is a viable solution.

I would be more than happy to contribute a PR if we can agree on an acceptable path forward.

**Describe alternatives you've considered**
We have written a [transformer layer](https://gitlab.com/gitlab-org/gitlab/-/issues/415687) in Ruby on Rails which takes the data returned by Cube and fills in the missing dates. However, this isn't really scalable as it's all done in memory whilst Clickhouse is designed for this kind of work. Our transformer also doesn't support the more complex query options provided by Cube.

**Additional context**
https://gitlab.com/gitlab-org/gitlab/-/issues/417231

mrenigma:
[...]

Issue #3799, opened 2021-12-10
Post-process result-set result


louis-vincent1:
**Is your feature request related to a problem? Please describe.**
This issue has been opened in the past (Post processing results on server side #1286)
But the initial issuer rapidly changed is approach.

As for me, as a backend developer I would like to enhance/enrich ResultSet return by CubeJS (or by the cache).

I need to do more complex linear algebra operation over my ResultSet which requires math library, but CubeJS forces me to complexify my architecture by either:

- moving complex algorithm on the frontend side
- relying on other service that frontend call after cubejs returned an unusable resultset, which complexify my developement setup when testing locally or doing automatic testing. We try to enforce a "dumb frontend" philosophy where complex routine and processing should be handle on the backend to avoid overwhelming the frontend since frontend codebase tends to grow a lot faster and is much harder to test (in our case).

**Describe the solution you'd like**

IMO, the simplest and most flexible solution would simply be a hook in the `CreateOptions` interface called `resultSetRewrite` that could have the following signature:

```
type ResultSet = any
type ResultSetRewriteFn = (query: Query, context: RequestContext, resultSet: ResultSet) => Promise<ResultSet>;
```

The `query: Query` is to allow the developer to have conditional ResultSet rewrite based on the input query
The `context: RequestContext` is additional context information that could be use for conditional rewrite and specific use-case.
The `resultSet: ResultSet` is of course the resultset returned by either Cubejs or Redis (cache), i.e. user should not worry about cache and should recompute it each time.

Furthermore, there could be tricky situation about caching whereas a user may want transparent caching or not at all. Integrating with cache could be done in another iteration, but one way to make caching integration explicit is by providing an additional function
in the `CreateOption` called `cachedResultSetRewrites: ResultSetRewriteFn`.

**Describe alternatives you've considered**

- Coding in PLV8 or PLPerl in Postgres, but my database is hosted in AWS so a lot of features are unavailable due to security issues and lack of superuser permissions. Moreover, it forces you to code in a procedural way that may leak cursor and plan, crashing runtime may cause database crashing also...
- Calling another service after the frontend received the result set.
- Putting the complex algorithm on the frontend side.
- Trying to intercept call the cubejs to hack my way in rewriting the resultset myself.

**Additional context**

ivan-vdovin:
@louis-vincent1 
Thank you for your suggestion! 
[...]

PR #7116, opened 2023-09-11
feat(@cubejs-backend/schema-compiler): Add WITH FILL to Clickhouse query adapter

mrenigma:
**Check List**

#6893

In this PR, a new env var named `CUBEJS_DB_CLICKHOUSE_WITHFILL` was added. This env var toggles the ability to use the [`WITH FILL` modifier](https://clickhouse.com/docs/en/sql-reference/statements/select/order-by#order-by-expr-with-fill-modifier) for Clickhouse queries.

The `WITH FILL` modifier will fill any dates within the defined date range, which don't have any data, with `0` values. Doing this means that when you receive the data, you receive the full date range provided, not just the data stored in the DB. The benefit for this is that you can show a true x or y-axis spanning the full date range provided which makes charts far less likely to misrepresent the data.

Setting the env var to `false` or not setting it at all will mean that there are no changes to the current Clickhouse behaviours. This is an opt-in feature.

Also, if a date range isn't provided, the `WITH FILL` modifier will be skipped as it requires a `FROM` and `TO` to function.

**Known possible issues**

Using the `WITH FILL` modifier will result in Clickhouse returning more rows than the normal query. This may result in two potential problems (hence the opt-in):

1. An increase in query times or performance degradation depending upon the Clickhouse server setup and Cube resources.
2. If a `limit` is set, the user needs to be aware that using `WITH FILL` may require that `limit` to be increased depending upon the date range/granularity being queried. One approach we're looking at internally is to dynamically adjust the `limit`  we send to Cube by the number of days/granularity based upon the date range used. We opted to not include it in this PR as it's an implementation detail.

### DIFF
--- a/docs/content/Configuration/Databases/ClickHouse.mdx
+++ b/docs/content/Configuration/Databases/ClickHouse.mdx
@@ -33,6 +33,7 @@ CUBEJS_DB_PASS=**********
 | `CUBEJS_DB_USER`                | The username used to connect to the database                                        | A valid database username |    ✅    |                                    ✅                                    |
 | `CUBEJS_DB_PASS`                | The password used to connect to the database                                        | A valid database password |    ✅    |                                    ✅                                    |
 | `CUBEJS_DB_CLICKHOUSE_READONLY` | Whether the ClickHouse user has read-only access or not                             | `true`, `false`           |    ❌    |                                    ✅                                    |
+| `CUBEJS_DB_CLICKHOUSE_WITHFILL` | Whether to use Clickhouse's `WITH FILL` modifier for date ranged queries            | `true`, `false`           |    ❌    |                                    ✅                                    |
 | `CUBEJS_CONCURRENCY`            | The number of concurrent connections each queue has to the database. Default is `5` | A valid number            |    ❌    |                                    ❌                                    |
 | `CUBEJS_DB_MAX_POOL`            | The maximum number of concurrent database connections to pool. Default is `20`      | A valid number            |    ❌    |                                    ✅                                    |
 
@@ -82,11 +83,21 @@ You can connect to a ClickHouse database when your user's permissions are
 [restricted][clickhouse-readonly] to read-only, by setting
 `CUBEJS_DB_CLICKHOUSE_READONLY` to `true`.
 
+You can use Clickhouse's [`WITH FILL` MODIFIER][clickhouse-with-fill] to fill
+any dates within the defined date range with `0` values. This will make sure
+that the date range is fully filled with data. This is useful for showing a
+true axis on a chart or exporting data for all dates given.
+
+Be aware that using `WITH FILL` may increase query times or resources used and you may need to increase the `limit`
+provided to your Cube queries.
+
 [clickhouse]: https://clickhouse.tech/
 [clickhouse-docs-users]:
   https://clickhouse.tech/docs/en/operations/settings/settings-users/
 [clickhouse-readonly]:
   https://clickhouse.tech/docs/en/operations/settings/permissions-for-queries/#settings_readonly
+[clickhouse-with-fill]:
+  https://clickhouse.com/docs/en/sql-reference/statements/select/order-by#order-by-expr-with-fill-modifier
 [ref-caching-using-preaggs-build-strats]:
   /caching/using-pre-aggregations#pre-aggregation-build-strategies
 [ref-config-multiple-ds-decorating-env]:
--- a/docs/content/Reference/Configuration/Environment-Variables-Reference.mdx
+++ b/docs/content/Reference/Configuration/Environment-Variables-Reference.mdx
@@ -177,6 +177,14 @@ Whether the ClickHouse user has read-only access or not.
 | --------------- | ---------------------- | --------------------- |
 | `true`, `false` | N/A                    | N/A                   |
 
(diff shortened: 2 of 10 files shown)


mrenigma:
@paveltiunov it's been a few weeks, so I just wondered if you had any initial thoughts regarding this PR? 🙏

igorlukanin:
Thank you @mrenigma! This PR looks good to me. I'd like to know what @paveltiunov thinks as well

paveltiunov:
@mrenigma Hey Robert! Thanks for contributing this! While Cube currently definitely lacks server-side post-processing, we believe drivers aren't the right place for this layer. The reason for it if we add some functionality to a driver, it should be supported by the majority of popular data warehouses and fill missing dates feature AFAIK is quite unique to clickhouse. Given that we'd be open to fill missing dates functionality together with server-side pivot however it probably should be done as a post-processing phase to general querying. It also most probably should have pretty same semantics as those we have in the Cube client.

mrenigma:
@paveltiunov many thanks for your response. I originally created an [issue](https://github.com/cube-js/cube/issues/6893) to get initial feedback before creating this PR, but it didn't get any traction. So I opted to work on the feature as:

1. Clickhouse driver is [community-supported and not in active development](https://www.npmjs.com/package/@cubejs-backend/clickhouse-driver#support).
2. This functionality is important to my team over at GitLab, and we wanted to give back to the community if we could.

I would like to respond to your points as I'm not entirely sure if I get what you're trying to say. Please let me know if I'm reading your response wrong or there is further information I haven't considered.

> While Cube currently definitely lacks server-side post-processing, we believe drivers aren't the right place for this layer. 

I don't believe this change fits into the server-side post-processing concept. As an example, one of Cube's current [open issues](https://github.com/cube-js/cube/issues/3799) would definitely fall into this category because it takes the result and looks to enrich it with additional data either through third-party sources or internal functions.

Which is actually something we at GitLab are already doing to some extent using a proxy layer and data transformer: https://gitlab.com/gitlab-org/gitlab/-/issues/415687.

However, the drawback to this approach is that it completely ignores any limits set by the Cube query (as the results process has already taken place). It is also more intensive because it loops through the results array to find the dates that might be missing from the date range.

Which is why I opted for using Clickhouse's built-in ability to do this for us as it takes into account any limits defined by the query; and is far more performant that trying to modify the result after the fact.

Which leads onto your next point:

> The reason for it if we add some functionality to a driver, it should be supported by the majority of popular data warehouses and fill missing dates feature AFAIK is quite unique to clickhouse.

I 💯 agree that this is very Clickhouse specific. I'm not aware of other drivers that support this kind of functionality (although there may be functionality I'm unaware of). Hence, why it's behind the clickhouse specific environment flag:

1. To opt in to your queries being altered at query time.
5. Because it's a Clickhouse specific feature, if a user is using multiple drivers, this approach will not work.

I placed it in the driver because of how specific to Clickhouse this feature is and how this acts almost like a query setting.

However, having said that:

> Given that we'd be open to fill missing dates functionality together with server-side pivot however it probably should be done as a post-processing phase to general querying. It also most probably should have pretty same semantics as those we have in the Cube client.

I would be very much open to moving some of the logic elsewhere in the stack if it is felt that other drivers could leverage this process in future. Although Clickhouse doesn't need to do this post-processing, I can see how other drivers might want to plug into the process. The part that makes this difficult is that in this scenario the implementation for Clickhouse would probably be vastly different from the other drivers due to how close to the query Clickhouse would run it.

So in the spirit of iteration, I would argue that this PR could be classes as “the first iteration” with subsequent iterations moving shared logic and creating a unified internal API for other drivers to leverage.

---

I'm interested to hear your thoughts, and to discuss a path forward here.

mrenigma:
@paveltiunov it's been a month since my last message, I'd really like to continue our discussion, but do let me know if this is a complete no-go.

paveltiunov:
@mrenigma Hey Robert! Sorry for the delay. I still feel it should be part of the bigger generic backend fill missing dates feature. I'm not opposed to pushing down fill missing dates implementation to drivers where appropriate. However, API-wise, it should be exposed on a query level, not an env flag. I think the next step to bring this to life is to design such changes to API.

mrenigma:
@paveltiunov apologies it's my turn to take forever to respond! I've been very busy with work and personal life.

> I still feel it should be part of the bigger generic backend fill missing dates feature. I'm not opposed to pushing down fill missing dates implementation to drivers where appropriate. However, API-wise, it should be exposed on a query level, not an env flag. I think the next step to bring this to life is to design such changes to API.

Thank you for verifying your position here. With this in mind I'm going to close this PR for two reasons:

1. This PR has been open a long time, there has been considerable changes made to the drivers and compilers, so the work here would need a considerable refactor to function as-is. And since, this PR doesn't meet the API requirements, it makes more sense to start again.
1. Creating a driver-agnostic API wasn't the driving force behind this change. The hope was to leverage the `WITH FILL` function that Clickhouse provides, and make use of the performance gains made there. Having it at the API level means needing to parse and process a potentially O(n) data array after it's been returned by the drivers and made usable by the internal API's.

Of course, if anyone else would like to pick up this work then they are more than welcome to update and take on https://github.com/cube-js/cube/issues/6893 🙂.

I appreciate everyone's contributions and feedback in this PR ❤️

Issue #7166, opened 2023-09-23
Vue package ignores `null` value after herusictics


DPHonys:
**Describe the bug**
The Frontend package for Vue ignores the `null` value for the pivot config when returned from the `defaultHerusictics` method that can be found in the core package [here](https://github.com/cube-js/cube/blob/0fb0ea7ad67a9546dc1173dd6c7ff6ea1c1e822b/packages/cubejs-client-core/src/utils.js#L89).

I've noticed that the query builder in the Vue package is behaving differently than in playground (from the code I've guessed it's using the React version of the package) when it comes to pivot generation after heuristics are run. The only difference in code I've found is that the React package uses a returned `null` value from the `defaultHerusictics` method and Vue does not.

React package
```js
finalState.pivotConfig = ResultSet.getNormalizedPivotConfig(
     finalState.query,
     finalState.pivotConfig !== undefined ? finalState.pivotConfig : statePivotConfig
);
```

Vue package
```js
this.pivotConfig = ResultSet.getNormalizedPivotConfig(
     validatedQuery,
     pivotConfig || this.pivotConfig
);
```

The difference is the React package passes the `null` value to the `getNormalizedPivotConfig` with only an exception when the returned value is `undefined` then it passes the current pivot config into the ResultSet `getNormalizedPivotConfig` method. 

In the meantime, the Vue package ignores the `null` and `undefined` values totally because the second parameter is always evaluated to the current pivot config so the generated pivot is not the same as in the React package.
```js
null || "ABC" // => "ABC"
undefined || "ABC" // => "ABC"
```

**Expected behavior**
The front-end packages should behave the same despite the framework used.

**Version:**
0.33.59

I've created a [PR](https://github.com/cube-js/cube/pull/7167) that fixes this bug in the Vue package so it behaves the same as the React package.

Thanks for reading 🙂

PR #7167, opened 2023-09-24
fix(@cubejs-client/vue): Pivot config can use null from heuristics

DPHonys:
**Check List**

[#7166] Vue package ignores `null` value after herusictics

- Fixes this bug I've made an issue on. Where the Vue frontend package totally ignores `null` / `undefined` from `defaultHerusictics` method so the pivot config is not correctly updated and it behaves differently from it's React counterpart.

- The only thing I've changed is that now we check if the value of the pivot config returned from `defaultHerusictics` is `undefined` then we pass the current pivot config but if it is `null` then we pass the `null` into the `ResultSet.getNormalizedPivotConfig` method.


paveltiunov:
@DPHonys Thanks for contributing! Could you please add a test for it?

DPHonys:
Hi @paveltiunov, after multiple hours of getting empty pivotConfig I've found out I needed to use $nextTick to get updated properties so I hope the test is correctly composed. 🙂

DPHonys:
@paveltiunov Also added the same fix and test to the Vue3 package. 🙂

PR #7174, opened 2023-09-27
feat(duckdb-driver): upgrade duckdb driver to 0.9.1

MikeNitsenko:


MikeNitsenko:
constantly getting the timeout for `yarn install`
the cause is mentioned here: https://github.com/duckdb/duckdb/issues/9113

as per the [comment](https://github.com/duckdb/duckdb/issues/9113#issuecomment-1736209359):
> Node CI job for the release has failed without deploying the right artefacts, and currently it's in a state that requires manual builds.
We are aware of this, it's lagging behind the other builds, but might need another day or two to sort this out properly on all platforms.

ovr (approved):
LGTM

PR #7198, opened 2023-10-06
Order can be an array of objects (firestore storing support)

patrikfejda:
**Check List**

https://cube-js.slack.com/archives/C04NYBJP7RQ/p1695722856481539

Add support for storing order as array of objects as firestore doesnt allow storing array of arrays.

### DIFF
--- a/packages/cubejs-api-gateway/src/query.js
+++ b/packages/cubejs-api-gateway/src/query.js
@@ -103,7 +103,10 @@ const querySchema = Joi.object().keys({
   }).oxor('dateRange', 'compareDateRange')),
   order: Joi.alternatives(
     Joi.object().pattern(id, Joi.valid('asc', 'desc')),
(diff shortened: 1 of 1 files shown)


paveltiunov:
@patrikfejda Could you please elaborate a little bit on your use case here? I'm curious why the format you propose can't be transformed into arrays on the client side.

patrikfejda:
@paveltiunov Hey, the use case is a need to use Alternative order format, while storing the query properties in Firestore (or other no-sql databases). This is unfortunatelly not possible, because it doesnt allow storin array of arrays.

In our specific usecase we store the cubes as documents in firestore and send the whole document as a parameter of a request to cube. Sure, for our usecase we can use a workaround on client, which would transform array of objects into array of arrays, but I am convinced I am not the only one using firestore for storing and my PR would really help many. In addition, it is a very small change in CubeJs with no other code dependecies.

PR #7311, opened 2023-10-26
fix(jdbc-driver): node-java-maven package.json path issue

MikeNitsenko:
Addressed an issue where the JDBC driver, utilizing the `node-java-maven` library, attempted to read a `package.json` file in the YAML deployment template directory. As this file was absent, queries failed to execute. The fix involves specifying the path to a valid `package.json` file using the `packageJsonPath` option in `node-java-maven`.


ovr (approved):
LGTM

Issue #6792, opened 2023-07-07
is there a reason why `meta` fields are not also included in `cube`s and `view`s ?


jxperf:
**Problem**
as titled

dodgex:
Hello @igorlukanin ,

while I am not the original poster of this question, i would like the same feature.

We are building a reporting fronted using cube as backend and having `meta` on cubes would allow to add stuff as descriptions and other metadata as groups. e.g we have multiple reporting areas in our cube schema and in our fronted they are seperated in stuff like "Human Resources", "Financial" and "Sales". right now we abuse the cube title and have like "AREA_NAME \ CUBE_NAME" to achieve that. having a dedicated meta field would be better in my opinion. :)

dodgex:
there is another issue with the same request btw: https://github.com/cube-js/cube/issues/1539

barakcoh:
same for us, we'd like to add metadata for the UI on how to display the cube (or hide it) in different parts of our app. 
More broadly, I think the same thinking that allowed for `meta` on members should apply here, without having to predict and evaluate the merits of specific use-cases, especially as this seems rather straightforward to implement.

igorlukanin:
Now you can use `meta` on cubes and views:
- https://cube.dev/docs/reference/data-model/cube#meta
- https://cube.dev/docs/reference/data-model/view#meta

Issue #1539, opened 2020-12-07
`meta` property on cube definition


TheSPD:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
The way we can add `meta` on each dimension or measure, we should be able to add a `meta` on the cubes themselves.

**Describe alternatives you've considered**
Having the meta JSON in the `description` as string

**Additional context**
Would be a great addition for extensibility.

igorlukanin:
Now you can use `meta` on cubes and views:
- https://cube.dev/docs/reference/data-model/cube#meta
- https://cube.dev/docs/reference/data-model/view#meta

PR #7327, opened 2023-10-27
feat(schema-compiler): add meta property for cube definition

mharrisb1:

#1539 and #6792

Adds support for `meta` property at the cube and view level.

Examples (added to reference docs):

```javascript
cube(`orders`, {
  sql_table: `orders`,
  title: `Product Orders`,
  meta: {
    any: "value",
  }
});
```

```yaml
cubes:
  - name: orders
    sql_table: orders
    title: Product Orders
    meta:
      any: value
```


mharrisb1:
I allowed for arbitrary values in `meta` including for `null` types. Adding that to the logic in `transpileYaml` broke the expectation that `null` values will cause an error so I added an additional predicate that will allow for `null` values but only within the `meta` field.

I really didn't want to treat `meta` as a special field so I'm open to suggestions on what to do there. I can fully remove support for `null` (which will need to be added as a clause in the docs) but I could see scenarios where that would be useful.

paveltiunov:
@mharrisb1 Overall, it looks good to me! Could you please elaborate on null use case for meta?

mharrisb1:
@paveltiunov yeah, of course. 

My main point was mostly regarding YAML transpilation.

The following test relied on the fact that `null` wasn't being handled in [`transpileYaml`](https://github.com/cube-js/cube/blob/b8aa735dd98cc74df10f3fcea980ee064b4c7c15/packages/cubejs-schema-compiler/src/compiler/YamlCompiler.ts#L137):

Test: `Yaml Schema Testing` > `null for string field`
https://github.com/cube-js/cube/blob/b8aa735dd98cc74df10f3fcea980ee064b4c7c15/packages/cubejs-schema-compiler/test/unit/yaml-schema.test.ts#L77

To support any values in `meta` I did add handling for `null` values. This broke that test case so to fix I added a new condition to only handle `null` values iff `meta` is in the property path.

That didn't seem like the best way to go about it so I just wanted to draw attention to it.

If nulls are allowed in parts of the file but not others it might be weird to a user (who doesn't know the internals of the schema compiling package) that this would be valid:

```yml
cubes:
  - name: orders
    sql: `select * from public.orders`
    meta:
      some_key: null
    ...
```

But this would result in a somewhat ambiguous error, `Unexpected input during yaml transpiling: null`:

```yml
cubes:
  - name: orders
    sql: `select * from public.orders`
    title: null
    meta:
      some_key: null
    ...
```

Any thoughts on this?

One solution could be to just not handle `null` in the `meta` property even though there are use cases for needing it. Maybe users can just use an empty string instead.

Another could be to handle `null` values anywhere in the file. Instead of `null` being generally unexpected, the failure could come from the specific property like `Unexpected input for property "title": null`.

mharrisb1:
@paveltiunov truthfully I don't have any in mind (at least not for my team), but I'm sure other teams might have a need.

Maybe something like:

```yml
cubes:
  - name: orders
    sql: `select * from public.orders`
    meta:
      icon_png: null

  - name: accounts
    sql: `select * from public.accounts`
    meta:
      icon_png: gs://path/to/some/blob.png
```

A user can just use an empty string instead of null but they'd have to run into the unexpected input error first before they would know to try that, and that just feels like throwing darts at the wall and seeing what sticks.

Maybe for now, instead of me anticipating an imaginary use case, we remove support for null values in meta and let someone post an issue if they have a concrete use case.

mharrisb1:
@paveltiunov I went ahead and removed support for nulls

igorlukanin:
I think there's also a use case when `meta` is generated programmatically or brought from an external system that has flexible rules for allowed values. For example, I can see [dbt models' `meta`](https://docs.getdbt.com/reference/resource-configs/meta) directly injected as cubes' `meta` as part of dbt integration; my understanding is that it can be anything in there, including `null`s.

igorlukanin:
@paveltiunov @mharrisb1 Do we need more tests here, e.g., for `view`s + to check that `meta` values are exposed via `/load` and `/meta` endpoints of the REST API?

mharrisb1:
> I think there's also a use case when `meta` is generated programmatically or brought from an external system that has flexible rules for allowed values. For example, I can see [dbt models' `meta`](https://docs.getdbt.com/reference/resource-configs/meta) directly injected as cubes' `meta` as part of dbt integration; my understanding is that it can be anything in there, including `null`s.

Good call. Developer flexibility is probably the way to go.

@paveltiunov I haven't dug in too deep on the compiler code for this PR, but what would happen if `null` is handled in `transpileYaml`?

For example, if we had this YAML config:

```yml
cubes:
  - name: orders
    sql: `select * from public.orders`
    title: null
```

Would an error be thrown for `title` having a null value *after* transpilation with this JS code?

```js
cube(`orders`, {
  sql: `select * from public.orders`,
  title: null,
});
```

If so, then could we just transpile the YAML code as is and let the error come from JS?

mharrisb1:
> Do we need more tests here, e.g., for `view`s + to check that `meta` values are exposed via `/load` and `/meta` endpoints of the REST API?

@igorlukanin another good call. I was manually testing this by building the dev Docker image and running locally but actual tests would be better.

Can you point me to the integration tests where this would be added? I'm still trying to get familiarized with the codebase

benhickson:
i think that `meta` should be any valid JSON-serializable value, and nulls are meaningful values.

paveltiunov:
@mharrisb1 Ah. Gotcha. I thought there was `meta: null` use case. But I agree we need to support nested nulls inside meta for sure. So ideally to have it back.

mharrisb1:
I reverted the removal of null support in meta 👍

jwilson232:
this would be super useful for my team!

hannosgit:
@paveltiunov Do you have any timeline on when this could be integrated into the main branch?

mharrisb1:
Saw that the commit for docs was deleted in previous large merge. Added back now.

PR #7355, opened 2023-10-29
fix(duckdb-driver): use schema param in information_schema queries

MikeNitsenko:
There is a change on DuckDB's side, so previously `SET schema=<value>` worked for `information_schema` queries.
Now, we need the combination of `SET` and `WHERE` clause.
- `WHERE` clause will narrow the list of returned schemas
- `SET` will allow using these tables without referencing the catalog to resolve ambiguity.

> There is confusion with the param naming on DuckDB's side. While the param is called `schema` in [docs](https://duckdb.org/docs/archive/0.9.1/sql/configuration), when you issue `SET schema=<value>` it sets the **catalog**.

### DIFF
--- a/packages/cubejs-duckdb-driver/src/DuckDBDriver.ts
+++ b/packages/cubejs-duckdb-driver/src/DuckDBDriver.ts
@@ -26,10 +26,14 @@ type InitPromise = {
 export class DuckDBDriver extends BaseDriver implements DriverInterface {
   protected initPromise: Promise<InitPromise> | null = null;
 
+  private schema: string;
+
   public constructor(
     protected readonly config: DuckDBDriverConfiguration = {},
   ) {
     super();
+
+    this.schema = getEnv('duckdbSchema', this.config);
   }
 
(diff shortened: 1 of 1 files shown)


packages/cubejs-duckdb-driver/src/DuckDBDriver.ts
```diff
@@ -26,10 +26,14 @@ type InitPromise = {
...
   protected initPromise: Promise<InitPromise> | null = null;
 
+  private schema: string;
+
   public constructor(
     protected readonly config: DuckDBDriverConfiguration = {},
   ) {
     super();
+
+    this.schema = getEnv('duckdbSchema', this.config);
```
ovr, line 36: Let's allow the override schema by passing it with config. We use the same approach for all drivers.

ovr (approved):

ovr (changes requested):
unit tests failed...

MikeNitsenko:
> unit tests failed...

@ovr I restarted tests and they passed ok

PR #7398, opened 2023-11-10
chore: Remove the reference to NODE_ENV=production

igorlukanin:
1. Remove the reference to `NODE_ENV=production`; it's obsolete since Cube is now run in Docker.
2. Fix wording: 'developer mode` → `development mode`; we have the latter in [docs](https://cube.dev/docs/product/configuration#development-mode).

For context — conversation in Slack: https://cube-js.slack.com/archives/C04NYBJP7RQ/p1699263331658569

### DIFF
--- a/packages/cubejs-server-core/src/core/DevServer.ts
+++ b/packages/cubejs-server-core/src/core/DevServer.ts
@@ -52,7 +52,7 @@ export class DevServer {
     const cubejsToken = jwt.sign({}, options.apiSecret || 'secret', { expiresIn: '1d' });
 
     if (process.env.NODE_ENV !== 'production') {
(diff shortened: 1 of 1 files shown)


packages/cubejs-server-core/src/core/DevServer.ts
```diff
@@ -52,7 +52,7 @@ export class DevServer {
     const cubejsToken = jwt.sign({}, options.apiSecret || 'secret', { expiresIn: '1d' });
 
     if (process.env.NODE_ENV !== 'production') {
```
ovr, line 54: `if (process.env.NODE_ENV !== 'production') {` @igorlukanin  :)

Issue #7407, opened 2023-11-13
Athena driver allow to pass credentialsProvider from config


trompa:
**Is your feature request related to a problem? Please describe.**
Hello, we are not allowed to use static credentials as per compliance with some standards, so we cannot use accesKeyID and secretAccessKey.
Passing a token wouldnt either work, as it needs to be refreshed.
Credentials field gets override by constructor so we cannot pass a crendentialsProvider

**Describe the solution you'd like**
I want to be able to pass a credentialsProvider on the config and have it forwarded to the AthenaClient SDK class so it handles authentication internally.
crendentials field in the AthenClientConfig accepts [both credentials or credentials Provider](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-client-athena/Interface/AthenaClientConfig/) 
But it gets override.
i'd want it to left untouched if it comes filled from caller:
so a change from 

```      credentials: accessKeyId && secretAccessKey
        ? { accessKeyId, secretAccessKey }
        : undefined,
```
to
```
credentials = config.credentials || (accessKeyId && secretAccessKey ? { accessKeyId, secretAccessKey } : undefined);
```
Would be enough to allow us to pass the provider

**Describe alternatives you've considered**

issue #6036 explores 2 options.
Their option 2 would be superseded by this one, as it would allow to pass the object with token from outside.

So it would be their Option 1: make the config over-writeable again
Or the option presented here.

As probably the change to make the config non over-writable was done on purpose. I'd go with the option presented here.

**Additional context**

trompa:
I created this PR with the proposed changes
https://github.com/cube-js/cube/pull/7429

PR #7429, opened 2023-11-17
fix(athena-driver) #7407 allow to pass credentials

trompa:
**Check List**

[For example #12]

[Description goes here]

### DIFF
--- a/packages/cubejs-athena-driver/src/AthenaDriver.ts
+++ b/packages/cubejs-athena-driver/src/AthenaDriver.ts
@@ -125,7 +125,13 @@ export class AthenaDriver extends BaseDriver implements DriverInterface {
       config.secretAccessKey ||
       getEnv('athenaAwsSecret', { dataSource });
 
-    const { schema, credentials, ...restConfig } = config;
+    const { schema, ...restConfig } = config;
+    
+    let { credentials } = config;
+    
+    if (typeof credentials === undefined) {
+      credentials = accessKeyId && secretAccessKey ? { accessKeyId, secretAccessKey } : undefined;
+    }
 
     this.schema = schema ||
       getEnv('dbName', { dataSource }) ||
@@ -133,10 +139,7 @@ export class AthenaDriver extends BaseDriver implements DriverInterface {
 
     this.config = {
       ...restConfig,
-      credentials: typeof credentials !== undefined? credentials :
-       accessKeyId && secretAccessKey
-        ? { accessKeyId, secretAccessKey }
-        : undefined,
+      credentials,
       region:
         config.region ||
         getEnv('athenaAwsRegion', { dataSource }),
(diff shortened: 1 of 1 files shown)


trompa:
I havent added an example for this use case to the docs, but I can do if requested.
We are currently using this autorefreshing credential:

```
const createAssumeRoleCredentials = async (sts, roleArn) => {
  const roleSessionName = 'my-session';

  const refreshCredentials = async () => {
      try {
          const data = await sts.assumeRole({
              RoleArn: roleArn,
              RoleSessionName: roleSessionName,
              DurationSeconds: 900 // TODO: For testing, need to remove
          }).promise();

          credentials.accessKeyId = data.Credentials.AccessKeyId;
          credentials.secretAccessKey = data.Credentials.SecretAccessKey;
          credentials.sessionToken = data.Credentials.SessionToken;
          credentials.expireTime = data.Credentials.Expiration;
      } catch (err) {
          console.error("Error assuming role:", err);
          throw err; // Rethrow the error for the caller to handle
      }
  };

  let credentials = new AWS.Credentials();
  credentials.refresh = refreshCredentials;

  // Initially refresh the credentials
  try {
      await refreshCredentials();
      console.log("Initial credential refresh successful");
  } catch (err) {
      console.error("Initial credential refresh failed:", err);
      throw err; // Rethrow the error for the caller to handle
  }

  return credentials;
};
```

Which could be a great example, no matter what kind of role the user would want to use, as i feel the most important part is where the credential.refresh method gets overriden.

trompa:
I've actually found that this change wont work if I want to pass a provider. It compares to undefined and will override it.
will push a change where it checks typeof, so promises are also evaluated correctly

packages/cubejs-athena-driver/src/AthenaDriver.ts
```diff
@@ -126,16 +126,20 @@ export class AthenaDriver extends BaseDriver implements DriverInterface {
...
     this.schema = schema ||
       getEnv('dbName', { dataSource }) ||
       getEnv('dbSchema', { dataSource });
 
     this.config = {
       ...restConfig,
-      credentials: accessKeyId && secretAccessKey
-        ? { accessKeyId, secretAccessKey }
-        : undefined,
+      credentials,
```
ovr, line 142: I suggest simplifying this code by moving the original code for `credentials`:

```
credentials: accessKeyId && secretAccessKey
        ? { accessKeyId, secretAccessKey }
        : undefined,
```

before `...restConfig`.

It will allow anyone to redeclare `credentials` by passing it via `constructor`

ovr: image

ovr:
Thank you @trompa, for your contribution! 🍰

M1ke:
This appears to cause a regression for services using instance metadata and not passing credentials in. We have a working cube install in production on v0.33 that when upgraded to v0.34 returns:

> Resolved credential object is not valid

Rolling back to v0.33 resolved this. Checking the diff I can't see exactly why this might be the case, unless `...restConfig` is exposing some requirement for credentials that's overwriting the `credentials` key in the object.

ovr:
@M1ke 

> This appears to cause a regression for services using instance metadata and not passing credentials in.

It doesn't change that behavior. `credentials` is defined in both versions (before/after). This PR allows to override default credentials. Do you pass `undefined` as `credentials`?

Thanks

PR #7474, opened 2023-11-29
bugfix: Invalid QuestDB timestamp floor year unit

peterklingelhofer:
**Check List**

When aggregating data on the year level, we see this error in QuestDB in the generated SQL:
```sql
SELECT timestamp_floor(‘Y’, to_timezone("data_hourly".timestamp, 'UTC')) "data_hourly__timestamp_year",
FROM "data_hourly"
LIMIT 5000;
```

`invalid unit 'Y'`

As you can see in the [QuestDB documentation](https://questdb.io/docs/reference/function/date-time/), `Y` is not a valid option.

Adjusting to use the lowercase `y` yields results as expected:
```sql
SELECT timestamp_floor('y', to_timezone("data_hourly".timestamp, 'UTC')) "data_hourly__timestamp_year",
FROM "data_hourly"
LIMIT 5000;
```

Note: This only seems to be a problem if we try to grab `timestamp__year` in the GraphQL query.

```gql
query CubeQuery  { 
  cube(where: {dataHourly: {timestamp: {inDateRange: "This year" } }}) {
    dataHourly(orderBy: {timestamp: asc}) {
      data_column_1
      data_column_2
      timestamp {
        year
      }
    }
  }
}
```
Yields the above `invalid unit 'Y'` error.

Whereas this query works fine:
```gql
query CubeQuery  { 
  cube(where: {dataHourly: {timestamp: {inDateRange: "This year" } }}) {
    dataHourly(orderBy: {timestamp: asc}) {
      data_column_1
      data_column_2
    }
  }
}
```


paveltiunov:
@peterklingelhofer Thanks for contributing! Could you please add a test for it?

peterklingelhofer:
> @peterklingelhofer Thanks for contributing! Could you please add a test for it?

Added base and base snake case GraphQL queries to test for this: https://github.com/cube-js/cube/pull/7474/commits/a3238addea5934ed08bcdc6b7c19503234675e9d

peterklingelhofer:
@paveltiunov not sure if the tests are just broken with respect to testing year (I noticed there were no year tests before I added this one) or if it's my change. Regardless, adding a GraphQL query to test is not specifically testing the QuestDB driver from what I can tell.

Unfortunately I don't have enough time to dig further - I can close this out and simply create an issue linking to this MR for context. It's not really that big of a deal, if you want data for "This year" it's probably not that hard to figure out what year this year is, and I don't need to for my use-case, just was a bug I noticed in passing that I figured I'd make a quick attempt on.

paveltiunov:
@peterklingelhofer I guess snapshots weren't updated.

PR #7520, opened 2023-12-12
fix(api-gateway): Fallback to global error middelware for async handlers

ovr:
Hello!

Thanks

### DIFF
--- a/packages/cubejs-api-gateway/src/gateway.ts
+++ b/packages/cubejs-api-gateway/src/gateway.ts
@@ -15,7 +15,8 @@ import type {
   ErrorRequestHandler,
   NextFunction,
   RequestHandler,
-  Response,
+  Request as ExpressRequest,
+  Response as ExpressResponse,
 } from 'express';
 import {
   QueryType,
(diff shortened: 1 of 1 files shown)


MikeNitsenko (approved):

RusovDmitriy (approved):

packages/cubejs-api-gateway/src/gateway.ts
```diff
@@ -91,9 +92,18 @@ import {
 
 const memberExpressionRegex = /^([a-zA-Z0-9_]+).([a-zA-Z0-9_]+):\(([a-zA-Z0-9_,]+)\):(.*)$/;
 
-/**
- * API gateway server class.
- */
+function userAsyncHandler(handler: (req: Request & { context: ExtendedRequestContext }, res: ExpressResponse) => Promise<void>) {
+  return (req: ExpressRequest, res: ExpressResponse, next: NextFunction) => {
```
tenphi, line 96: We need to add `async` before the function to make sure it will not throw a sync error.

packages/cubejs-api-gateway/src/gateway.ts
```diff
@@ -91,9 +92,18 @@ import {
...
+    handler(req as any, res).catch(next);
+  };
+}
+
+function systemAsyncHandler(handler: (req: Request & { context: ExtendedRequestContext }, res: ExpressResponse) => Promise<void>) {
+  return (req: ExpressRequest, res: ExpressResponse, next: NextFunction) => {
```
tenphi, line 102: Here as well.

packages/cubejs-api-gateway/src/gateway.ts
ovr: 1. `handler` is limited with TS by returning type = `=> Promise<void>`.
2. function returns wrapper = sync handler = express.Handler

tenphi: You can't limit async functions by Typescript: 

https://www.typescriptlang.org/play?#code/C4TwDgpgBAggziAdgYwGIFcVQLxQBQCWiAJhAB4BcUi6AtgEYQBOAlDgHxQAKTA9rQTgQAPAENEIdgG4AUDOS9EcYFFEIUGFFXhI0mZDnxFSZNtk4BvGVBtQCAMyMlyUYVAAMbK7Z9RgACz4Ad2oIEIBRJj4mPAByUWAE0ViWWR8AX2tbJghgdCZEbj4BIQA6HLheABsANwhCZ1NZdNkgA

This snippet does create a function that throws a sync error despite being TS-validated.

ovr: Technically it's possible. For example, It's possible to do process.exit, but we don't monkey patch process.exit to protect it, because this approach is incorrect and we cannot imagine that someone will write code like this and it will pass review.

PR #7521, opened 2023-12-12
fix(api-gateway): Catch error from contextToApiScopes

ovr:
Hello!

Thanks

### DIFF
--- a/packages/cubejs-api-gateway/src/gateway.ts
+++ b/packages/cubejs-api-gateway/src/gateway.ts
@@ -210,23 +210,13 @@ class ApiGateway {
 
     app.use(
       `${this.basePath}/graphql`,
-      [
-        ...userMiddlewares,
-        async (req, res, next) => {
-          try {
-            await this.assertApiScope(
-              'graphql',
-              req?.context?.securityContext
-            );
-            if (next) next();
-          } catch (e: unknown) {
-            if (e instanceof CubejsHandlerError) {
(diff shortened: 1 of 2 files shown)


packages/cubejs-api-gateway/src/gateway.ts
```diff
@@ -210,23 +210,13 @@ class ApiGateway {
...
-          try {
-            await this.assertApiScope(
-              'graphql',
-              req?.context?.securityContext
-            );
-            if (next) next();
-          } catch (e: unknown) {
-            if (e instanceof CubejsHandlerError) {
-              res.status(e.status).json({ error: e.message });
-            }
```
ovr, line 225: There is no call for `next` when error != `CubejsHandlerError`.
This will lead to unhandled connection.

packages/cubejs-api-gateway/src/gateway.ts
```diff
@@ -339,29 +329,20 @@ class ApiGateway {
...
-        ...userMiddlewares,
-        async (req, res, next) => {
-          try {
-            await this.assertApiScope(
-              'meta',
-              req?.context?.securityContext
-            );
-            if (next) next();
-          } catch (e: unknown) {
-            if (e instanceof CubejsHandlerError) {
```
ovr, line 352: There is no call for `next` when error != `CubejsHandlerError`.
This will lead to unhandled connection.

RusovDmitriy (approved):

PR #7645, opened 2024-01-11
fix(playground): replace antd overlay

vasilev-alex:


packages/cubejs-playground/src/components/SectionRow.tsx
```diff
@@ -8,7 +8,7 @@ const SectionRow = styled.div`
 
   && > * {
     margin-right: 8px !important;
-    margin-bottom: 8px !important;
+    // margin-bottom: 8px !important;
```
furkando, line 11: comment left here

PR #7617, opened 2024-01-09
feat(view): members inherit title attribute

CallumWalterWhite:
**Check List**
- [x] Tests has been run in packages where changes made if available
- [x] Linter has been run for changed code
- [x] Tests for the changes have been added if not covered yet
- [x] Docs have been added / updated if required

**Issue Reference this PR resolves**

#7475

**Description of Changes Made (if issue reference is not provided)**

Modified CubeSymbol.js to assign the resolved members' title.

PR #7692, opened 2024-01-26
fix: Revert drillMembers include in views as it's a breaking change. …

paveltiunov:
…Need to re-introduce it through evaluation on cube level and validating drillMembers are included in view

**Check List**

#7617


TobiasMaehl-pIX:
@paveltiunov Any update on when this will be reintroduced? This is one of the major pain points we have with Cube right now.

PR #7853, opened 2024-02-28
fix: Ungrouped filtered count should contain `CASE WHEN` statement bu…

paveltiunov:
…t renders as `1`

**Check List**

### DIFF
--- a/packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
+++ b/packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
@@ -2005,7 +2005,7 @@ export class BaseQuery {
     }
     if (this.ungrouped) {
       if (symbol.type === 'count' || symbol.type === 'countDistinct' || symbol.type === 'countDistinctApprox') {
-        return evaluateSql === '*' ? '1' : evaluateSql;
+        return evaluateSql === '*' ? '1' : this.caseWhenStatement([{ sql: `(${evaluateSql}) IS NOT NULL`, label: `1` }]);
       } else {
         return evaluateSql;
       }
--- a/packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
+++ b/packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
@@ -2005,7 +2005,7 @@ export class BaseQuery {
     }
     if (this.ungrouped) {
       if (symbol.type === 'count' || symbol.type === 'countDistinct' || symbol.type === 'countDistinctApprox') {
-        return '1';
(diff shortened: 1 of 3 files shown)


packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
```diff
@@ -2005,7 +2005,7 @@ export class BaseQuery {
     }
     if (this.ungrouped) {
       if (symbol.type === 'count' || symbol.type === 'countDistinct' || symbol.type === 'countDistinctApprox') {
-        return '1';
+        return evaluateSql === '*' ? '1' : this.caseWhenStatement([{ sql: `(${evaluateSql}) IS NOT NULL`, label: `1` }]);
```
adamstruck, line 2008: I am still seeing issues related to this. The new case statement of the form `CASE WHEN (<dimension>) IS NOT NULL THEN 1 END` produces the wrong result when it is included in a `COUNT(DISTINCT <dimension>`. 

Note: the example below is truncated to highlight the problematic generated SQL.

```
SELECT 
  "case_when_nexus_" "nexus__utilizati",
  "count_distinct_n" "nexus__utilizati_1" 
FROM (
  SELECT
    CASE WHEN "nexus__utilizations"."point__in__time_" THEN ? ELSE ? END "case_when_nexus_",
    COUNT(DISTINCT "nexus__utilizations"."utilizations__un") "count_distinct_n" 
  FROM (
    SELECT
      CASE WHEN ("utilizations".observation_id) IS NOT NULL THEN 1 END "utilizations__un",
      "point__in__time__hvhn".is_hvhn "point__in__time_"
    FROM
      yellow.metric_speed_train__utilizations__v1 AS "utilizations"
      LEFT JOIN yellow.data_science__member_feature_bank__is_hvhn__v1 AS "point__in__time__hvhn" ON "utilizations".entity_id = "point__in__time__hvhn".entity_id and "utilizations".window_end_date BETWEEN "point__in__time__hvhn".start_date AND "point__in__time__hvhn".end_date
    WHERE 
      (
        (
          ("utilizations".window_end_date >= from_iso8601_timestamp('2023-03-02T00:00:00.000Z'))
          AND ("utilizations".window_end_date < from_iso8601_timestamp('2024-03-01T00:00:00.000Z'))
        )
      )
    LIMIT 10000
  ) AS "nexus__utilizations" 
) AS "nexus__utilizations" 
```

adamstruck: https://github.com/cube-js/cube/issues/7858

Issue #7545, opened 2023-12-18
cubestore memory leak


schneefux:
**Describe the bug**
I am running a self-hosted cubestore router with docker in a Nomad cluster. Since the last upgrade, after about 10 minutes, cubestore OOMs or becomes unresponsive to readyz checks and gets restarted by the Nomad scheduler. There are 3-5 cube API instances connected to my cubestore router, doing 10-50 queries per second each. There are no preaggregations or workers set up.

Memory usage before, with v0.34.33:
<img width="494" alt="image" src="https://github.com/cube-js/cube/assets/15379000/df3300b2-ba28-43ed-a385-f40021680b5d">

Memory usage now, with v0.34.36:
<img width="494" alt="image" src="https://github.com/cube-js/cube/assets/15379000/8545a481-ce09-43cc-b660-c047d7cd1a29">

ovr:
Hello @schneefux ,

I made some improvements across the last releases.
Could you give it a try [v0.35.8](https://github.com/cube-js/cube/releases/tag/v0.35.8)?

Thanks

igorlukanin:
@schneefux Did you try the latest version? I've seen reports in the community Slack (https://slack.cube.dev) that it makes a real difference with regards to memory.

igorlukanin:
Closing due to inactivity—also, according to user reports, the memory usage has indeed decreased in v0.35.x and v0.36.x.

schneefux:
I tried upgrading from v0.35 to v1.3 today, replacing Redis by cubestore and unfortunately I am still experiencing performance issues related to cubestore.

Here are some more details about my setup. Does anything stand out?

## Configuration

### API nodes

  * 3 instances by default, can autoscale up to 10
  * 512 MB RAM allocated for each node
  * image is based on `cubejs/cube:v1.3`

```env
NODE_OPTIONS = "--max-old-space-size=512"
CUBEJS_CONCURRENCY = "20"
CUBEJS_DB_MAX_POOL = "128"
CUBEJS_LOG_LEVEL = "info"
CUBEJS_DB_QUERY_TIMEOUT = "2m"
CUBEJS_API_SECRET = "(redacted)"
CUBEJS_DB_HOST = "clickhouse"
CUBEJS_CUBESTORE_HOST = "cubestore"
CUBEJS_CUBESTORE_PORT = "28500"
```

### Refresh nodes

  * 1 instance
  * 192 MB RAM
  * same image as API nodes

```env
[...]

PR #7959, opened 2024-03-18
fix: Redis driver throws TypeError under high load

schneefux:
**Check List**

N/A

Fixes "TypeError: result is not iterable at RedisQueueDriverConnection.retrieveForProcessing". The return value of `execAsync` is undefined if the redis transaction fails due to out of memory errors, see https://redis.io/docs/interact/transactions/#errors-inside-a-transaction. This is a regression of 101b85f0a2e5852a01b19ecd43d75736124664b9.

### DIFF
--- a/packages/cubejs-query-orchestrator/src/orchestrator/RedisQueueDriver.ts
+++ b/packages/cubejs-query-orchestrator/src/orchestrator/RedisQueueDriver.ts
@@ -291,20 +291,23 @@ export class RedisQueueDriverConnection implements QueueDriverConnectionInterfac
           .zadd([this.heartBeatRedisKey(), 'NX', new Date().getTime(), this.redisHash(queryKey)])
           .execAsync();
 
-      if (result) {
-        result[4] = JSON.parse(result[4]);
+      if (!result) {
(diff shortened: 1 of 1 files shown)


igorlukanin:
Hi @schneefux 👋 I'll let @paveltiunov to review this one but, just FYI, Redis support is deprecated and is subject to be removed from Cube: https://cube.dev/docs/product/deployment#redis

paveltiunov:
@schneefux Redis support is deprecated, and we do not accept any changes to this code base. Please consider moving to Cube Store.

schneefux:
Since migrating to cubestore in early fall last year, I've had severe performance issues, the cube API servers and cubestore would randomly hang up. This was probably caused by a memory leak introduced in 0.34, see #7545.

I have been running the Redis store for multiple years without any issues so I'd prefer to use it as long as possible until cubestore is more stable.

paveltiunov:
@schneefux Sorry to hear that! We didn't see anything like that in our production environments, so it might be configuration differences. We're open to merging any Cube Store fixes regarding OOM, and I believe this is the only way we have right now.

PR #7967, opened 2024-03-19
feat(cubestore): Reduce memory usage while writting WALs

ovr:
Hello!

Thanks

### DIFF
--- a/rust/cubestore/cubestore/src/metastore/rocks_fs.rs
+++ b/rust/cubestore/cubestore/src/metastore/rocks_fs.rs
@@ -30,7 +30,7 @@ pub trait MetaStoreFs: Send + Sync {
         &self,
         dir: &str,
         seq_number: u64,
-        serializer: &WriteBatchContainer,
+        serializer: WriteBatchContainer,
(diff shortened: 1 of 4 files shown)


rust/cubestore/cubestore/src/metastore/rocks_fs.rs
```diff
@@ -348,11 +348,13 @@ impl MetaStoreFs for BaseRocksStoreFs {
         &self,
         dir: &str,
         seq_number: u64,
-        serializer: &WriteBatchContainer,
+        serializer: WriteBatchContainer,
     ) -> Result<u64, CubeError> {
         let log_name = format!("{}/{}.flex", dir, seq_number);
         let file_name = self.remote_fs.local_file(log_name.clone()).await?;
+
         serializer.write_to_file(&file_name).await?;
```
ovr, line 356: We should lose `WriteBatchContainer` with move semantic because `upload_file` can take a long time for uploading file—for example, s3 backend upload file by 8MIB chunks.

waralexrom (approved):

Issue #7446, opened 2023-11-22
Unable to get total number of records using "total":true attribute for MSSQL database.


niravjprabtani:
**Describe the bug**
We are currently implementing server-side pagination in a web application, utilizing the "total": true attribute to obtain an exact count of records based on given conditions, irrespective of limit and offset. This helps in acquiring insights into the total number of pages for the application. However, when executing the Cube.js API, an error occurs.

```
{
    "error": "Error: The ORDER BY clause is invalid in views, inline functions, derived tables, subqueries, and common table expressions, unless TOP, OFFSET or FOR XML is also specified.",
    "requestId": "dd9ab55d-3e9f-4db8-ab96-7526a0c7eeb8-span-1"
    
```
**Investigation and Findings**
Upon switching the database from MSSQL to Snowflake, it became evident that the issue pertains specifically to MSSQL. Further investigation revealed that Cube.js internally executes an identical query without limit and offset to obtain the total record count for the API request. Cube.js wraps the original query within a "select count(*) from" statement. This method fails in MSSQL due to its lack of support for an order statement within this clause.

**To Reproduce**
1. Use MSSQL as a data source
2. Keep "total":true in the JSON request for any API
3. Execute load endpoint for Cube.js API

**Expected behavior**
It should give results along with the total number of records.

**Minimally reproducible Cube Schema**
You can use selects without tables in order to achieve that as follows.

```javascript
cube(`Orders`, {
  sql: `
  select 1 as id, 100 as amount, 'new' status
  UNION ALL
  select 2 as id, 200 as amount, 'new' status
  UNION ALL
  select 3 as id, 300 as amount, 'processed' status
  UNION ALL
  select 4 as id, 500 as amount, 'processed' status
  UNION ALL
  select 5 as id, 600 as amount, 'shipped' status
  `,
  measures: {
    count: {
      type: `count`,
    },
    totalAmount: {
      sql: `amount`,
      type: `sum`,
    },
    toRemove: {
      type: `count`,
    },
  },
  dimensions: {
    status: {
      sql: `status`,
      type: `string`,
    },
  },
});
```
Please invoke below API request to get the API response

**Sample request**
URL : http://localhost:4000/cubejs-api/v1/load
Query param :
```
{
    "dimensions": [
        "Orders.status"
    ],
    "timeDimensions": [],
    "total": true,
    "limit":10,
    "offset":1
}
``` 
This should provide list of status along with total number of records.

**Version:**
Here is my sample docker-compose.xml

[...]

PR #8060, opened 2024-04-01
fix(api-gateway): fixes an issue where queries t…

rdwoodring:
fix(api-gateway) fix(schema-compiler): fixes an issue where queries to get the total count of results were incorrectly applying sorting from the original query and also were getting default ordering applied when the query ordering was stripped out

**Check List**

[#7446](https://github.com/cube-js/cube/issues/7446)

- updates api gateway to remove `order` field from queries that are getting a count of total results matching
- updates the base query's logic for defaulting `order` so that it does an early return and _does not_ default `order` if the query is getting the total (follows the same path as the logic doing an early return for preaggreation queries)

### DIFF
--- a/packages/cubejs-api-gateway/src/gateway.ts
+++ b/packages/cubejs-api-gateway/src/gateway.ts
@@ -1421,7 +1421,7 @@ class ApiGateway {
       const normalizedTotal = structuredClone(normalizedQuery);
       normalizedTotal.totalQuery = true;
       
-      delete normalizedTotal.order
+      delete normalizedTotal.order;
 
       normalizedTotal.limit = null;
       normalizedTotal.rowLimit = null;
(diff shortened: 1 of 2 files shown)


paveltiunov:
@rdwoodring Thanks for contributing! We'd need E2E tests for that. You can put it in the Postgres smoke test.

rdwoodring:
Any specific tests you'd like to see, @paveltiunov ? Just that queries with `totalQuery` return results without an error? Or can I actually test what query is being generated against the database to see that it does not have an order by?

Thanks

paveltiunov:
@rdwoodring The later

niravjprabtani:
Could we please get this PR reviewed ? I am facing issues in one of the project due to this and I had to execute multiple same queries just to get total number of records.

PR #8115, opened 2024-04-10
feat: primary and foreign keys driver queries

vasilev-alex:


packages/cubejs-base-driver/src/BaseDriver.ts
```diff
@@ -40,6 +40,7 @@ import {
   QueryTablesResult,
   QueryColumnsResult,
   TableMemoryData,
+  PrimeryKeysQueryResult,
```
ovr, line 43: `Primary`

packages/cubejs-base-driver/src/BaseDriver.ts
```diff
@@ -326,10 +349,24 @@ export abstract class BaseDriver implements DriverInterface {
     return sortByKeys(result);
   }
 
-  public tablesSchema() {
+  public async tablesSchema() {
```
ovr, line 352: Think about `BigQuery.tablesSchema`.

PR #8182, opened 2024-04-23
feat(firebolt): Switch to the new auth method

ptiurin:
**Check List**

Firebolt has a new version that requires service account authentication in order to connect programmatically. Reflecting these changes in code as well as docs.

Username/password is now legacy, working only for customers on the old version of Firebolt, who are gradually being migrated. `firebolt-sdk` handles connection to the right version, but needs the correct credentials wrapper, which is determined based on the presence of @ symbol in the email (service account id is alpha-numeric only).

CUBEJS_DB_USER/PASS environment variables are used to be backwards-compatible with the connections setup on the old versions of the connector. 

CUBEJS_FIREBOLT_ACCOUNT is required for the new authentication to succeed.


ptiurin:
Credentials for testing you already poses should work in legacy mode, however a new set has been setup so service account auth can be tested. Let me know how can I send you the new credentials so you can update them in the CI.

paveltiunov:
@ptiurin Could you please fix the tests?

Issue #8244, opened 2024-05-08
Documentation for MSSQL custom CA certificates is incorrect / incomplete


sjauld:
**Describe the bug**
You cannot connect to an MSSQL database over SSL using a self-signed CA by following the instructions in the documentation.

**To Reproduce**
1. Read the docs
2. Follow the docs
3. Attempt to connect to an MSSQL database
4. Fail

**Expected behavior**
Following the documentation should enable me to connect.

**Screenshots**
N/A.

**Minimally reproducible Cube Schema**
N/A.

**Version:**
v0.35.29

**Additional context**
PR to follow

PR #8245, opened 2024-05-08
Clarify instructions for using custom CA bundles for MSSQL connections

sjauld:
**Check List**
- [ ] ~Tests has been run in packages where changes made if available~
- [ ] ~Linter has been run for changed code~
- [ ] ~Tests for the changes have been added if not covered yet~

#8244 

Just an update to the documentation to save others time

### DIFF
--- a/docs/pages/product/configuration/data-sources/ms-sql.mdx
+++ b/docs/pages/product/configuration/data-sources/ms-sql.mdx
@@ -78,9 +78,14 @@ Microsoft SQL Server does not support export buckets.
 ## SSL
 
 To enable SSL-encrypted connections between Cube and Microsoft SQL Server, set
-the `CUBEJS_DB_SSL` environment variable to `true`. For more information on how
-to configure custom certificates, please check out [Enable SSL Connections to
-the Database][ref-recipe-enable-ssl].
+the `CUBEJS_DB_SSL` environment variable to `true`.
+
+To use a custom (or self-signed) certificate chain, set the `NODE_EXTRA_CA_CERTS`
+environment variable to `/path/to/certificate/bundle.cert`.
+
+Please note that the configuration settings on the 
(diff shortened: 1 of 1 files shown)


jufemaiz (approved):

igorlukanin:
Hi @sjauld 👋 

Thanks for contributing this! I think I have a couple questions here.

Let me present my thinking first:
- By looking at Cube's MS SQL [driver code](https://github.com/cube-js/cube/blob/master/packages/cubejs-mssql-driver/driver/MSSqlDriver.js#L66-L78), I see that the `encrypt` option is passed to the underlying driver (value is taken from `CUBEJS_DB_SSL` env var).
- In the [underlying driver docs](https://github.com/tediousjs/node-mssql?tab=readme-ov-file#longer-example-connect-via-config-object), I see that both `encrypt` and `trustServerCertificate` are required for self-signed certificates to work.
- So, I assume that the docs' suggestion to use `CUBEJS_DB_SSL` is correct. The only caveat is that, in case of a self-signed certificate, `trustServerCertificate` should also be passed.

What do you think? Does my line of thought seems to be correct?

Also, if you instantiate your MS SQL driver in the following way, are you able to connect? It would be fantastic if you can test the following setup and report back if it worked for you:

In `cube.js` file:
```js
module.exports = {
  driverFactory: ({ securityContext, dataSource }) => {
    return {
      type: 'mssql',
      host: '...',
      user: '...',
      password: '...',
      database: '...',
      options: {
        encrypt: true,
        trustServerCertificate: true
      }
    }
  }
};
```

Thank you in advance!

sjauld:
Hi Igor,

I'm not really interested in blindly trusting the server certificate, I want to verify the chain, hence including the bundle in my container. 

That said, I'm 99% sure that your suggestion will work.

Cheers,
Stuart

PR #8252, opened 2024-05-09
Warn in rolling window pages that Redshift does not support month or year intervals

morgan-at-cube:
Add warning box to two pages: Please note that Redshift does not support intervals of type month or year.

Context: https://cubedevinc.slack.com/archives/C021C7NT7RT/p1715293531642229

**Check List**

[For example #12]

[Description goes here]

### DIFF
--- a/docs/pages/guides/recipes/data-modeling/period-over-period.mdx
+++ b/docs/pages/guides/recipes/data-modeling/period-over-period.mdx
@@ -17,6 +17,12 @@ _previous period_.
 these `rolling_window` measures and uses them in a calculation, e.g.,
 divides or subtracts them.
 
+<WarningBox>
+
+Please note that Redshift does not support intervals of type month or year.
+
+</WarningBox>
+
 The following data model allows to calculate a month-over-month change of
 some value. `current_month_sum` and `previous_month_sum` measures define
 two rolling windows and the `month_over_month_ratio` measure divides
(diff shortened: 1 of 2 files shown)


docs/pages/guides/recipes/data-modeling/period-over-period.mdx
```diff
@@ -17,6 +17,12 @@ _previous period_.
 these `rolling_window` measures and uses them in a calculation, e.g.,
 divides or subtracts them.
 
+<WarningBox>
+
+Please note that Redshift does not support intervals of type month or year.
+
+</WarningBox>
+
```
igorlukanin, line 25: @morgan-at-cube I don't think we need to put this warning on this particular page. If we do, then we'd probably need to have it at every page mentioning `rolling_window` which is probably not the way to go.

docs/pages/reference/data-model/measures.mdx
```diff
@@ -257,6 +257,12 @@ with a defined date range.
 
 </WarningBox>
 
+<WarningBox>
+
+Please note that Redshift does not support intervals of type month or year.
+
+</WarningBox>
+
```
igorlukanin, line 265: @morgan-at-cube Do we know how it breaks? E.g., what is the exact error message?

I think we can probably do the following:
* Add the error message and the note on the Redshift data source page.
* Have a more generic note on this page, e.g., "There are known limitations for `rolling_window` support when using Redhsift (link to the data source page)". It's gonna be more extendable this way.

morgan-at-cube: Hey @igorlukanin , getting back to you about this has been in my to-do list for ages, sorry about that! I don't have the error message unfortunately :( Do you think we can still add a generic error message to this page anyway?

Issue #8268, opened 2024-05-15
Orchestration API on multi datasource using improper driver


pauldheinrichs:
**Describe the bug**
When attempting to run the orchestration on a druid datasource to rebuild pre-aggregations. The following error is spit out in the API logs after running. 

(IF i'm just doing it wrong... let me know 😓 )

```
{
  "action": "post",
      "selector": {
      "contexts": [{ "securityContext": {} }],
      "timezones": ["UTC"],
        "datasources": ["druid"]
    }
}
```

```
{"message":"Downloading external pre-aggregation via query error","queryKey":
[...]

PR #8279, opened 2024-05-19
fix(schema-compiler) Resolve rollup join pre-aggregation param allocator issue

pauldheinrichs:
**Check List**

Resolves: https://github.com/cube-js/cube/issues/8268

Resolves:
  - Pre aggregation rollup_join query invalidations for cross datasource utilizing different parameter symbols. 
  - IE: Oracle, Postgres, Druid, MSSQL

---
Summary:

So this is a fun one, here's the situation. At first i thought this was entirely community druid problem but tracked it down to an underlying issue with rollup_joins.

If we rollup_join two datasources that contain a different paramAllocator symbol (IE: `$1` and `?`) the default paramAllocator would inject the wrong parameters into the pre-aggregation sql code invalidating the cache on the API node and causing weird behaviour. This would only present itself when actually querying the data and thus causing issues. 

---

Smoke tested: 
 - Refresh schedules refreshing pre-aggregations
 - Generated pre-aggregations

Thanks ❤️

### DIFF
--- a/packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
+++ b/packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
@@ -2651,7 +2651,7 @@ export class BaseQuery {
 
   newSubQuery(options) {
     const QueryClass = this.constructor;
-    return new QueryClass(this.compilers, this.subQueryOptions({...options, paramAllocator: undefined }));
+    return new QueryClass(this.compilers, { ...this.subQueryOptions(options), paramAllocator: undefined });
   }
 
   newSubQueryForCube(cube, options) {
--- a/packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
+++ b/packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
@@ -2651,7 +2651,7 @@ export class BaseQuery {
 
   newSubQuery(options) {
     const QueryClass = this.constructor;
-    return new QueryClass(this.compilers, { ...this.subQueryOptions(options), paramAllocator: undefined });
+    return new QueryClass(this.compilers, this.subQueryOptions({...options, paramAllocator: undefined }));
   }
 
   newSubQueryForCube(cube, options) {
--- a/packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
+++ b/packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
@@ -2655,9 +2655,14 @@ export class BaseQuery {
   }
 
   newSubQueryForCube(cube, options) {
-    return this.options.queryFactory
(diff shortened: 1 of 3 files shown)


packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
```diff
@@ -2655,9 +2655,14 @@ export class BaseQuery {
...
   newSubQueryForCube(cube, options) {
-    return this.options.queryFactory
-      ? this.options.queryFactory.createQuery(cube, this.compilers, this.subQueryOptions(options))
-      : this.newSubQuery(options);
+    if (this.options.queryFactory) {
+      // When dealing with rollup joins, it's crucial to use the correct parameter allocator for the specific cube in use.
+      // By default, we'll use BaseQuery, but it's important to note that different databases (Oracle, PostgreSQL, MySQL, Druid, etc.)
+      // have unique parameter allocator symbols. Using the wrong allocator can break the query, especially when rollup joins involve
+      // different cubes that require different allocators.
+      return this.options.queryFactory.createQuery(cube, this.compilers, { ...this.subQueryOptions(options), paramAllocator: undefined });
```
pauldheinrichs, line 2663: Unset the paramAllocator set bysubQueryPtions to allow QueryClasses to assume default values

paveltiunov:
@pauldheinrichs Thanks for contributing! Could you please set the allocator to undefined in the branch without queryFactory as well? It should be consistent.

pauldheinrichs:
> @pauldheinrichs Thanks for contributing! Could you please set the allocator to undefined in the branch without queryFactory as well? It should be consistent.

For sure, i've updated the `newSubQuery` method to behave the same!

pauldheinrichs:
@paveltiunov sorry for the ping - curious if there's any other adjustments needed from my end here to push this forward!

packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
```diff
@@ -2651,13 +2651,18 @@ export class BaseQuery {
 
   newSubQuery(options) {
     const QueryClass = this.constructor;
-    return new QueryClass(this.compilers, this.subQueryOptions(options));
+    return new QueryClass(this.compilers, { ...this.subQueryOptions(options), paramAllocator: undefined });
```
paveltiunov, line 2654: We can't unset param allocator here as it should reused in case of actual `sub_query: true` members.

pauldheinrichs: What would you recommend here then, reverting the last commit and pushing forward with just the initial implementation?

paveltiunov:
Closing as it was fixed in #8619

PR #8300, opened 2024-05-27
docs: MDX API

MikeNitsenko:


docs/pages/product/apis-integrations/_meta.js
```diff
@@ -2,6 +2,7 @@ module.exports = {
   "sql-api": "SQL API",
   "rest-api": "REST API",
   "graphql-api": "GraphQL API",
+  "mdx-api": "MDX API",
```
igorlukanin, line 5: Following [this discussion](https://cubedevinc.slack.com/archives/C019TT9S57W/p1717414449720169), it would be great to also include the MDX API to the API overview page.

docs/pages/product/apis-integrations/mdx-api.mdx
```diff
@@ -0,0 +1,76 @@
...
+The MDX (MultiDimensional eXpressions) API provides a robust interface for connecting desktop applications, such as Microsoft Excel, to the Cube Cloud semantic layer via the XMLA (XML for Analysis) standard. 
+
+Key Features:
+
+- <b>Direct connectivity:</b> Connect Excel and other compatible applications directly to Cube Cloud using standard XMLA protocols.
+- <b>Advanced analytical functions:</b> Utilize the power of MDX to execute sophisticated queries that include slicing, dicing, drilling down, and rolling up of data.
+- <b>Real-time access:</b> Fetch live data from Cube Cloud, ensuring that your analyses and reports always reflect the most current information.
+
+<WarningBox>
+  The AI API is available on [Cube
```
igorlukanin, line 12: I believe there's a typo here.

docs/pages/product/apis-integrations/mdx-api.mdx
```diff
@@ -0,0 +1,76 @@
+# MDX API
+
+The MDX (MultiDimensional eXpressions) API provides a robust interface for connecting desktop applications, such as Microsoft Excel, to the Cube Cloud semantic layer via the XMLA (XML for Analysis) standard.
```
igorlukanin, line 3: - Do we know any application, other than Microsoft Excel, that can be used with the MDX API? If we do, let's name it. If we don't, let's remove the mention of it. As a rule of thumb, it's great to be more specific than vague in the docs.
- Can we add links to some neutral/authoritative pages for MDX and XMLA? Wikipedia or some Microsoft resources would be great.

docs/pages/product/apis-integrations/mdx-api.mdx
```diff
@@ -0,0 +1,76 @@
...
+
+<InfoBox>
+  Please note that only the [views](/reference/data-model/view) will be available for connections via MDX API
+</InfoBox>
+
+To use the MDX API, set up one or more views before getting started.
+
+#### Dimension hierarchies
+
+MDX API supports dimension hierarchies. You can define multiple hierarchies in the view's schema file.
```
igorlukanin, line 38: e don't use the word "schema" to mean "data model" anymore.

docs/pages/product/apis-integrations/mdx-api.mdx
```diff
@@ -0,0 +1,76 @@
...
+
+### Data modeling
+
+<InfoBox>
+  Please note that only the [views](/reference/data-model/view) will be available for connections via MDX API
+</InfoBox>
+
+To use the MDX API, set up one or more views before getting started.
+
+#### Dimension hierarchies
```
igorlukanin, line 36: I'm not sure this should be part of the MDX API docs. It would be great to document this in the data modeling section and provide a link pointing there.

docs/pages/product/apis-integrations/mdx-api.mdx
```diff
@@ -0,0 +1,76 @@
...
+    meta:
+      hierarchies:
+        - name: "Geography"
+          levels:
+            - country
+            - state
+            - city
+```
+
+#### Measure groups
```
igorlukanin, line 54: I'm not sure this should be part of the MDX API docs. It would be great to document this in the data modeling section and provide a link pointing there.

docs/pages/product/apis-integrations/mdx-api.mdx
igorlukanin: It would also be great, just like on REST API and SQL API pages, to reference a use case, explain query format, show an example request. Also, since Microsoft Excel is mentioned here, and it was never supported before, it would be great to add a page for Microsoft Excel, similarly to what we already have for [Power BI](https://cube.dev/docs/product/configuration/visualization-tools/powerbi).

keydunov (approved):

PR #8325, opened 2024-06-05
feat(cubesql): support `GREATEST`/`LEAST` function for scalars

KSDaemon:
**Check List**

This PR includes a few related features:

* Support `GREATEST` SQL function with an arbitrary number of arguments
* Support `LEAST` SQL function with an arbitrary number of arguments (previously it supported only 2 args)
* Allow float scalars in LEAST/GREATEST functions
* Support GREATEST/LEAST functions in drivers/dialects that follow the `ignore null` approach (see the table below)

The SQL standard requires greatest and least to return null in case one argument is null. This is different from aggregate functions, in particular min and max. While min and max ignore null values and return the smallest or largest value from the remaining set, greatest and least respect null values in so far as the presence of a single null value means that the highest or lowest value cannot be determined without doubt.

Here is how different databases treat NULLs in these functions:

| DB         | Respect null | Ignore null |
|------------|--------------|-------------|
| BigQuery   | *            |             |
| RedShift   |              | *           |
| SnowFlake  | *            |             |
| Databricks |              | *           |
| Athena (Trino)     | *            |             |
| DB2        | *            |             |
| H2         | *            |             |
| MariaDB    | *            |             |
| MySQL      | *            |             |
| Oracle DB  | *            |             |
| PostgreSQL |              | *           |
| SQL Server |              | *           |
| ClickHouse  | *            |             |
| DuckDB |              | *           |
| Yellowbrick |              | *           |
| PrestoDB  | *            |             |
| Vertica  | *            |             |

However, in Postgres NULL values in the list are ignored. We follow Postgres: The result will be NULL only if all the expressions evaluate to NULL.


rust/cubesql/cubesql/src/compile/engine/udf/common.rs
```diff
@@ -484,83 +484,215 @@ pub fn create_if_udf() -> ScalarUDF {
...
-                    )));
+                            current_max.value(0) > casted_arg.value(0)
+                        }
+                        DataType::Int64 => {
+                            let current_max =
+                                downcast_primitive_arg!(current_max, "current_max", Int64Type);
+                            let casted_arg =
+                                downcast_primitive_arg!(casted_arg, "casted_arg", Int64Type);
+
+                            current_max.value(0) > casted_arg.value(0)
```
ovr, line 534: This function will not work with tables. You can test it with something like

```sql
SELECT least(t.a, t.b, t.c) FROM (
    SELECT 1 as a, 2 as b, 3 as c UNION ALL
    SELECT 3, 2, 1 UNION ALL
) as t
```

rust/cubesql/cubesql/src/compile/engine/udf/common.rs
```diff
@@ -484,83 +484,215 @@ pub fn create_if_udf() -> ScalarUDF {
...
+            max_value = match max_value {
+                Some(current_max) => {
+                    let is_greater = match base_type {
+                        DataType::UInt64 => {
+                            let current_max =
+                                downcast_primitive_arg!(current_max, "current_max", UInt64Type);
+                            let casted_arg =
+                                downcast_primitive_arg!(casted_arg, "casted_arg", UInt64Type);
+
+                            current_max.value(0) < casted_arg.value(0)
```
ovr, line 633: This function will not work with tables. You can test it with something like

```sql
SELECT least(t.a, t.b, t.c) FROM (
    SELECT 1 as a, 2 as b, 3 as c UNION ALL
    SELECT 3, 2, 1 UNION ALL
) as t
```

rust/cubesql/cubesql/src/compile/engine/udf/common.rs
KSDaemon: Yep! Got it! Thnx for pointing! I'll fix it!

rust/cubesql/cubesql/src/compile/engine/udf/common.rs
KSDaemon: Done!

MazterQyou (approved):

ovr (approved):

PR #8389, opened 2024-06-25
fix: move from/to parsing to the very top

tsaplev:
According to the documentation, there's an option to use [the syntax "from ... to ..." for dateRange](https://cube.dev/docs/product/apis-integrations/rest-api/query-format#relative-date-range), which uses the [Chrono library](https://github.com/wanasit/chrono) under the hood.

My intention is to get a relative dateRange for the entire week a week ago. The prompt I use is `from (14 days before this week) to (7 days before this week)`. If I put the prompt into Chrono standalone, it works perfectly and returns the dates I expect. However, Cube processes it differently — it only returns the date range for "last week" and ignores the rest of the prompt.

I investigated the source code and realised this issue occurs because of the order of conditions in `[dateParser.js](https://github.com/cube-js/cube/blob/master/packages/cubejs-api-gateway/src/dateParser.js#L67)`. If a prompt contains "this week", "this month", "this day", etc., only this part will be handled, and the rest will be ignored. My proposal in this PR is to move the "from / to" condition to the very top, ensuring that 'from / to' will also be handled if they contain "this week", "this month", "this day", etc.

### DIFF
--- a/packages/cubejs-api-gateway/src/dateParser.js
+++ b/packages/cubejs-api-gateway/src/dateParser.js
@@ -21,7 +21,30 @@ export function dateParser(dateString, timezone, now = new Date()) {
   let momentRange;
   dateString = dateString.toLowerCase();
 
-  if (dateString.match(/(this|last|next)\s+(day|week|month|year|quarter|hour|minute|second)/)) {
+  if (dateString.match(/^from (.*) to (.*)$/)) {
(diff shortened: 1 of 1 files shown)


igorlukanin:
Thanks for your PR @tsaplev 🙌 

While this change generally makes sense to me, I believe applying it will be a breaking change since other Cube users might already depend on the particular order of relative date rules resolution. So, we would need to do at least a minor release if we merge this changes. I\ll let @paveltiunov to chime in here as well.

PR #8402, opened 2024-06-26
feat: Make graceful shutdown add fatal messages in postgres

srh:
Aside from fatal error message injection, we also add logic to wait on and properly join the postgres connections and tasks.  Otherwise we would not properly gracefully shut down.

One easily reversible decision made was to avoid passing the CancellationToken down recursively and instead track partially written socket buffers.  Because we have other CancellationTokens floating around in query logic (which IIRC are not actually used -- nothing triggers them) that might have a reason to change in the future.

### DIFF
--- a/packages/cubejs-api-gateway/src/sql-server.ts
+++ b/packages/cubejs-api-gateway/src/sql-server.ts
@@ -1,6 +1,7 @@
 import {
   setupLogger,
   registerInterface,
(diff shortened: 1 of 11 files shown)


ovr (approved):
LGTM 🚀

paveltiunov (approved):

PR #8411, opened 2024-07-01
feat: Smart and Semi-Fast shutdown

srh:
"Smart" matches the behavior of Postgres -- it stops accepting connections but lets them keep running -- and right now it's not possible.  "Semi-Fast" is willing to end a connection once it has nothing active going on and is triggered by SIGTERM (when graceful shutdown is configured to be enabled).

### DIFF
--- a/packages/cubejs-api-gateway/src/sql-server.ts
+++ b/packages/cubejs-api-gateway/src/sql-server.ts
@@ -7,6 +7,7 @@ import {
   Request as NativeRequest,
   LoadRequestMeta,
 } from '@cubejs-backend/native';
+import type { ShutdownMode } from '@cubejs-backend/native';
 import { displayCLIWarning, getEnv } from '@cubejs-backend/shared';
 
 import * as crypto from 'crypto';
@@ -332,7 +333,7 @@ export class SQLServer {
     // @todo Implement
   }
 
-  public async shutdown(): Promise<void> {
-    await shutdownInterface(this.sqlInterfaceInstance!);
+  public async shutdown(mode: ShutdownMode): Promise<void> {
+    await shutdownInterface(this.sqlInterfaceInstance!, mode);
   }
 }
--- a/rust/cubesql/cubesql/src/sql/postgres/service.rs
+++ b/rust/cubesql/cubesql/src/sql/postgres/service.rs
@@ -65,7 +65,7 @@ impl ProcessingLoop for PostgresServer {
                                 break;
                             }
                             None => {
-                                panic!("Unreachable None case greater than some other value");
+                                unreachable!("mode compared greater than something; it can't be None");
                             }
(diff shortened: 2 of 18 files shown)


KSDaemon (commented):
LGTM :) 
Posted one question/comment.

packages/cubejs-api-gateway/src/sql-server.ts
```diff
@@ -332,7 +333,7 @@ export class SQLServer {
     // @todo Implement
   }
 
-  public async shutdown(): Promise<void> {
-    await shutdownInterface(this.sqlInterfaceInstance!);
+  public async shutdown(mode: ShutdownMode): Promise<void> {
```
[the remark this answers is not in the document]
srh: I don't think it does because there are only a few callers, and I don't have any reason to want a zero-argument `shutdown()` interface to be available.

rust/cubesql/cubesql/src/sql/postgres/service.rs
```diff
@@ -34,19 +34,40 @@ impl ProcessingLoop for PostgresServer {
...
+
+                                semifast_shutdown_interruptor.cancel();
+                                break;
+                            }
+                            Some(ShutdownMode::Smart) => {
+                                trace!("[pg] Stopping processing_loop via interruptor, smart mode");
+                                break;
+                            }
+                            None => {
+                                panic!("Unreachable None case greater than some other value");
```
ovr, line 68: WDYT about using `unreachable!` ?)

rust/cubesql/cubesql/src/sql/postgres/service.rs
```diff
@@ -121,24 +144,64 @@ impl ProcessingLoop for PostgresServer {
...
+                            }
+                            Some(ShutdownMode::SemiFast) => {
+                                trace!("[pg] Stopping processing_loop via channel: upgrading to semifast mode");
+
+                                semifast_shutdown_interruptor.cancel();
+                            }
+                            _ => {
+                                // Because of comparisons made, the smallest and 2nd smallest
+                                // Option<ShutdownMode> values are impossible.
+                                panic!("Unreachable mode={:?} case", active_shutdown_mode);
```
ovr, line 175: WDYT about `unreachable!` ?

rust/cubesql/cubesql/src/sql/postgres/service.rs
srh: n00b moment; replaced the two panics with unreachable.

rust/cubesql/cubesql/src/sql/postgres/service.rs
srh: fixed

paveltiunov (approved):

Issue #7062, opened 2023-08-23
FILTER_PARAMS with views


danielli-ziprecruiter:
**Problem**
Is this the intended behavior when applying `FILTER_PARAMS` with views?
If I use `FILTER_PARAMS` in a measure:
```
cubes:
  - name: my_cube
    measures: 
      - name: my_measure
        type: count
        filters:
          - sql: {FILTER_PARAMS.my_cube.my_dimension.filter(filter)}
```
with a view that uses the cube:
```
views:
  - name: my_view
    cubes:
      - join_path: my_cube
        includes: "*"
```
The `FILTER_PARAMS` doesn't apply if I query the view with the SQL API:
```
SELECT count(my_measure)
FROM my_view
WHERE my_dimension = x
```
since this references `my_view.my_dimension`, not `my_cube.my_dimension`. 

Is this the intended behavior? There are some workarounds like adding the view's dimension as a FILTER_PARAM:
```
sql: {FILTER_PARAMS.my_cube.my_dimension.filter(filter)} 
  and {FILTER_PARAMS.my_view.my_dimension.filter(filter)}
```
Or joining with the cube so the cube's dimension can be referenced:
```
SELECT count(my_measure)
FROM my_view CROSS JOIN my_cube
WHERE my_cube.my_dimension = x
```
But these seem like hacky solutions.

paveltiunov:
@danielli-ziprecruiter That's a good one! So it's actually right behavior by design, but I agree it might be suboptimal in some cases. The obvious intuition is to always use leaf members for `FILTER_PARAMS` but it'll most likely break existing calculated members. So we'd need to think about trade-offs here.

igorlukanin:
Hi @danielli-ziprecruiter 👋 Could you please try [v0.35.80](https://github.com/cube-js/cube/commit/95beaa43e049c808c859cdab92e10e992ae49934)? The issue should be resolved in this version.

igorlukanin:
I believe this one is resolved.

PR #8466, opened 2024-07-15
fix(schema-compiler): propagate FILTER_PARAMS from view to inner cube's SELECT

KSDaemon:
Fix with propagating FILTER_PARAMS from views to inner Cube's SELECT statements.

**Check List**

### DIFF
--- a/packages/cubejs-schema-compiler/src/adapter/BaseGroupFilter.ts
+++ b/packages/cubejs-schema-compiler/src/adapter/BaseGroupFilter.ts
@@ -1,5 +1,3 @@
-import R from 'ramda';
-
 export class BaseGroupFilter {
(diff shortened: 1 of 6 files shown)


igorlukanin:
@KSDaemon Will it close #7062?

KSDaemon:
@igorlukanin Probably yes :)

packages/cubejs-schema-compiler/src/adapter/BaseGroupFilter.ts
```diff
@@ -1,5 +1,3 @@
-import R from 'ramda';
```
ovr, line 1: ❤️

paveltiunov (approved):

PR #8511, opened 2024-07-24
feat(native): Add datetime types support to generate_series()

KSDaemon:
This PR supports date and timestamp types in `generate_series()` UDTF.
So now you can issue SQL like this:
```sql
SELECT generate_series('2024-07-23'::date, '2024-07-28'::date); -- 1 day will be default step
SELECT generate_series('2016-07-23'::date, '2024-07-28'::date, '3 years'::interval);
SELECT generate_series('2014-07-23 00:00:00'::timestamp, '2024-10-28 00:00:00'::timestamp, '2 years'::interval);
SELECT generate_series('2010-01-01 00:00:00'::timestamp, '2024-07-25 00:01:30'::timestamp, '6 years 5 months 4 days 3 hours 2 minutes 1 seconds'::interval);
```

**Check List**


rust/cubesql/cubesql/src/compile/engine/udf/common.rs
```diff
@@ -2313,6 +2313,78 @@ macro_rules! generate_series_udtf {
...
+                                let current_dt = NaiveDateTime::from_timestamp_opt(
+                                    (current as i64) / 1_000_000_000,
+                                    (current % 1_000_000_000) as u32,
+                                )
+                                .unwrap();
+                                let res =
+                                    date_addsub_month_day_nano(current_dt, step, true).unwrap();
+                                current = res.timestamp_nanos_opt().unwrap() as $PRIMITIVE_TYPE;
+                            } else {
+                                panic!("Unsupported type");
```
ovr, line 2375: When something can fail, it will fail.

1. When the error depends on data (input, arguments, options). it's important to write type of it.
2. When you can return error, then you should return error. Please replace panic with error.

Ty

rust/cubesql/cubesql/src/compile/engine/udf/common.rs
```diff
@@ -2313,6 +2313,78 @@ macro_rules! generate_series_udtf {
...
+                        let mut current = start;
+                        loop {
+                            if current > end {
+                                break;
+                            }
+                            builder.append_value(current).unwrap();
+
+                            section_size += 1;
+
+                            if $ARGS[0].as_any().downcast_ref::<Date32Array>().is_some() {
```
ovr, line 2354: This macro will expand 3 times. These checks will be in all the codes. it's better to pass block as macros's argument in the calls.

rust/cubesql/cubesql/src/compile/engine/udf/common.rs
KSDaemon: Good catch! Sorry, there was error here of course, missed during debugging %)

rust/cubesql/cubesql/src/compile/engine/udf/common.rs
KSDaemon: Yeah! Good idea! Thnx!

rust/cubesql/cubesql/src/compile/engine/udf/common.rs
```diff
@@ -2313,6 +2313,80 @@ macro_rules! generate_series_udtf {
...
+                                let current_dt =
+                                    NaiveDateTime::from_timestamp_opt((current as i64) * 86400, 0)
+                                        .unwrap();
+                                let res =
+                                    date_addsub_month_day_nano(current_dt, step, true).unwrap();
```
ovr, line 2359: Unwrap -> err

rust/cubesql/cubesql/src/compile/engine/udf/common.rs
```diff
@@ -2313,6 +2313,80 @@ macro_rules! generate_series_udtf {
...
+                                .downcast_ref::<TimestampNanosecondArray>()
+                                .is_some()
+                            {
+                                let current_dt = NaiveDateTime::from_timestamp_opt(
+                                    (current as i64) / 1_000_000_000,
+                                    (current % 1_000_000_000) as u32,
+                                )
+                                .unwrap();
+                                let res =
+                                    date_addsub_month_day_nano(current_dt, step, true).unwrap();
```
ovr, line 2372: Unwrap -> err

KSDaemon: I think this is a good to go:
```rust
let res = date_addsub_month_day_nano(current_dt, step, true)?;
```

rust/cubesql/cubesql/src/compile/engine/udf/common.rs
KSDaemon: Rewrote with few macros without if-else-checks.

ovr (approved):

rust/cubesql/cubesql/src/compile/engine/udf/common.rs
```diff
@@ -2313,6 +2313,75 @@ macro_rules! generate_series_udtf {
     }};
 }
 
+macro_rules! generate_series_helper_date32 {
+    ($CURRENT:ident, $STEP:ident, $PRIMITIVE_TYPE: ident) => {
+        let current_dt = NaiveDateTime::from_timestamp_opt(($CURRENT as i64) * 86400, 0).unwrap();
```
ovr, line 2318: Could we make this unwrap as error?

rust/cubesql/cubesql/src/compile/engine/udf/common.rs
```diff
@@ -2313,6 +2313,75 @@ macro_rules! generate_series_udtf {
...
+    };
+}
+
+macro_rules! generate_series_helper_timestamp {
+    ($CURRENT:ident, $STEP:ident, $PRIMITIVE_TYPE: ident) => {
+        let current_dt = NaiveDateTime::from_timestamp_opt(
+            ($CURRENT as i64) / 1_000_000_000,
+            ($CURRENT % 1_000_000_000) as u32,
+        )
+        .unwrap();
```
ovr, line 2330: Could we make this unwrap as error?

rust/cubesql/cubesql/src/compile/engine/udf/common.rs
KSDaemon: Yes, we could :)

KSDaemon: Done!

PR #8534, opened 2024-08-02
Make split rules for scalar functions variadic

mcheshkov:
**Check List**

Add separate split rules for handling `ScalarFunctionExprArgs` of any size. This allows simple pass-through for any scalar function. E.g. `SUBSTR` can now push down dimension from any argument, not only from first one.

### DIFF
--- a/rust/cubesql/cubesql/src/compile/rewrite/rules/split/functions.rs
+++ b/rust/cubesql/cubesql/src/compile/rewrite/rules/split/functions.rs
@@ -22,12 +22,22 @@ impl SplitRules {
             rules,
         );
 
-        self.scalar_fn_args_pass_through_rules("Trunc", false, rules);
-        self.scalar_fn_args_pass_through_rules("Lower", false, rules);
-        self.scalar_fn_args_pass_through_rules("Upper", false, rules);
-        self.scalar_fn_args_pass_through_rules("Ceil", false, rules);
-        self.scalar_fn_args_pass_through_rules("Floor", false, rules);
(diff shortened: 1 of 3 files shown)


rust/cubesql/cubesql/src/compile/rewrite/rules/split/functions.rs
```diff
@@ -9,88 +9,52 @@ impl SplitRules {
...
+            ("Trunc", false),
+            ("Lower", false),
+            ("Upper", false),
+            ("Ceil", false),
+            ("Floor", false),
+            ("CharacterLength", false),
+            ("Substr", false),
+            ("Lpad", false),
+            ("Rpad", false),
+        ];
```
KSDaemon, line 35: Are there any functions with `true` projection on the radars? If no, maybe simplify this to just func names and pass always `false` in loop below?

KSDaemon (approved):
👍🏻 LGTM. Left one minor note.

rust/cubesql/cubesql/src/compile/rewrite/rules/split/functions.rs
mcheshkov: There's `coalesce` and `nullif`, they have projections enabled ATM, but I plan to disable projections for them as well.
In general there's stuff like `to_hex(integer)` or ` chr(integer)` that, I think, can have projections enabled. Are they on the radars? I have no idea :)

paveltiunov (approved):

Issue #8488, opened 2024-07-17
Fiscal and custom granularity support


igorlukanin:
Cube supports pre-aggregating data by a [time dimension](https://cube.dev/docs/reference/data-model/pre-aggregations#time_dimension) with respect to a chosen [granularity](https://cube.dev/docs/reference/data-model/pre-aggregations#granularity). Currently, the following granularities are supported: `second`, `minute`, `hour`, `day`, `week`, `month`, `quarter`, and `year`.

We would like to make granularities configurable, including the support for such granularities like "fiscal quarter" (not necessarily aligned with the start of a year).

It's currently planned for October 2024.

Related issues:
- #177

igorlukanin:
Documentation:
- [Time dimensions](https://cube.dev/docs/product/data-modeling/concepts#time-dimensions) in the data modeling concepts.
- [`granularities` parameter](https://cube.dev/docs/reference/data-model/dimensions#granularities) in the data modeling reference.
- Using [proxy dimensions](https://cube.dev/docs/product/data-modeling/concepts/calculated-members#time-dimension-granularity) with time dimension granularities.
- [Implementing custom time dimension granularities](https://cube.dev/docs/guides/recipes/data-modeling/custom-granularity) recipe.

Announced in this blog post: https://cube.dev/blog/introducing-custom-time-dimension-granularities

PR #8537, opened 2024-08-05
feat(schema-compiler): custom granularity support

### LINKED ISSUE #177: Alternate week granularity to allow for week starting either Sunday OR Monday and other granularities
We work with a dataset where some customers consider a week differently, where a "week" is starting either Sunday or Monday. Cube hits most of our use cases, but the `week` granularity option conforms to the standard `week` observed by Postgres (our current adaptor). So, for some customers, we cannot use a week granularity and this breaks some reporting needs we have.

I'd like to be able to utilize a secondary `week` granularity option to opt out of the default behaviour, and roll up data to a week starting *Sunday* instead of a week starting *Monday*. I recognize that Postgres conforms to the ISO standard of a week starting on a Monday, and would want to maintain that as a the default behavior, hence adding `week_sunday` as an alternate option.

I've been able to patch in support for a week starting Sunday by hotfixing the Postgres adaptor and where Cube validates the incoming query, but I suspect (though have not tested) this will break any option to pre-aggregate.

KSDaemon:
This PR brings custom granularity support into the Cube.

Now it is possible to define any number of custom granularities for time dimension. Custom granularity maybe specified in two forms: 
* Interval + offset. You can look at this as a supercharged date_trunc version. This way you can define only intervals that are divisible by a year with no remainder (like for example`2 months`).
* Interval + origin point. This is the most flexible variant. You can choose any arbitrary interval (e.g. `2 months 3 weeks 4 days 12 hours`) and set the origin pivot point for calculating intervals. Origin maybe specified in date or timestamp if needed.

A few examples:
```yaml
.....
dimensions:
  - name: createdAt
    sql: created_at
    type: time
    granularities:
      - name: six_months
        interval: 6 months # no offset specified, meaning the start of the year
      - name: three_months_offset
        interval: 3 months
        offset: 2 weeks # the pivot point in this case will be 15 Jan
      - name: fiscal_year_1st_april
        interval: 1 year
        origin: >  # Note that as date includes dashes, you need to escape it with quotes or use `>` variant
          2024-04-01
      - name: timestamp_origin_example
        interval: 24 days
        origin: "2024-02-15 10:15:25"
```

What's else inside?

* [x] Custom granularity support in YAML cube definitions
* [x] Custom granularity cube scheme validation
* [x] Custom granularity interval generation for rolling windows
* [x] Custom granularity support in BaseQuery
* [x] Custom granularity support for PostgeSQL
* [x] Custom granularity support for MySQL
* [x] Custom granularity support for BigQuery
* [x] Custom granularity support for Databricks
* [x] Custom granularity support for MS SQL Server
* [x] Custom granularity support for Snowflake
* [x] Custom granularity support for ClickHouse
* [x] Custom granularity support for DuckDB

This PR probably closes #177 and partially #8488

**Check List**


packages/cubejs-backend-shared/src/time.ts
```diff
@@ -26,10 +31,113 @@ export const TIME_SERIES: Record<string, (range: DateRange, timestampPrecision:
...
+export function parseSqlInterval(intervalStr: SqlInterval): ParsedInterval {
+  const interval: ParsedInterval = {};
+  const parts = intervalStr.split(/\s+/);
+
+  for (let i = 0; i < parts.length; i += 2) {
+    const value = parseInt(parts[i], 10);
+    const unit = parts[i + 1];
+
+    // Remove ending 's' (e.g., 'days' -> 'day')
+    const singularUnit = (unit.endsWith('s') ? unit.slice(0, -1) : unit) as unitOfTime.DurationConstructor;
```
ovr, line 49: But it doesn't validate any errors with granularities. We can make regexp more strict. For example 

`/^(-?\d+) (second|minute|hour|day|week|month|quarter|year)s?$/`

Or validate it here with string, because you are casting `string` to `unitOfTime.DurationConstructor`.

Additional: Could you add a test with incorrect input for granularity unit?

packages/cubejs-backend-shared/src/time.ts
```diff
@@ -26,10 +31,113 @@ export const TIME_SERIES: Record<string, (range: DateRange, timestampPrecision:
...
+
+export const timeSeriesFromCustomInterval = (intervalStr: string, [startStr, endStr]: QueryDateRange, origin: moment.Moment, options: TimeSeriesOptions = { timestampPrecision: 3 }): QueryDateRange[] => {
+  const intervalParsed = parseSqlInterval(intervalStr);
+  const start = moment(startStr);
+  const end = moment(endStr);
+  let alignedStart = alignToOrigin(start, intervalParsed, origin);
+
+  const dates: QueryDateRange[] = [];
+
+  while (alignedStart.isBefore(end)) {
```
ovr, line 124: I am thinking about protection for super large time series. for example, user can specify 2 seconds interval from 1970 to 2050 and it will do a CPU spike (I don't think that manipulation with dates are fast + OOM?)

I belive, we should care about it. It's vector for DDoS

packages/cubejs-backend-shared/src/time.ts
```diff
@@ -26,10 +31,113 @@ export const TIME_SERIES: Record<string, (range: DateRange, timestampPrecision:
...
+    intervalOp = isIntervalNegative ? addInterval : subtractInterval;
+
+    while (offsetDate.isAfter(startDate)) {
+      offsetDate = intervalOp(offsetDate, interval);
+    }
+    alignedDate = offsetDate;
+  } else {
+    intervalOp = isIntervalNegative ? subtractInterval : addInterval;
+
+    while (offsetDate.isBefore(startDate)) {
```
ovr, line 103: I am thinking about protection for super large time series. for example, user can specify 2 seconds interval from 1970 to 2050 and it will do a CPU spike (I don't think that manipulation with dates are fast + OOM?)

I belive, we should care about it. It's vector for DDoS

packages/cubejs-backend-shared/src/time.ts
```diff
@@ -26,10 +31,113 @@ export const TIME_SERIES: Record<string, (range: DateRange, timestampPrecision:
...
+  if (offsetDate.isBefore(origin)) {
+    isIntervalNegative = true;
+  }
+
+  offsetDate = origin.clone();
+
+  if (startDate.isBefore(origin)) {
+    intervalOp = isIntervalNegative ? addInterval : subtractInterval;
+
+    while (offsetDate.isAfter(startDate)) {
```
ovr, line 96: I am thinking about protection for super large time series. for example, user can specify 2 seconds interval from 1970 to 2050 and it will do a CPU spike (I don't think that manipulation with dates are fast + OOM?)

I belive, we should care about it. It's vector for DDoS

packages/cubejs-backend-shared/src/time.ts
KSDaemon: There is no need to validate anything here as the interval comes from the cube configuration definition that was validated during the parsing/init phase.

packages/cubejs-api-gateway/src/query.js
```diff
@@ -103,7 +103,7 @@ const querySchema = Joi.object().keys({
   filters: Joi.array().items(oneFilter, oneCondition),
   timeDimensions: Joi.array().items(Joi.object().keys({
     dimension: id.required(),
-    granularity: Joi.valid('quarter', 'day', 'month', 'year', 'week', 'hour', 'minute', 'second', null),
+    granularity: Joi.string(), // Custom granularities have arbitrary names
```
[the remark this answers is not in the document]
KSDaemon: Then we need to limit for example `timezones` of `member expressions (any kind of names of something)`

packages/cubejs-backend-shared/src/time.ts
KSDaemon: How can you distinguish between malicious and legitimate requests? I think it's not possible.

ovr: By limiting via configuration, we should throw an error when the size of the times series reaches N. сС @paveltiunov

packages/cubejs-api-gateway/src/query.js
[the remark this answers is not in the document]
ovr: @KSDaemon Do we plan to support custom granularity without defining it? For example 5 minutes?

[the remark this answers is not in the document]
KSDaemon: anyway... added some limits.

packages/cubejs-backend-shared/src/time.ts
igorlukanin: Users could use `second` or `minute` as a granularity for many years and I don't recall I single complaint about degraded performance due to this. So I'm not sure if we should have safeguards for such use cases.

I strongly feel we should not add an env var a a config option that we'll need to document, etc. If we really wanna do that, maybe it could be just a warning if the size of time series is more than N, and we just choose that N ourselves.

But I would prefer we just don't do anything here.

KSDaemon: Made a hard limit (50000) for now, and left a todo to make it configurable. This will be done in the one of upcoming PRs.

paveltiunov (approved):

PR #8552, opened 2024-08-07
fix(cubesql): Use date_part => date_part + date_trunc split only with appropriate date_part argument

mcheshkov:
**Check List**

Rules for splitting date_part calls now parses date_part argument and use it to check if split is possible, and what granularity to use in date_trunc

### Examples

`EXTRACT(EPOCH FROM dim_date0)`
**Before**
Split to `date_part('epoch', date_trunc('epoch', dim_date0))`, which is incorrect, `date_trunc` does not accept `'epoch'`, and would issue CubeScan with `"timeDimensions": [{"dimension": "dim_date0", "granularity": "epoch"}]`
**After**
Split as a simple dimension request and `date_part('epoch', ...` projection

`EXTRACT(DOW FROM dim_date0)`
**Before**
Split to `date_part('day', date_trunc('day', dim_date0))`, which is incorrect, truncation is correct, but `date_part('day', ...` is not
**After**
Split as `date_part('dow', date_trunc('day', dim_date0))`

### DIFF
--- a/rust/cubesql/cubesql/src/compile/rewrite/rules/split/dates.rs
+++ b/rust/cubesql/cubesql/src/compile/rewrite/rules/split/dates.rs
@@ -2,7 +2,11 @@ use crate::{
     compile::rewrite::{
         analysis::{ConstantFolding, LogicalPlanAnalysis},
         cast_expr, literal_expr,
-        rules::{members::min_granularity, split::SplitRules, utils::parse_granularity_string},
+        rules::{
+            members::min_granularity,
+            split::SplitRules,
+            utils::{DatePartToken, SpecialTimeUnitToken},
+        },
         LiteralExprValue, LogicalPlanLanguage,
     },
     var,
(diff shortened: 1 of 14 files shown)


rust/cubesql/cubesql/src/compile/rewrite/rules/split/dates.rs
```diff
@@ -136,19 +157,86 @@ impl SplitRules {
...
-                    subst.insert(output_granularity_var, output_granularity);
-                    return true;
+                let date_part: DatePartToken = if let Ok(date_part) = date_part.parse() {
+                    date_part
+                } else {
+                    return false;
+                };
+                if !matches!(
+                    date_part,
+                    // Julian sates are very similar to epoch, so we handle it here as well
```
KSDaemon, line 175:
```suggestion
                    // Julian dates are very similar to epoch, so we handle it here as well
```

KSDaemon (approved):
👍🏻  LGTM

rust/cubesql/cubesql/src/compile/rewrite/rules/split/dates.rs
```diff
@@ -136,19 +157,86 @@ impl SplitRules {
...
+                let date_part: DatePartToken = if let Ok(date_part) = date_part.parse() {
+                    date_part
+                } else {
+                    return false;
+                };
+                let trunc_granularity = if let Some(token) = date_part.delta_for_trunc() {
+                    token
+                } else {
+                    return false;
+                };
```
MazterQyou, line 228:
```suggestion
                let Some(trunc_granularity) = date_part.delta_for_trunc() else {
                    return false;
                };
```

rust/cubesql/cubesql/src/compile/rewrite/rules/split/dates.rs
```diff
@@ -136,19 +157,86 @@ impl SplitRules {
...
+
+                let new_date_part = egraph.add(LogicalPlanLanguage::LiteralExprValue(
+                    LiteralExprValue(ScalarValue::Utf8(Some(date_part.as_str().to_string()))),
+                ));
```
MazterQyou, line 232: Same as above; we're copying a constant-folded scalar, so we don't need `?new_date_part` at all.

rust/cubesql/cubesql/src/compile/rewrite/rules/split/dates.rs
```diff
@@ -136,19 +157,86 @@ impl SplitRules {
...
+                    DatePartToken::Special(SpecialTimeUnitToken::Epoch)
+                        | DatePartToken::Special(SpecialTimeUnitToken::Julian)
+                ) {
+                    return false;
                 }
+
+                let new_date_part = egraph.add(LogicalPlanLanguage::LiteralExprValue(
+                    LiteralExprValue(ScalarValue::Utf8(Some(date_part.as_str().to_string()))),
+                ));
+                subst.insert(new_date_part_var, new_date_part);
```
MazterQyou, line 185: We don't actually transform `date_part` here, effectively using it as is. Considering we could get it from constantfolding, we might as well just reuse `?date_part` placeholder, we don't need `?new_date_part` at all.

MazterQyou (approved):

rust/cubesql/cubesql/src/compile/rewrite/rules/split/dates.rs
mcheshkov: We are changing it: parsing can handle synonyms like `qtr` for `quarter`, but `as_str` always return “standard” representation (I didn’t actually check standard text, only PostgreSQL docs)

rust/cubesql/cubesql/src/compile/rewrite/rules/split/dates.rs
mcheshkov: let-else bindings! Of course! I’ll  fix it.

rust/cubesql/cubesql/src/compile/rewrite/rules/split/dates.rs
MazterQyou: Ah, I beg your pardon, I missed `date_part` being shadowed on parse. 👍

paveltiunov (approved):

PR #8556, opened 2024-08-08
BUG: Fix QueryBuilder to AVOID Maximum recursive updates exceeded

nsandroni:
fix: Maximum recursive updates exceeded

**Check List**

Passing a dynamic query to the query builder, can cause "Maximum recursive updates exceeded"

Avoiding the preValidatedQuery mutation inside the computed block, the error is no loger thronw

### DIFF
--- a/packages/cubejs-client-vue3/src/QueryBuilder.js
+++ b/packages/cubejs-client-vue3/src/QueryBuilder.js
@@ -393,7 +393,6 @@ export default {
         this.skipHeuristics = false;
       }
 
-      this.prevValidatedQuery = validatedQuery;
       return validatedQuery;
     },
   },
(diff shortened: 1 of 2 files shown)


ovr:
@nsandroni, Could you please write a unit test for it?

Thanks

nsandroni:
it turned out that not only prevValidatedQuery can cause a computed loop, but also pivotConfig and every parameter of the query saved inside copyQueryFromProps (that are used inside the validation), i added a check before the update to prevent computed loop with the same values

nsandroni:
with (from?) the version 0.35.23 the bug is not visible...

igorlukanin:
@nsandroni Do you mean that the bug does not reproduce on the latest version anymore?

nsandroni:
> @nsandroni Do you mean that the bug does not reproduce on the latest version anymore?

No, currently i'm not experiencing this bug anymore... I'm closing this PR for now, if i'll see again in the future i'll open another one

PR #8572, opened 2024-08-12
chore(cubesql): Initial support for custom protocols

ovr:


MazterQyou (approved):

rust/cubesql/cubesql/src/compile/protocol.rs
```diff
@@ -0,0 +1,95 @@
...
+
+    fn table_name_by_table_provider(
+        &self,
+        table_provider: Arc<dyn datasource::TableProvider>,
+    ) -> Result<String, CubeError>;
+}
+
+impl PartialEq for dyn DatabaseProtocolDetails {
+    fn eq(&self, other: &Self) -> bool {
+        self.get_name() == other.get_name()
```
mcheshkov, line 30: Given that we are using `'static` trait objects, why not just rely on `Any::type_id`? `TypeId` can be both hashed and compared. 
Do we expect to have multiple types with same `get_name()`, that we should treat as a same protocol?
How this would handle rogue external type with `get_name() ==  "postgres"`?

mcheshkov (commented):
Cool! I had just one question about `TypeId`.

Also now we can extract two separate ZSTs, that will `impl DatabaseProtocolDetails` for MySQL and PostgreSQL respectively, so their implementation will be coherent, and leave `DatabaseProtocol` as simple enum dispatch. And switches over `DatabaseProtocol` scattered around could be changed to just method calls.

I'm not sure about `ServerManager` lifetime, and how to approach  `DatabaseVariables` storage and  `read_variables`/`write_variables`. 

But if already feels like we could lower `postgres_default_global_variables`, `mysql_default_global_variables` and huge chunks of `QueryPlanner::set_variable_to_plan`  to `DatabaseProtocolDetails` and impls

PR #8596, opened 2024-08-19
chore(databricks-jdbc): do not add extra backticks to table name if backticks already present

MikeNitsenko:
**Check List**

[For example #12]

[Description goes here]

### DIFF
--- a/packages/cubejs-databricks-jdbc-driver/src/DatabricksDriver.ts
+++ b/packages/cubejs-databricks-jdbc-driver/src/DatabricksDriver.ts
@@ -589,6 +589,9 @@ export class DatabricksDriver extends JDBCDriver {
    * Returns quoted string.
    */
   protected quoteIdentifier(identifier: string): string {
+    if (identifier.startsWith('`') && identifier.endsWith('`')) {
+      return identifier;
+    }
     return `\`${identifier}\``;
   }
 


packages/cubejs-databricks-jdbc-driver/src/DatabricksDriver.ts
```diff
@@ -589,6 +589,9 @@ export class DatabricksDriver extends JDBCDriver {
    * Returns quoted string.
    */
   protected quoteIdentifier(identifier: string): string {
+    if (identifier.startsWith('`') && identifier.endsWith('`')) {
```
asmisha, line 592:
```suggestion
    if (identifier.startsWith('`') || identifier.endsWith('`')) {
```
I'm not sure it's very relevant, but I think in case of a malformed table name, we should not escape the name as well.

asmisha (approved):

packages/cubejs-databricks-jdbc-driver/src/DatabricksDriver.ts
MikeNitsenko: It seems like an impossible situation, and also, having backticks in Databricks' table name should not be allowed by databricks itself. So I will need further investigation

MikeNitsenko: ok, so it makes sense now: if the table name includes hyphens (for example `my-table-name`) then we add in query processing, so this fix should work

PR #8598, opened 2024-08-19
fix(cubesql): Reduce memory usage while converting to DataFrame

ovr:


rust/cubesql/cubesql/src/sql/dataframe.rs
```diff
@@ -422,9 +422,9 @@ pub fn arrow_to_column_type(arrow_type: DataType) -> Result<ColumnType, CubeErro
     }
 }
 
-pub fn batch_to_dataframe(
+pub fn batches_to_dataframe(
     schema: &Schema,
-    batches: &Vec<RecordBatch>,
+    batches: Vec<RecordBatch>,
```
ovr, line 427: Let's move control over all batches for own iterator

rust/cubesql/cubesql/src/sql/dataframe.rs
```diff
@@ -437,7 +437,7 @@ pub fn batch_to_dataframe(
         ));
     }
 
-    for batch in batches.iter() {
+    for batch in batches.into_iter() {
```
ovr, line 440: into_iter will drop every batch sequentially after convert

srh (approved):

MazterQyou (approved):

PR #8612, opened 2024-08-22
chore(cubesql): Support extra_id for Session

ovr:


MazterQyou (approved):

rust/cubesql/cubesql/src/sql/session.rs
```diff
@@ -62,6 +62,8 @@ pub enum QueryState {
 pub struct SessionState {
     // connection id, immutable
     pub connection_id: u32,
+    // Can be UUID or anything else. MDX uses UUID
+    pub extra_id: Option<String>,
```
mcheshkov, line 66: Do we want to limit `extra_id`  to UTF-8? Maybe `Vec<u8>`/`Bytes` would be better. UUID is representable with 16 raw bytes, AFAIK `uuid` crate does exactly that.

rust/cubesql/cubesql/src/sql/session.rs
```diff
@@ -447,17 +412,58 @@ pub struct SessionProcessList {
     pub database: Option<String>,
 }
 
+impl From<&Arc<Session>> for SessionProcessList {
```
mcheshkov, line 415: Do we need Arc here? It feels like `impl From<&Session>` should be enough here

rust/cubesql/cubesql/src/sql/session_manager.rs
```diff
@@ -41,60 +50,72 @@ impl SessionManager {
...
 
-        guard
-            .values()
-            .map(Session::to_stat_activity)
-            .collect::<Vec<SessionStatActivity>>()
+        Ok(session_ref)
     }
 
-    pub async fn process_list(self: &Arc<Self>) -> Vec<SessionProcessList> {
+    pub async fn map_sessions<T: for<'a> From<&'a Arc<Session>>>(self: &Arc<Self>) -> Vec<T> {
```
mcheshkov, line 90: Current `map_sessions` usages don't need `Arc` part, `From<&'a Session>` should be enough.
`Arc` here can be useful (e.g. to be able to clone `Arc` instance), but as it is now it's bit confusing to me

rust/cubesql/cubesql/src/sql/session_manager.rs
```diff
@@ -41,60 +50,72 @@ impl SessionManager {
...
-        };
-
-        let session_ref = Arc::new(sess);
+        });
 
         let mut guard = self.sessions.write().await;
 
-        guard.insert(connection_id, session_ref.clone());
+        if let Some(extra_id) = extra_id {
+            if guard.uid_to_session.contains_key(&extra_id) {
```
mcheshkov, line 75: AFAIU, `extra_id` is user-controlled, so this place can trigger hashing of strings of arbitrary length.
Do we want to limit it?

mcheshkov (commented):
LGTM, noticed some minor stuff

rust/cubesql/cubesql/src/sql/session.rs
ovr: 👍

PR #8616, opened 2024-08-22
feat(cubesql): Introduce max sessions limit

ovr:


rust/cubesql/cubesql/src/sql/session_manager.rs
```diff
@@ -71,10 +71,17 @@ impl SessionManager {
 
         let mut guard = self.sessions.write().await;
 
+        if guard.sessions.len() > self.server.config_obj.max_sessions() {
```
mcheshkov, line 74: I think it should be `>=`

mcheshkov (approved):

rust/cubesql/cubesql/src/sql/session_manager.rs
ovr: Sure, nice catch :+1:

rust/cubesql/cubesql/src/sql/postgres/service.rs
```diff
@@ -105,6 +106,22 @@ impl ProcessingLoop for PostgresServer {
                 Ok(r) => r,
                 Err(err) => {
                     error!("Session creation error: {}", err);
+
+                    let error_response = protocol::ErrorResponse::error(
+                        protocol::ErrorCode::TooManyConnections,
+                        err.to_string(),
+                    );
+
+                    if let Err(err) = buffer::write_message(
```
ovr, line 115: image

Issue #8153, opened 2024-04-15
MIN Time aggregation function represented as numeric


pauldheinrichs:
**Describe the bug**
I've noticed when attempting to utilize a time aggregation in a few different ways i'm getting odd behavrious in each scenario.

Scenario A)
```
type: min
sql: time
```
- GraphQL: represented as numeric that breaks the query
- SQL API: Represented as numeric that always results in null
- Pre aggregations work... but not in any sort of useful functionality as i can't query the measure

Okay - that's fine let's try something else

Scenario B) 
 - Let's do a bit of coercion to make things work

```
type: time
sql: MIN(time)

```
 - PreAggregations: no longer are applicable when requesting the measure in my query. 
 - Graphql: Represented as string that also returns unqueryable data
 - SQL API: represented as string (this actually works)

Tested: 
 - ON redshift driver
 - On Druid driver

No difference 

**Expected behavior**

I would anticipate in scenario 1 to be able to query the metric
in scenario 2 i would expect the pre-aggregation to apply
**Screenshots**
scenario A)
![image](https://github.com/cube-js/cube/assets/90930991/8505eb5f-2d9b-49c6-8479-58aa06bf423a)

Scenario B)
![image](https://github.com/cube-js/cube/assets/90930991/ac03d1d4-21c9-4599-b590-83bbfb88bec8)

**Minimally reproducible Cube Schema**
In case your bug report is data modelling related please put your minimally reproducible Cube Schema here.
You can use selects without tables in order to achieve that as follows.

```javascript
cube(`test`, {
  sql: `
    SELECT 1 as userid, '2021-01-01' as signupdate
    UNION ALL
    SELECT 2 as userid, '2021-01-02' as signupdate
    UNION ALL
    SELECT 3 as userid, '2021-01-03' as signupdate
    UNION ALL
    SELECT 4 as userid, '2021-01-04' as signupdate
  `,
  description: `user signups/created events from vidyard dashboard`,
  // datasource: `default`,

  measures: {
    // does not work with pre-aggregations
    min_time_broken_pre_aggs: {
      sql: `MIN(signupdate)`,
      type: `time`,
    },

    // is a numeric value unqueryable
    min_time_unqueryable: {
      sql: `signupdate`,
      type: `min`,
    },
  },
 
  dimensions: {
    userid: {
      description: `id generated when user is created`,
      sql: `userid`,
      type: `number`,
      primary_key: true,
    },

    signupdate: {
      description: `date when the user is created`,
[...]

PR #8623, opened 2024-08-23
fix(cubesql): Use proper column type for time measures

mcheshkov:
**Check List**

[Description goes here]

Use `ColumnType::Timestamp` for column for `min/max(time)` measures instead `ColumnType::Double`. Before it would try to parse strings like `"2024-12-31T00:00:00.000"` from CubeScan as `f64`.


pauldheinrichs:
Resolves this issue 🚀 
 - https://github.com/cube-js/cube/issues/8153

PR #8633, opened 2024-08-26
feat(cubesql): CubeScan - don't clone strings for non stream response

ovr:


rust/cubesql/cubesql/src/compile/engine/df/scan.rs
```diff
@@ -475,31 +478,28 @@ impl ValueObject for JsonValueObject {
...
-            .unwrap_or(&Value::Null)
-            // TODO expose strings as references to avoid clonning
-            .clone();
+
+        let value = as_object.get(field_name).unwrap_or(&Value::Null);
+
         Ok(match value {
-            Value::String(s) => FieldValue::String(s),
+            Value::String(s) => FieldValue::StringRef(s),
             Value::Number(n) => FieldValue::Number(n.as_f64().ok_or(
```
ovr, line 499: There is an additional boost here (we don't clone it), while n is used as a reference. 

image

rust/cubesql/cubesql/src/compile/engine/df/scan.rs
```diff
@@ -446,8 +446,11 @@ struct CubeScanExecutionPlan {
 }
 
 #[derive(Debug)]
-pub enum FieldValue {
+pub enum FieldValue<'a> {
+    // Neon doesn't allow us to build string reference, because V8 uses UTF-16, while Rust uses UTF-8
     String(String),
+    // Best variant for us with strings, because we dont need to clone string
+    StringRef(&'a String),
```
mcheshkov, line 453: Why not reference to `str`? I don't think reference to `String` gets us anything functionally, but ties to specific allocation pattern.

rust/cubesql/cubesql/src/compile/engine/df/scan.rs
```diff
@@ -446,8 +446,11 @@ struct CubeScanExecutionPlan {
 }
 
 #[derive(Debug)]
-pub enum FieldValue {
+pub enum FieldValue<'a> {
```
mcheshkov, line 449: Add meaningful name for lifetime, like `'str`/`'string' please

rust/cubesql/cubesql/src/compile/engine/df/scan.rs
```diff
@@ -475,31 +478,28 @@ impl ValueObject for JsonValueObject {
...
-    fn get<'a>(
-        &'a mut self,
+    fn get(
+        &mut self,
         index: usize,
         field_name: &str,
     ) -> std::result::Result<FieldValue, CubeError> {
-        let option = self.rows[index].as_object_mut();
-        let as_object = if let Some(as_object) = option {
+        let as_object = if let Some(as_object) = self.rows[index].as_object() {
```
mcheshkov, line 486: Can be represented as `let Some(as_object) = ... else {}`

mcheshkov:
Given that we now have both owned and borrowed variants in `FieldValue` maybe we should merge them to single `String` containing `Cow<'a, str>`?

rust/cubesql/cubesql/src/compile/engine/df/scan.rs
ovr: `'str` cannot be used because I combine branches via `|`. In Rust, it is a requirement to align them to the same type for usage.

rust/cubesql/cubesql/src/compile/engine/df/scan.rs
MazterQyou: I believe it's pretty much the convention to have it as `'a` at least when you only have one lifetime to care about. It's pretty easy to follow with one lifetime.

MazterQyou (approved):

rust/cubesql/cubesql/src/compile/engine/df/scan.rs
ovr: With `Cow<'str, str>` is not an option because it's meme

PR #8638, opened 2024-08-26
fix(cubesql): Don't clone response, improve performance

ovr:
serde_json::from_value requires passing the own of the value; it's why we used `clone` here. The biggest problem is that we used it for successful responses, which is a large structure.

PR fixes that issue.

### DIFF
--- a/packages/cubejs-backend-native/src/transport.rs
+++ b/packages/cubejs-backend-native/src/transport.rs
@@ -385,7 +385,7 @@ impl TransportService for NodeBridgeTransport {
             #[cfg(not(debug_assertions))]
             trace!("[transport] Request <- <hidden>");
 
-            if let Some(error_value) = response.as_object().map(|r| r.get("error")).flatten() {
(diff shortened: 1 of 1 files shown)


packages/cubejs-backend-native/src/transport.rs
```diff
@@ -377,38 +377,43 @@ impl TransportService for NodeBridgeTransport {
...
             }
+
             let response: serde_json::Value = result?;
+
             #[cfg(debug_assertions)]
             trace!("[transport] Request <- {:?}", response);
             #[cfg(not(debug_assertions))]
             trace!("[transport] Request <- <hidden>");
 
-            let load_err = match serde_json::from_value::<TransportLoadResponse>(response.clone()) {
```
ovr, line 386: serde_json::from_value requires passing the own of the value; it's why we used `clone` here. The biggest problem is that we used it for successful responses, which is a large structure.

PR fixes that issue.

PR #8665, opened 2024-09-03
fix(snowflake): fixing the snowflake schema fetcher logic (#7440)

### LINKED ISSUE #7440: Snowflake Tables 'Undefined'
**Describe the bug**
After successfully authenticating to Snowflake and running the initial information schema query, Cube returns 'Undefined' for our tables in the Data Model section.  I can see initial information schema query successful runs, but I never get the results back into Cube.

**To Reproduce**
1. Connect to a Snowflake instance
2. Wait for initial response
3. See error in Data Model

**Expected behavior**

**Screenshots**
Screenshot 2023-11-20 at 3 13 33 PM

**Version:**
[e.g. 0.4.5]

**Additional context**

christianallred-nomi:
**Check List**

Closes #7440


igorlukanin:
Hi @christianallred-nomi 👋 

Thanks for your contribution! Do you think this is a breaking change?

christianallred:
It should not be no. 

A while ago snowflake made a decision to always cap case their responses. This caused an Outage for us. And is what causes the bug this or is aiming to fix with cube. 

The fact that this is working for some snowflake users means they have this setting set globally instead of by session. 

I liked this solution because it fixes the problem for affected users but should leave the existing users the same. 

I have not had time to test this 100% however so if someone with a working snowflake instance can test it, that would be great:

christianallred:
An alternate decision that would not require a session update would be to update the snowflake driver to overwrite the result parser function

paveltiunov:
@christianallred Thanks for contributing it! I think it's definitely great feature however as of now it should go under the flag which we'd enable by default going forward.

christianallred:
> @christianallred Thanks for contributing it! I think it's definitely great feature however as of now it should go under the flag which we'd enable by default going forward.

Whatever y'all decide is fine. It’s just broken for our use case until a fix is decided on. FWIW I don’t really have skin in the game anymore as I am leaving the org that was looking at cube as a solution. Cheers :)

mattssll:
@paveltiunov any info about this? I'm having issue with the undefined in snowflake
https://github.com/cube-js/cube/issues/7440

paveltiunov:
@mattssll We're open for this change if it'd land under the flag. If you have capacity to fix PR we'd be happy to merge it.

KSDaemon:
@christianallred-nomi if you don't know, here are a few tips on how to add flag. It's pretty easy!

Add a new env in [env.ts](https://github.com/cube-js/cube/blob/master/packages/cubejs-backend-shared/src/env.ts#L1372), something like:
```js
  /**
   * Snowflake flag controlling QUOTED_IDENTIFIERS_IGNORE_CASE.
   * Defaults to FALSE if not set
   */
  snowflakeQuotedIdentIgnoreCase: ({
    dataSource
  }: {
    dataSource: string,
  }) => (
    const val = process.env[
      keyByDataSource(
        'CUBEJS_DB_SNOWFLAKE_QUOTED_IDENTIFIERS_IGNORE_CASE',
        dataSource,
      )
    ];
    if (val) {
      if (val.toLocaleLowerCase() === 'true') {
        return true;
      } else if (val.toLowerCase() === 'false') {
        return false;
      } else {
        throw new TypeError(
          `The ${
            keyByDataSource(
              'CUBEJS_DB_SNOWFLAKE_QUOTED_IDENTIFIERS_IGNORE_CASE',
              dataSource,
            )
          } must be either 'true' or 'false'.`
        );
      }
    } else {
      return false;
    }
  ),
```

Then read and store the flag value in driver's [constructor](https://github.com/cube-js/cube/blob/master/packages/cubejs-snowflake-driver/src/SnowflakeDriver.ts#L253, smth like:
```js
this.ignoreCase = getEnv('snowflakeQuotedIdentIgnoreCase', { dataSource })
```

And check it in the `initConnection()`, or simply pass as arg to execute.
```js
if (this.ignoreCase) {
   await this.execute(connection, 'ALTER SESSION SET QUOTED_IDENTIFIERS_IGNORE_CASE = FALSE', [], false);
}
```

PR #8677, opened 2024-09-06
ci(cubesql): Attach cubenativeutils and cubesqlplanner to cargo fmt and cargo check

mcheshkov:
**Check List**

Those are separate workspaces, with separate Cargo.lock, so we should at least check that lockfiles are up to date


KSDaemon (approved):
👍🏻

PR #8712, opened 2024-09-16
feat(schema-compiler): exact match preagg with custom granularity without the need for allow_non_strict_date_range_match option

KSDaemon:
**Check List**

### DIFF
--- a/packages/cubejs-schema-compiler/src/adapter/PreAggregations.js
+++ b/packages/cubejs-schema-compiler/src/adapter/PreAggregations.js
@@ -431,7 +431,11 @@ export class PreAggregations {
   static sortTimeDimensionsWithRollupGranularity(timeDimensions) {
     return timeDimensions && R.sortBy(
       R.prop(0),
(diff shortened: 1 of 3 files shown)


packages/cubejs-schema-compiler/src/adapter/PreAggregations.js
```diff
@@ -431,7 +431,11 @@ export class PreAggregations {
   static sortTimeDimensionsWithRollupGranularity(timeDimensions) {
     return timeDimensions && R.sortBy(
       R.prop(0),
-      timeDimensions.map(d => [d.expressionPath(), d.rollupGranularity()])
+      timeDimensions.map(d => (d.isPredefinedGranularity() ?
+        [d.expressionPath(), d.rollupGranularity()] :
```
ovr, line 435: It's better to align this two tuples to the same size (3). Let's use `, null]`

ovr (approved):
LGTM, minor nit

PR #8730, opened 2024-09-19
feat(snowflake-driver): support azure exports buckets

KSDaemon:
This PR brings support for Azure buckets in Snowflake driver and also `STORAGE_INTEGRATION` option for AWS S3 buckets.

Other changes included:
* Refactored extraction of S3 signed urls of exported files in BaseDriver
* All drivers that could export to S3 now use the same implementation from BaseDriver
* Refactored extraction of Azure signed urls of exported files in BaseDriver
* All drivers that could export to Azure now use the same implementation from BaseDriver
* Refactored extraction of GCP signed urls of exported files in BaseDriver
* All drivers that could export to GCP now use the same implementation from BaseDriver (besides BigQuery)

**Check List**


packages/cubejs-base-driver/src/BaseDriver.ts
```diff
@@ -672,4 +693,72 @@ export abstract class BaseDriver implements DriverInterface {
...
+    const account = parts[0];
+    const container = parts[1].split('/')[0];
+    const credential = new StorageSharedKeyCredential(account, azureConfig.azureKey);
+    const url = `https://${account}.blob.core.windows.net`;
+    const blobServiceClient = azureConfig.sasToken ?
+      new BlobServiceClient(`${url}?${azureConfig.sasToken}`) :
+      new BlobServiceClient(url, credential);
+
+    const csvFiles: string[] = [];
+    const expr = new RegExp(`${tableName}\\/.*\\.csv(\\.gz)?$`, 'i');
```
srh, line 740: I don't understand the RegExp here. First of all, if listBlobsFlat has `prefix: tableName` as a parameter, why do we need to filter its results by tableName?  Could the regex just be `\\.csv(\\.gz)?$`  ?

Second of all, tableName can and will have regexp special characters -- we put in `.` ourself.  So there is a problem there.

Third of all, if the filtering is useful, and tableName is the suffix of some other one -- let's say we have both "abc.def" and "fooabc.def" -- then this regexp would match some directory that we would want to filter out.

Also it can match tableName at arbitrary levels, possibly matching other directory-level names coincidentally -- maybe not possible with Azure, I don't know.

The old code gets `foldername` value it uses in a different manner, but it could very well have similar issues.

KSDaemon: Good points! Maybe `.startsWith()` would be enough!

KSDaemon: Table name is required if for some reason there are different files (or exports) placed in the same dir

srh: So, the prefix option "filters the results to return only containers whose name *begins* with the specified prefix".
(Is that a mistake and does that mean blobs instead of containers?)  By prefix do they actually mean a full path element?

How do we know, right now, if tableName is `foo`, this won't return csvs for the table `foo2`?

paveltiunov (approved):

packages/cubejs-base-driver/src/BaseDriver.ts
KSDaemon: prefix filters objects INSIDE container, not the containers. So it should be what we actually need.

Answering your second question — I believe the current impl — no way.
I think it can be done by `prefix: `${tableName}/`` - including slash in the prefix. It seems that should do the trick

PR #8774, opened 2024-10-03
fix(schema-compiler): ExportNamedDeclaration for variables object

arkadiusz-swierczek-pm:
**Check List**

Class `ImportExportTranspiler` contains a bug that prevent using exported object syntax from cubes.

Related: https://cube.dev/docs/product/data-modeling/dynamic/code-reusability-export-and-import#import-from-parent-directories

The changes were tesed using https://astexplorer.net/ and content:
```
export const helperFunction = () => 'hello'
export { helperFunction as alias }
export default helperFunction
```

Without the fix, the Compiler throws error `TypeError: Cannot read properties of null (reading 'declarations')`. After the fix the babel transpiles the file to:
```
const helperFunction = () => 'hello';

addExport({
  helperFunction: helperFunction
})
addExport({
  alias: helperFunction
});
setExport(helperFunction);
```


KSDaemon (approved):
👍🏻 Looks good! @arkadiusz-swierczek-pm thnx for fixing this!

KSDaemon:
@arkadiusz-swierczek-pm, it would be awesome if you could add a kind of basic test for this in [transpilers.test.ts](https://github.com/cube-js/cube/blob/master/packages/cubejs-schema-compiler/test/unit/transpilers.test.ts)

packages/cubejs-schema-compiler/src/compiler/transpilers/ImportExportTranspiler.ts
```diff
@@ -50,7 +50,7 @@ export class ImportExportTranspiler implements TranspilerInterface {
           }
         });
         const addExportCall = t.callExpression(t.identifier('addExport'), [t.objectExpression(<t.ObjectProperty[]>declarations)]);
-        if (path.get('declaration')) {
+        if ('declaration' in path.node && path.node.declaration) {
           path.replaceWithMultiple([
             // @todo fix without any
             (<any>path.get('declaration')).node,
```
ovr, line 56: As I can see, `path.get('declaration')` is still used in `path.replaceWithMultiple`.
Two different nodes in `if` condition and `replaceWithMultiple` call.

PR #8781, opened 2024-10-04
fix: Render LIMIT 0 and OFFSET 0 properly in SQL templates

mcheshkov:
**Check List**

`if limit` condition in Jinja will fail for both `None` and `Some(0)`
This is more or less fine for offset, but completely wrong for limit. For consistency both are changed.


srh (approved):

KSDaemon (approved):
👍🏻 LGTM!

ovr (approved):

PR #8784, opened 2024-10-04
fix: Add explicit parameters limit in PostgreSQL and Redshift drivers

mcheshkov:
**Check List**

`PostgresDriver` uses `pg` module as a driver, and `pg` does not check incoming params, and if there's too many it just sends incorrect bind message. It would end up as "expected a lot of params, but 0 was passed" error from server.

`RedshiftDriver` extends `PostgresDriver`, but there's another catch: Redshift server misbehaves a little, and breaks on 32768 or more parameters with `there is no parameter $-32768`.

There's not much we can do about it, but at least we can make error messages clear.


srh (approved):

KSDaemon (commented):
👍🏻

PR #8789, opened 2024-10-07
fix(cubestore): don't add all measures to default preagg index

KSDaemon:
This change improves the cube store preaggregation creation: the default index for a preaggregation now excludes all measures from the sorting keys

**Check List**

### DIFF
--- a/docs/pages/product/caching/using-pre-aggregations.mdx
+++ b/docs/pages/product/caching/using-pre-aggregations.mdx
@@ -286,7 +286,7 @@ monitor Cube Store workers load, and iterate as needed.
 
 **When you define a pre-aggregation without any explicit indexes, the default
 index is created.** In this index, dimensions come first, time dimensions come
-second, and measures come last.
+second.
 
 When you define additional indexes, you don't incur any additional costs on
 the data warehouse side. However, the pre-aggregation build time for a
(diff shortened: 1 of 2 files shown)


igorlukanin:
@KSDaemon I believe this has to be changed: https://cube.dev/docs/product/caching/using-pre-aggregations#using-indexes

Should this be considered a breaking change? Also, will upgrading to this version result in pre-aggregations being rebuilt, entirely or partially?

waralexrom (approved):

KSDaemon:
@igorlukanin Actually `No` :) It's internal optimization, so indexes are still copies of all preagg data. The difference is that now, sorting keys for the index do not include measures (as it's mostly useless, no filtering is done by measures, potentially they may be useful for postagregation filtering (aka `HAVING xxx`). But that is another story.

Rgd existing indexes — there should be no changes. This technique will be applied only for new index creation.

igorlukanin:
Docs still need to be updated wrt what is included in the default index, don't they?

KSDaemon:
> When you define a pre-aggregation without any explicit indexes, the default index is created. In this index, dimensions come first, time dimensions come second, and measures come last.

This is still true :)

igorlukanin:
@KSDaemon Hopefully I did it correctly.

PR #8791, opened 2024-10-07
docs: Data-at-rest encryption in Cube Store and customer-provided key…

igorlukanin:


docs/pages/product/workspace/encryption-keys.mdx
```diff
@@ -0,0 +1,89 @@
...
+refresh worker, or manually.
+
+You can check which encryption key is used by any pre-aggregation partition by
+querying `system.tables` in Cube Store via [SQL Runner][ref-sql-runner]:
+
+<Screenshot src="https://ucarecdn.com/017ca9d6-e8d2-4896-9324-1bec38aaa621/" />
+
+<WarningBox>
+
+Only newly built or rebuilt pre-aggregation partitions will use the newly addded
```
srh, line 63:
```suggestion
Only newly built or rebuilt pre-aggregation partitions will use the newly added
```

docs/pages/product/workspace/encryption-keys.mdx
```diff
@@ -0,0 +1,89 @@
...
+## Customer-managed keys for Cube Store
+
+On the <Btn>Encryption Keys</Btn> page, you can see all previously provided keys:
+
+<Screenshot src="https://ucarecdn.com/48038ac1-fdf1-4c87-8860-ac503bfcdac3/" />
+
+### Add a key
+
+To add an encryption key, click <Btn>Create</Btn> to open a modal window.
+Provide any string as the key name and paste a [Base64-encoded][link-base64]
```
srh, line 26: It specifically has to be standard base64 in its canonical representation.  I think it is a good idea to say it has to be standard base64, and change the link to specifically point to something defining standard base64.  I like how this Google documentation does so: https://cloud.google.com/storage/docs/encryption/customer-supplied-keys

Just to be clear, we actually do reject base64 that isn't in its canonical representation -- we check `x == encode(decode(x))`.  The UI provides instant feedback to the user if this is not the case.

srh (changes requested):

docs/pages/product/workspace/encryption-keys.mdx
```diff
@@ -0,0 +1,89 @@
...
+querying `system.tables` in Cube Store via [SQL Runner][ref-sql-runner]:
+
+<Screenshot src="https://ucarecdn.com/017ca9d6-e8d2-4896-9324-1bec38aaa621/" />
+
+<WarningBox>
+
+Only newly built or rebuilt pre-aggregation partitions will use the newly addded
+encryption key. Previously built partitions will still be encrypted using
+previously provided keys. If you [drop a key](#drop-a-key) before these partitions
+are rebuilt, querying them will yield an error.
```
srh, line 66: One thing that came to mind when reading this is that users might be afraid they can't rebuild pre-aggregations, if they don't have a correct mental model of how pre-aggregations work. (I don't necessarily have an accurate simulation of users though.) At the cost of brevity, it might be worth a sentence making it clear the pre-agg can still be rebuilt from the underlying data.

srh (approved):

PR #8796, opened 2024-10-10
fix(cubesql): Ignore `__user IS NOT NULL` filter

MazterQyou:
**Check List**

This PR ignores `__user IS NOT NULL` filter expression when replacing filters for Cube. The filter is redundant and has no effect, but forces SQL push down without the fix. Related test is included.

### DIFF
--- a/rust/cubesql/cubesql/src/compile/rewrite/rules/filters.rs
+++ b/rust/cubesql/cubesql/src/compile/rewrite/rules/filters.rs
@@ -349,7 +349,14 @@ impl RewriteRules for FilterRules {
                     "?filter_aliases",
                 ),
                 change_user_member("?user"),
-                self.transform_change_user_eq("?column", "?literal", "?user"),
+                self.transform_change_user_eq(
+                    "?column",
+                    "?literal",
+                    "?alias_to_cube",
+                    "?members",
+                    "?filter_aliases",
+                    "?user",
+                ),
             ),
             transforming_rewrite(
                 "change-user-equal-filter",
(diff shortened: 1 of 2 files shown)


rust/cubesql/cubesql/src/compile/rewrite/rules/filters.rs
```diff
@@ -3338,6 +3354,22 @@ impl FilterRules {
...
     }
 
+    fn transform_user_is_not_null(
+        &self,
+        column_var: &'static str,
+    ) -> impl Fn(&mut EGraph<LogicalPlanLanguage, LogicalPlanAnalysis>, &mut Subst) -> bool {
+        let column_var = var!(column_var);
+
+        move |egraph, subst| {
+            for column in var_iter!(egraph[subst[column_var]], ColumnExprColumn).cloned() {
```
mcheshkov, line 3364: I think this rule should rely on something like`alias_to_cubes` from left side or  `cube_reference` from analysis here, to check column origin instead of name.

For a query like this rule should not apply, but it will:
```
SELECT
    __user,
    2
FROM (
    SELECT
        1 as __user,
        customer_gender
    FROM
        KibanaSampleDataEcommerce
) t
WHERE
      NOT __user IS NULL
  AND __user = 'gopher'
```


And for a query like this rule should apply, but it won't:
```
SELECT
    fooser,
    2
FROM (
    SELECT
        __user as fooser,
        customer_gender
    FROM
        KibanaSampleDataEcommerce
) t
WHERE
      NOT fooser IS NULL
  AND fooser = 'gopher'
```

MazterQyou: Good point, but the behavior of this transform fn is consistent with `transform_change_user_eq`; changing them both right now is out of scope and unlikely a good idea. Besides, `__user` is not really a selectable column, so shouldn't be part of a subquery anyway; it is for filtering only.

mcheshkov: Oh, `__user` in select is just for example sake, should not be relevant.
My point is that both `__user` as a column name can come from alias straight from query, and it should not trip this rule, but actual `__user` column can be renamed before using, so `transform_change_user_eq` is equally broken.
Can you at least add a couple TODOs and ignored tests?

MazterQyou: Updated the PR to respect realiasing of `__user` in both cases. 🙌

ovr (approved):

PR #8812, opened 2024-10-14
fix(cubesql): Split `__user` WHERE predicate into separate filter node

MazterQyou:
**Check List**

This PR adds an optimizer that splits meta filters (`WHERE __user...`) into a separate `Filter` node. This helps with queries where filter contains expressions that cannot be pushed down to CubeScan. In this case SQL push down is applied but meta filters should go into CubeScan anyway. Related tests are included.


mcheshkov (approved):
Let's rebase on fresh `master` and land it

PR #8815, opened 2024-10-15
feat(Firebolt): Use driver's own test connection method

ptiurin:
**Check List**

Delegating connection verification to the Firebolt SDK driver. This allows more flexibility and more importantly avoids unnecessary charges to Firebolt's customers: Firebolt engines have auto-stop functionality that allows them to spin down when not in use. However, healthcheck queries can keep them up, despite not having any useful load. Firebolt's SDK provides a way to check if the connection is working without resetting the auto-stop timer. If the engine is down or there are connectivity errors `testConnection` will still error out, as before.


KSDaemon (approved):
👍🏻

ovr:
Thank you @ptiurin 🍰

Issue #8829, opened 2024-10-17
Add support for Arrow Flight JDBC driver


gingermike:
**Is your feature request related to a problem? Please describe.**

We currently use Dremio as our backend and while the RESTful driver provided by the community works, the Arrow Flight JDBC driver should be significantly more performant especially with larger volumes of data

**Describe the solution you'd like**

Extend the jdbc cube driver to use `org.apache.arrow.driver.jdbc.ArrowFlightJdbcDriver`

**Describe alternatives you've considered**

A possible alternative is creating another bespoke Dremio driver that calls Dremio using the Flight RPC protocol directly, rather than using the JDBC wrapper. I _think_ this is possible in node - there's a wrapper that calls to an implementation in Rust, iirc

vincentditlevinz:
High,

I tried to just enhance `cubejs-jdbc-driver`, but I struggled with Java NoClassDefFoundError, more, the maven support downloaded corrupted jar files...  That's why I finally implemented this feature by following `packages/cubejs-databricks-jdbc-driver` as a template.

Here is the PR: https://github.com/cube-js/cube/pull/9789

PR #8830, opened 2024-10-17
Adding support for arrow flight JDBC driver

gingermike:
**Check List**

[Add support for Arrow Flight JDBC driver](https://github.com/cube-js/cube/issues/8829)

Extended the configuration as per the contribution guide to add Arrow Flight driver

### DIFF
--- a/packages/cubejs-jdbc-driver/src/supported-drivers.ts
+++ b/packages/cubejs-jdbc-driver/src/supported-drivers.ts
@@ -63,5 +63,19 @@ export const SupportedDrivers: Record<string, DriverOptionsInterface> = {
       user: process.env.CUBEJS_DB_USER,
       password: process.env.CUBEJS_DB_PASS,
     }
+  },
+  arrowflight: {
+    driverClass: 'org.apache.arrow.driver.jdbc.ArrowFlightJdbcDriver',
+    prepareConnectionQueries: [],
+    mavenDependency: {
+      groupId: 'org.apache.arrow',
+      artifactId: 'flight-sql-jdbc-driver',
+      version: '17.0.0'
+    },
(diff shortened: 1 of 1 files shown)


paveltiunov:
@gingermike Thanks for contributing! This part just adds JDBC driver but we'd need separate driver package. Please see https://github.com/cube-js/cube/blob/master/CONTRIBUTING.md#contributing-database-drivers.

gingermike:
Thanks @paveltiunov - are you suggesting that I need to write a standalone Flight RPC implementation? 

I was under the impression we could just extend the DbTypes for the JDBC driver as listed here: https://github.com/cube-js/cube/blob/master/CONTRIBUTING.md#implementing-a-jdbc-driver

The Flight SQL JDBC driver is very mature - I'm not sure how much value there would be in writing a native typescript version (that would end up having to call down to a native Rust implementation anyway)?

paveltiunov:
@gingermike Driver itself is just one part. In order to use it with Dremio for example you'd need also a dialect as it's different for every DB.

Raven888888:
Any progress on this? We would like to enjoy the benefit of Arrow with JDBC for standard databases like Postgres, Mysql and SQL Server.

igorlukanin:
> We would like to enjoy the benefit of Arrow with JDBC for standard databases like Postgres, Mysql and SQL Server.

@Raven888888 I'm pretty sure that these databases in particular can not benefit from using Arrow Flight since they are not columnar.

Also, generally, as Pavel mentioned above, this PR does not provide a data source driver as many more parts should be also implemented: https://github.com/cube-js/cube/blob/master/CONTRIBUTING.md#contributing-database-drivers

Thus, closing this PR.

PR #8835, opened 2024-10-18
refactor(cubesql): Extract cube join condition check for rewrites to separate function

mcheshkov:


mcheshkov:
Merged as part of https://github.com/cube-js/cube/pull/9032

Issue #4604, opened 2022-05-26
missing data is always forced to 0


vprus:
Since commit https://github.com/cube-js/cube.js/pull/3664/files if a pivot does not have any values for particular combination of dimensions, '0' is returned. This is suboptimal, because in a number of contexts, missing data and zero have different meaning. For example, I use cubejs to visualize certain business predictions, where missing data can indicate a technical problem, while "0" indicates a business problem.

**To reproduce** 

Start cubejs with docker with
```
% docker run -p 4000:4000 \
  -v ${PWD}:/cube/conf \
  -e CUBEJS_DEV_MODE=true \
  cubejs/cube
```

Make sure the version is 0.30.7 (the current one as of today) or later.

Connect to a database (I've used clickhouse)

Define schema as 
```
cube('Test', {
    sql: `
SELECT * FROM VALUES('x Int64, y Int64, value Int64', (1, 1, 7), (2, 2, 19))
`,



    measures: {
	value: {
	    sql: `value`,
	    type: `sum`
        }
    },

    dimensions: {
	x: {
	    sql: `x`,
	    type: `number`
	},
	y: {
	    sql: `y`,
	    type: `number`
	}
    }
})
```

Go to the build tab, and build the query as follows:
- Select value as measure
- Select x and y as dimensions
- In pivot config, use 'x' for x axis, and 'y,measures' for y axis
- Leave "Fill Missing Dates" at the default state of checked.

Run the query.

**Observed effect**

Each table cell we have no values for shows "0"

**Expected effect**

Each table cell we have no value for is empty. Previously (at least in version 0.27.35), that's exactly what happened.

**Version:**
0.30.7

vasilev-alex:
Hi @vprus!

Yeah, we missed that part out. But reverting this will cause a breaking change, which we only can do in the next minor release.

vprus:
Thank you, as soon as it is fixed eventually, I'll be happy, no rush.

igorlukanin:
Closing as it is now possible to use `fillWithValue` in the JavaScript SDK.

Issue #718, opened 2020-06-11
ResultSet fillMissingDates not working as expected if dimensions in the query


0is1:
**Describe the bug**

According to the documentation (https://cube.dev/docs/@cubejs-client-core#result-set-pivot)  `resultSet.tablePivot()` (and other pivot methods) should have `pivotConfig.fillMissingDates` true by default.

It seems that there's some sort of bug if dimensions are set to the query. All these (`resultSet.tablePivot()`, `resultSet.tablePivot({fillMissingDates: false})` and `resultSet.tablePivot(resultSet.normalizePivotConfig())`) return same data for this kind of query:
```
{
    dimensions: ["xxx.name", "Dates.date"],
    filters: [],
    measures: ["zzz.count"],
    offset: 0,
    timeDimensions: [{
        dateRange: ["2020-06-04T00:00:00.000", "2020-06-10T23:59:59.999"],
        dimension: "Dates.date"
    }],
    timezone: "Europe/Helsinki"
}
```

`resultSet.tablePivot(resultSet.normalizePivotConfig())` returns

```
[
    {xxx.name: "yyy", Dates.date: "2020-06-04T00:00:00.000", zzz.count: "1"},
    {xxx.name: "yyy", Dates.date: "2020-06-07T00:00:00.000", zzz.count: "1"},
    {xxx.name: "yyy", Dates.date: "2020-06-09T00:00:00.000", zzz.count: "1"},
    {xxx.name: "yyy2", Dates.date: "2020-06-04T00:00:00.000", zzz.count: "1"},
    {xxx.name: "yyy2", Dates.date: "2020-06-07T00:00:00.000", zzz.count: "1"},
    {xxx.name: "yyy2", Dates.date: "2020-06-09T00:00:00.000", zzz.count: "1"},
    {xxx.name: "yyy2", Dates.date: "2020-06-04T00:00:00.000", zzz.count: "1"},
    {xxx.name: "yyy2", Dates.date: "2020-06-07T00:00:00.000", zzz.count: "1"},
    {xxx.name: "yyy2", Dates.date: "2020-06-09T00:00:00.000", zzz.count: "1"},
    {xxx.name: "yyy3", Dates.date: "2020-06-04T00:00:00.000", zzz.count: "1"},
    {xxx.name: "yyy3", Dates.date: "2020-06-07T00:00:00.000", zzz.count: "1"},
[...]

PR #8843, opened 2024-10-20
feat(cubejs-client-core): Fill missing dates with custom measure value

nagulan23:
**Check List**

https://github.com/cube-js/cube/issues/4604
https://github.com/cube-js/cube/issues/718

Opens flexibility for users to provide custom value to be filled for the missing date.


igorlukanin:
Thanks for your contribution, @nagulan23 🙌 

Could you also please update the docs on this page? https://cube.dev/docs/reference/frontend/cubejs-client-core

nagulan23:
@igorlukanin Have updated the doc.

packages/cubejs-client-core/index.d.ts
```diff
@@ -252,9 +252,13 @@ declare module '@cubejs-client/core' {
...
     y?: string[];
     /**
-     * If `true` missing dates on the time dimensions will be filled with `0` for all measures.Note: the `fillMissingDates` option set to `true` will override any **order** applied to the query
+     * If `true` missing dates on the time dimensions will be filled with fillWithValue or `0` by default for all measures.Note: the `fillMissingDates` option set to `true` will override any **order** applied to the query
      */
     fillMissingDates?: boolean | null;
+    /**
+     * Value to autofill all the missing date's measure.
+     */
+    fillWithValue?: string | null;
```
KSDaemon, line 261: Maybe it should be Number instead?

KSDaemon: And even in tests you use `5` as number, and not as string.

KSDaemon:
Jentle ping @nagulan23

packages/cubejs-client-core/index.d.ts
nagulan23: @KSDaemon It was meant to be a string. For example lets take an example of users need the auto filled values to be '-' or 'N/A' or 'Missing'. So this enables a scope of allowing users to fill in the missing measure with any value: Number / String.

nagulan23: adding an extra testcase of string example

KSDaemon, line 261: Thnx for adding another test! 
As number is still a valid type for this, let's updat the typedef too:
```suggestion
    fillWithValue?: string | number | null;
```

PR #8853, opened 2024-10-22
chore(ci): Publish - ignore latest tag for docker builds on LTS branches

ovr:


.github/workflows/publish.yml
```diff
@@ -783,15 +796,30 @@ jobs:
       - name: Detect branch name
         id: detect_branch
         run: |
-          BRANCH_NAME=$(git branch -r --contains "$GITHUB_SHA" | grep -v 'HEAD' | head -n 1 | sed -e 's/.*origin\///')
-          echo "branch_name=$BRANCH_NAME" >> "$GITHUB_OUTPUT"
+          BRANCH=$(git branch -r --contains "$GITHUB_SHA" | grep -v 'HEAD' | head -n 1 | sed -e 's/.*origin\///')
```
ovr, line 799: This solution doesn't work when someone bases his branch on top of HEAD while the release is in progress. But doing it as a first step makes this risk lower.

Issue #8765, opened 2024-10-01
COUNT(*) Warning breaking Pushdown queries


pauldheinrichs:
**Describe the bug**

Since this [PR](https://github.com/cube-js/cube/pull/8667) has been merged - querying in the BI tool to check counts results in an error rather than getting a count. With pushdown enabled - i wouldn't have anticipated this would be the expected flow if a _count_ measure didn't exist. I understand to mitigate it's easily done by adding measures, it just felt like an unexpected outcome possibly? 

**To Reproduce**

**Expected behavior**

**Screenshots**

**Minimally reproducible Cube Schema**
In case your bug report is data modelling related please put your minimally reproducible Cube Schema here.
You can use selects without tables in order to achieve that as follows.

```javascript
cube(`Orders`, {
  sql: `
  select 1 as id, 100 as amount, 'new' status
  UNION ALL
  select 2 as id, 200 as amount, 'new' status
  UNION ALL
  select 3 as id, 300 as amount, 'processed' status
  UNION ALL
  select 4 as id, 500 as amount, 'processed' status
  UNION ALL
  select 5 as id, 600 as amount, 'shipped' status
  `,
  measures: {
    count: {
      type: `count`,
    },
    totalAmount: {
      sql: `amount`,
      type: `sum`,
    },
    toRemove: {
      type: `count`,
    },
  },
  dimensions: {
    status: {
      sql: `status`,
      type: `string`,
    },
  },
});
```

**Version:**
[e.g. 0.4.5]

**Additional context**

pauldheinrichs:
Querying on on both the view and the cube break with this. @igorlukanin Just an FYI feels like a un-intended result of this change

pauldheinrichs:
I've done more digging - and the source of the main issue in views container `prefix: true,` it cannot find the count(*) measure

PR #8899, opened 2024-11-01
feat(snowflake-driver): add ignore case statements flag

### LINKED ISSUE #7440: Snowflake Tables 'Undefined'

mike-taylor-nomihealth:
**Check List**

Closes #7440
Supersedes [#8765](https://github.com/cube-js/cube/pull/8665)

### DIFF
--- a/packages/cubejs-snowflake-driver/src/SnowflakeDriver.ts
+++ b/packages/cubejs-snowflake-driver/src/SnowflakeDriver.ts
@@ -450,6 +450,7 @@ export class SnowflakeDriver extends BaseDriver implements DriverInterface {
 
       await this.execute(connection, 'ALTER SESSION SET TIMEZONE = \'UTC\'', [], false);
       await this.execute(connection, `ALTER SESSION SET STATEMENT_TIMEOUT_IN_SECONDS = ${this.config.executionTimeout}`, [], false);
+      await this.execute(connection, 'ALTER SESSION SET QUOTED_IDENTIFIRS_IGNORE_CASE = FALSE', [], false);
 
       return connection;
     } catch (e) {
--- a/packages/cubejs-snowflake-driver/src/SnowflakeDriver.ts
+++ b/packages/cubejs-snowflake-driver/src/SnowflakeDriver.ts
@@ -173,6 +173,7 @@ interface SnowflakeDriverOptions {
   resultPrefetch?: number,
   exportBucket?: SnowflakeDriverExportBucket,
   executionTimeout?: number,
+  caseSensitiveIdentifiers?: boolean,
   application: string,
   readOnly?: boolean,
 
@@ -213,6 +214,7 @@ export class SnowflakeDriver extends BaseDriver implements DriverInterface {
       'CUBEJS_DB_SNOWFLAKE_PRIVATE_KEY_PATH',
       'CUBEJS_DB_SNOWFLAKE_PRIVATE_KEY_PASS',
       'CUBEJS_DB_SNOWFLAKE_OAUTH_TOKEN_PATH',
+      'CUBEJS_DB_SNOWFLAKE_QUOTED_IDENTIFIERS_IGNORE_CASE',
     ];
   }
 
@@ -279,6 +281,7 @@ export class SnowflakeDriver extends BaseDriver implements DriverInterface {
       exportBucket: this.getExportBucket(dataSource),
       resultPrefetch: 1,
       executionTimeout: getEnv('dbQueryTimeout', { dataSource }),
+      ignoreCase: getEnv('snowflakeQuotedIdentIgnoreCase', { dataSource }),
       exportBucketCsvEscapeSymbol: getEnv('dbExportBucketCsvEscapeSymbol', { dataSource }),
       application: 'CubeDev_Cube',
       ...config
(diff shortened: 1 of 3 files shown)


paveltiunov:
@mike-taylor-nomihealth Thanks for contributing! Could we please put it under the flag enabled by default? Thanks!

mike-taylor-nomihealth:
> @mike-taylor-nomihealth Thanks for contributing! Could we please put it under the flag enabled by default? Thanks!

I believe I got this all set up properly - let me know if this is not what you meant

packages/cubejs-snowflake-driver/src/SnowflakeDriver.ts
```diff
@@ -450,7 +453,9 @@ export class SnowflakeDriver extends BaseDriver implements DriverInterface {
 
       await this.execute(connection, 'ALTER SESSION SET TIMEZONE = \'UTC\'', [], false);
       await this.execute(connection, `ALTER SESSION SET STATEMENT_TIMEOUT_IN_SECONDS = ${this.config.executionTimeout}`, [], false);
-
+      if (this.ignoreCase) {
```
KSDaemon, line 456: I realized that we mess the value of our env `CUBEJS_DB_SNOWFLAKE_QUOTED_IDENTIFIERS_IGNORE_CASE` / `this.ignoreCase` and the value we pass to the Snowflake.

I mean — if we have `true` — we pass `false` and vice versa.
Maybe it would be better to align them. Possible options imho are:
1. Rename our var to smth like `CUBEJS_DB_SNOWFLAKE_QUOTED_IDENTIFIERS_RESPECT_CASE`. Then having it set to `true` — means exactly what we set by this statement — not ignoring the case.
2. Or.... Use `CUBEJS_DB_SNOWFLAKE_QUOTED_IDENTIFIERS_IGNORE_CASE` as is: if it is true — then — we should pass `true` to Snowflake. In this case, don't forget to update the default value if var is not set, as we want respect case by default.

KSDaemon (changes requested):
Hi @mike-taylor-nomihealth! 
Thank you for taking this up! Left some comments! WDYT?

packages/cubejs-snowflake-driver/src/SnowflakeDriver.ts
```diff
@@ -173,6 +173,7 @@ interface SnowflakeDriverOptions {
   resultPrefetch?: number,
   exportBucket?: SnowflakeDriverExportBucket,
   executionTimeout?: number,
+  caseSensitiveIdentifiers?: boolean,
```
KSDaemon, line 176: This should be `ignoreCase` I suppose

KSDaemon:
Gentle ping @mike-taylor-nomihealth!

packages/cubejs-snowflake-driver/src/SnowflakeDriver.ts
mike-taylor-nomihealth: Are we absolutely we sure we want to respect case by default for the snowflake driver given that they made the change to ignore it and upper() everthing?  It seems non-intuitive that their driver doesn't default to case insensitive when that's the behavior of their UI and everything in between, so if we set the default behavior to be case sensitive it feels like there are a lot of folks using this driver that are going to have to set a value explicitly when the default behavior probably should reflect how snowflake is using it (even if they didn't update their driver settings to reflect that by default.).

I made the change in the PR but am just playing devils advocate here, because it seems like we got here because this ISN'T the default behavior but probably should be because of how SNOWFLAKE decided to update their behaviors.

packages/cubejs-snowflake-driver/src/SnowflakeDriver.ts
KSDaemon, line 176:
```suggestion
  identIgnoreCase?: boolean,
```

packages/cubejs-snowflake-driver/src/SnowflakeDriver.ts
```diff
@@ -279,6 +281,7 @@ export class SnowflakeDriver extends BaseDriver implements DriverInterface {
       exportBucket: this.getExportBucket(dataSource),
       resultPrefetch: 1,
       executionTimeout: getEnv('dbQueryTimeout', { dataSource }),
+      ignoreCase: getEnv('snowflakeQuotedIdentIgnoreCase', { dataSource }),
```
KSDaemon, line 284:
```suggestion
      identIgnoreCase: getEnv('snowflakeQuotedIdentIgnoreCase', { dataSource }),
```

packages/cubejs-snowflake-driver/src/SnowflakeDriver.ts
```diff
@@ -450,7 +453,12 @@ export class SnowflakeDriver extends BaseDriver implements DriverInterface {
 
       await this.execute(connection, 'ALTER SESSION SET TIMEZONE = \'UTC\'', [], false);
       await this.execute(connection, `ALTER SESSION SET STATEMENT_TIMEOUT_IN_SECONDS = ${this.config.executionTimeout}`, [], false);
-
+      
+      // We only want to ignore the case if someone sets the value to false explicitly since the default assumption
+      // is that casing matters
+      if (!this.ignoreCase) {
```
KSDaemon, line 459:
```suggestion
      if (!this.identIgnoreCase) {
```

KSDaemon (changes requested):

KSDaemon:
Okay... I took it into my own hands.: https://github.com/cube-js/cube/pull/9131.

PR #8903, opened 2024-11-03
fix(jdbc-driver): Log errors from connection pool factory

mcheshkov:
**Check List**

`generic-pool` will just silently throw those away, so we should at least log them

### DIFF
--- a/packages/cubejs-jdbc-driver/src/JDBCDriver.ts
+++ b/packages/cubejs-jdbc-driver/src/JDBCDriver.ts
@@ -183,6 +183,14 @@ export class JDBCDriver extends BaseDriver {
       acquireTimeoutMillis: 120000,
       ...(poolOptions || {})
     }) as ExtendedPool;
+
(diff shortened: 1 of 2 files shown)


packages/cubejs-jdbc-driver/src/JDBCDriver.ts
```diff
@@ -183,6 +183,14 @@ export class JDBCDriver extends BaseDriver {
       acquireTimeoutMillis: 120000,
       ...(poolOptions || {})
     }) as ExtendedPool;
+
```
KSDaemon, line 186: If you're already touching this driver. Can you fix lint warning above too? (can't pin my comment there as there are no changes at that lines)

KSDaemon (approved):
👍🏻 LGTM! Left one minor note

packages/cubejs-jdbc-driver/src/JDBCDriver.ts
mcheshkov: Skimmed through, fixed all the minor issues.
Well, one not-so-minor, I guess: `destroy` function in pool's factory didn't actually call `close` on connection, only promisified it.

PR #8906, opened 2024-11-04
fix(schema-compiler): Support member alias for TD with granularity

mcheshkov:
**Check List**

There's no way for SQL API to specify time dimension in load query during SQL generation for CubeScan: only "dimension" part is supported by Cube ATM, and either SQL API should trim granularity from member path, and replicate alias logic from schema compiler, or schema compiler should support granularity in aliases


mcheshkov:
Merged in #8927

PR #8910, opened 2024-11-04
fix(dev_env_setup): use relative packages everywhere

johancube:
**Check List**

This solution ensures that all related packages work together seamlessly without concern for missing links.


dev_env_setup.sh
```diff
@@ -12,145 +12,75 @@ cd "$SCRIPT_DIR"
 echo "Running 'yarn install' in the root directory..."
 yarn install
 
-# Step 2: Run yarn build in the root directory
-echo "Running 'yarn build' in the root directory..."
-yarn build
```
KSDaemon, line 17: originally `yarn build` fires the `rollup` to build all the UMDs/ESMs in packages... You switched to trying running `yarn build` inside each pkg. But does this produce the same output?
Taking package `packages/cubejs-client-core` as an example:
There is no `build` script in `package.json` and there is no `ts` files. So new version won't do anything, while original `rollup` will prepare UMD for it.

dev_env_setup.sh
```diff
@@ -12,145 +12,75 @@ cd "$SCRIPT_DIR"
...
 done
 
-# Step 8: Run yarn link @cubejs-backend/<pkg> in packages/cubejs-server-core
-cd packages/cubejs-server-core
-for driver in "${selected_drivers[@]}"
+echo "Available database types:"
+PS3='Please select the database type: '
+select DB_TYPE in "${db_types[@]}"
 do
-    echo "Linking @cubejs-backend/$driver in packages/cubejs-server-core..."
```
KSDaemon, line 59: Are you sure linking drivers in server-core is no longer needed for proper working?

dev_env_setup.sh
```diff
@@ -12,145 +12,75 @@ cd "$SCRIPT_DIR"
...
-            echo "Linking $pkg..."
-            yarn link "$pkg"
-        done
-    else
-        echo "Don't forget to link packages that you plan to modify inside your project!"
+for package in "$SCRIPT_DIR"/packages/*; do
+    if [ -d "$package" ]; then
+        package_name=$(node -p "require('$package/package.json').name")
+        echo "Linking $package_name..."
+        yarn link "$package_name"
```
KSDaemon, line 78: It's probably good to cleanup the scripts in root package.json to remove potential confusion for those who will find it and try to use but it is outdated and not actual anymore. Like `"link:dev"`

KSDaemon (commented):
👍🏻 Great improvement and simplification! But I think there are some things to be resolved. Left comments inline.

dev_env_setup.sh
johancube: fixed

dev_env_setup.sh
johancube: I think so. This script is linking all the packages together.

package.json
```diff
@@ -23,8 +23,7 @@
     "tsc": "tsc --build",
     "tsc:watch": "tsc --build --watch",
     "clean": "rimraf packages/*/{tsconfig.tsbuildinfo,lib,dist}",
-    "postinstall": "lerna link",
-    "link:dev": "cd ./packages/cubejs-backend-shared && yarn link && cd ../cubejs-backend-cloud && yarn link && cd ../cubejs-backend-native && yarn link && cd ../cubejs-server && yarn link && cd ../cubejs-server-core && yarn link && cd ../cubejs-api-gateway && yarn link && cd ../cubejs-schema-compiler && yarn link && cd ../cubejs-query-orchestrator && yarn link && cd ../cubejs-athena-driver && yarn link && cd ../cubejs-duckdb-driver && yarn link && cd ../cubejs-bigquery-driver && yarn link && cd ../cubejs-postgres-driver && yarn link && cd ../cubejs-databricks-jdbc-driver && yarn link && cd ../cubejs-mssql-driver && yarn link && cd ../cubejs-clickhouse-driver && yarn link && cd ../cubejs-snowflake-driver && yarn link && cd ../cubejs-cubestore-driver && yarn link && cd ../cubejs-templates && yarn link && cd ../cubejs-client-core && yarn link && cd ../cubejs-client-ws-transport && yarn link && cd ../cubejs-playground && yarn link"
```
KSDaemon, line 27: You removed this, but how can developers mark packages as available for linking? Without running `yarn link` without arguments in all packages, you won't be able to link them in the cube project, so e.g. `yarn link @cubejs-backend/server` will fail, because there is no global link created in `packages/cubejs-server`

KSDaemon (approved):
👍🏻 Seems to be good! Let's try it :)

KSDaemon (approved):
👍🏻 Awesome!

PR #8927, opened 2024-11-07
feat(cubesql): Penalize zero members in wrapper

mcheshkov:
**Check List**

This would allow to extract fully assembled `CubeScan` under wrapper instead of `CubeScan(allMembers, ungrouped=true)`.

ATM there are two related components in cost: `non_detected_cube_scans` and `cube_members`
`non_detected_cube_scans` allows to penalize `CubeScan` without members specifically outside the wrapper. This is pretty hard penalty, queries like that are Not Good™
`cube_members` allows to prefer queries will less members, which seems fine. But on it's own it would prefer query with zero member, which is, actually, all the members.
So, new cost component added: `zero_members_wrapper`. It would stand right before `cube_members`, and allow to penalize no-members representation before `cube_members` starts impacting extraction.

New `CubeScan` extractions surfaced a couple of bugs related to aliasing in generated SQL, hence all the supporting stuff in this PR.


waralexrom (approved):

mcheshkov:
Due to our build order issues I can't add smoke test in the same PR, so I'll split it to separate one

paveltiunov (approved):

mcheshkov:
I was confused, and actually I can add a smoke test here! And it was failing for a good reason!
Fixed and updated.

PR #8928, opened 2024-11-07
feat(clickhouse-driver): Switch from apla-clickhouse to @clickhouse/client

mcheshkov:
**Check List**

Switch client library used to talk to ClickHouse to upstream one. Lots of related changes:

* Streaming now does not use streaming JSON parser, so we can't rely on `meta` field in JSON format. Instread it relies on `JSONCompactEachRowWithNamesAndTypes`: first two rows returned should contain names and types. https://clickhouse.com/docs/en/sql-reference/formats#jsoncompacteachrowwithnamesandtypes
* Streaming now use async iterators instread of Node.js streams internally. External API returns stream, as before
* Pooling moved completely to client library. `generic-pool` is not used at all, `dbMaxPoolSize`  is passed to client library to limit open sockets. New client maintains `http.Agent` internally, and have it's own idle timers, looks fine for us.
* Queries now does not send `session_id`, as we anyway expect queries to be independent, and don't use session-bound stuff, like temporary tables. Previous behaviour was kind of weird: session ids were attached to client in pool, but for every query it would pull acquire new client from pool, so nothing could actually utilize same session.
* `KILL QUERY` on cancellation now uses separate client instance, to avoid getting stuck on busy pool
* `query` method supports only `SELECT` queries, or other queries that return result sets. For DDL queries on this client library one have to use other methods. Because of that more overrides were necessary, like `dropTable`,  `createSchemaIfNotExists` or `createTable`.
* Driver now respects per-datasource `dbQueryTimeout` config

### DIFF
--- a/packages/cubejs-backend-shared/src/env.ts
+++ b/packages/cubejs-backend-shared/src/env.ts
@@ -9,7 +9,7 @@ export class InvalidConfiguration extends Error {
   }
 }
 
-export function convertTimeStrToMs(
+export function convertTimeStrToSeconds(
   input: string,
   envName: string,
   description: string = 'Must be a number in seconds or duration string (1s, 1m, 1h).',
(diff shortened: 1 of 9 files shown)


ovr (approved):
LGTM 🔥

paveltiunov (approved):

PR #8929, opened 2024-11-08
fix(cubesql): Allow aggregation pushdown only for unlimited CubeScan

mcheshkov:
**Check List**

`Aggregate(CubeScan(limit=n))` is not equivalent to `CubeScan(aggregation, limit=n)`
Former does aggregation over first n rows, latter returns first n rows after aggregation.

### DIFF
--- a/rust/cubesql/cubesql/src/compile/rewrite/rules/members.rs
+++ b/rust/cubesql/cubesql/src/compile/rewrite/rules/members.rs
@@ -137,8 +137,11 @@ impl RewriteRules for MemberRules {
                         "?old_members",
                         "?filters",
                         "?orders",
-                        "?limit",
-                        "?offset",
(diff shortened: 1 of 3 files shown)


rust/cubesql/cubesql/src/compile/rewrite/rules/members.rs
```diff
@@ -134,8 +134,11 @@ impl RewriteRules for MemberRules {
                         "?old_members",
                         "?filters",
                         "?orders",
-                        "?limit",
-                        "?offset",
+                        // If CubeScan already have limit and offset ot would be incorrect to push aggregation into it
```
KSDaemon, line 137:
```suggestion
                        // If CubeScan already have limit and offset it would be incorrect to push aggregation into it
```

paveltiunov (approved):

PR #8934, opened 2024-11-08
refactor(cubesql): Improve EGraph debugger, part 3: parsing and data preparation in UI side

mcheshkov:
**Check List**

Add proper parsing on UI side, to catch data format changes.
Move most of data preparation related to UI (like a concepts of edge or combo) to UI completely, simplify data collection on rewriter side.

### DIFF
--- a/rust/cubesql/cubesql/egraph-debug-template/src/index.tsx
+++ b/rust/cubesql/cubesql/egraph-debug-template/src/index.tsx
@@ -104,13 +104,16 @@ function prepareStates(states: InputData): Array<PreparedStateData> {
                 } as InputNodeData;
             })
             .concat(
-                egraph.eclasses.map((eclass) => {
-                    return {
-                        id: eclass.id.toString(),
-                        label: eclass.id.toString(),
-                        comboId: `c${eclass.id}`,
-                    } as InputNodeData;
-                }),
+                egraph.eclasses
+                    // render only canonical eclasses to avoid rendering empty nodes and combos for merged ones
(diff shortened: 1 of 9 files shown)


rust/cubesql/cubesql/egraph-debug-template/src/index.tsx
```diff
@@ -18,6 +18,7 @@ import type {
     NodeProps,
 } from 'reactflow';
 import 'reactflow/dist/style.css';
+import { z } from 'zod';
```
KSDaemon, line 21: Probably, not a big deal but still: 
We already have a [joi](joi.dev/) scheme validator in our codebase, so maybe we can reuse it here too? I know, `node_modules` is a black hole and could hold any number of packages) but... If we can reduce and reuse...Why not?

KSDaemon (approved):
👍🏻 LGTM! Left minor note.

rust/cubesql/cubesql/egraph-debug-template/src/index.tsx
mcheshkov: I'd rather switch to `zod` in those places :sweat_smile: 

Joi does not introduce static types in TypeScript along with building parser, so one need to manually support and ensure compatibility for both. There are solutions like `joi-to-typescript`, but, given the usual process of working with codegen in JS, it's not very comfortable.

Also egraph debugger is a completely separate package from monorepo, so it wouldn't pollute anything with unnecessary dependencies in "usual" case, and it would install them second time even if dependencies were same.

Issue #8920, opened 2024-11-06
Duckdb driver should look for credentials in the default provider chain for AWS S3


GabeInDevOps:
Rather than needing to create, manage and pass an AWS Access and Secret key for the Duckdb driver to read from S3, it should look for credentials in the default provider chain. This is best practice and helps increase the security posture of Cube.

This can be fixed by running the following against duckdb:
```
CREATE SECRET (
      TYPE S3,
      PROVIDER CREDENTIAL_CHAIN
  );
```

PR #8935, opened 2024-11-08
use default credential provider chain for duckdb s3 access

GabeInDevOps:
attempt at resolving issue [#8920](https://github.com/cube-js/cube/issues/8920)

### DIFF
--- a/packages/cubejs-backend-shared/src/env.ts
+++ b/packages/cubejs-backend-shared/src/env.ts
@@ -1613,6 +1613,16 @@ const variables: Record<string, (...args: any) => any> = {
     ]
   ),
 
+   duckdbS3UseCredentialChain: ({
+      dataSource
+    }: {
+      dataSource: string,
+    }) => (
+      process.env[
+        keyByDataSource('CUBEJS_DB_DUCKDB_S3_USE_CREDENTIAL_CHAIN', dataSource)
+      ]
+    ),
+
   /**
    * Presto catalog.
    */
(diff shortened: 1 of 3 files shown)


igorlukanin:
@GabeInDevOps Thanks for the contribution! Before it's merged, we'll need an update to docs (both for the env var reference page and the DuckDb page). Thank you in advance!

packages/cubejs-backend-shared/src/env.ts
```diff
@@ -1613,6 +1613,16 @@ const variables: Record<string, (...args: any) => any> = {
...
   ),
 
+   duckdbS3UseCredentialChain: ({
+      dataSource
+    }: {
+      dataSource: string,
+    }) => (
+      process.env[
+        keyByDataSource('CUBEJS_DB_DUCKDB_S3_USE_CREDENTIAL_CHAIN', dataSource)
+      ]
```
KSDaemon, line 1623: It would be better to parse and return a boolean here. This will simplify further checks. Have a look at this example:
```js
  snowflakeSessionKeepAlive: ({
    dataSource
  }: {
    dataSource: string,
  }) => {
    const val = process.env[
      keyByDataSource(
        'CUBEJS_DB_SNOWFLAKE_CLIENT_SESSION_KEEP_ALIVE',
        dataSource,
      )
    ];
    if (val) {
      if (val.toLocaleLowerCase() === 'true') {
        return true;
      } else if (val.toLowerCase() === 'false') {
        return false;
      } else {
        throw new TypeError(
          `The ${
            keyByDataSource(
              'CUBEJS_DB_SNOWFLAKE_CLIENT_SESSION_KEEP_ALIVE',
              dataSource,
            )
          } must be either 'true' or 'false'.`
        );
      }
    } else {
      return true;
    }
  },
```

This way, you can also control the default value if the env var is not set.

KSDaemon:
Gentle ping @GabeInDevOps. Could you please rebase on top of the latest master and wdyt about suggested changes?

PR #8937, opened 2024-11-09
fix(schema-compiler): fix FILTER_PARAMS to populate set and notSet filters in cube's SQL query

sumeetgajjar:
The presence of the following filters in a cube query: `set`, `notSet`, `eq NULL` or `NOT eq NULL` renders the corresponding FILTER_PARAMS as `1 = 1` irrespective of the cube or filter.

This is because the `renderFilterParams` method skips invoking `filter.conditionSql` when converting a `filter` to its corresponding SQL due to an empty `filterParams` array. 

For `set` and `notSet`, the filter cannot have value(s). 
For `eq NULL` or `NOT eq NULL` the `null` value is filtered by: https://github.com/cube-js/cube/blob/349597a4a1b10f6e76ef3bf18c37986cdd0962f8/packages/cubejs-schema-compiler/src/adapter/BaseFilter.ts#L115

This leads to an empty `filterParams` array, thus, skipping the filter SQL, thereby returning `1 = 1`.

This PR aims to fix the bug by restructuring the nested if-else block, thereby also improving the readability of the same.

**Check List**

### DIFF
--- a/packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
+++ b/packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
@@ -3809,21 +3809,20 @@ export class BaseQuery {
     if (!filterParamArg) {
       throw new Error(`FILTER_PARAMS arg not found for ${filter.measure || filter.dimension}`);
     }
-    if (
-      filterParams && filterParams.length
-    ) {
-      if (typeof filterParamArg.__column() === 'function') {
-        // eslint-disable-next-line prefer-spread
-        return filterParamArg.__column().apply(
-          null,
-          filterParams.map(allocateParam),
-        );
-      } else {
-        return filter.conditionSql(filterParamArg.__column());
-      }
-    } else {
+
+    if (typeof filterParamArg.__column() !== 'function') {
+      return filter.conditionSql(filterParamArg.__column());
+    }
+
+    if (!filterParams || !filterParams.length) {
       return '1 = 1';
     }
+
+    // eslint-disable-next-line prefer-spread
+    return filterParamArg.__column().apply(
+      null,
+      filterParams.map(allocateParam),
+    );
   }
 
   filterGroupFunction() {
--- a/packages/cubejs-schema-compiler/test/unit/base-query.test.ts
+++ b/packages/cubejs-schema-compiler/test/unit/base-query.test.ts
@@ -1762,6 +1762,84 @@ describe('SQL Generation', () => {
       expect(cubeSQL).toMatch(/\(\s*\(.*type\s*=\s*\$\d\$.*OR.*type\s*=\s*\$\d\$.*\)\s*AND\s*\(.*type\s*=\s*\$\d\$.*OR.*type\s*=\s*\$\d\$.*\)\s*\)/);
     });
 
+    it('equals NULL filter', async () => {
+      await compilers.compiler.compile();
+      const query = new BaseQuery(compilers, {
+        measures: ['Order.count'],
+        filters: [
+          {
+            and: [
+              {
+                member: 'Order.type',
+                operator: 'equals',
+                values: [null],
+              },
+            ]
+          }
+        ],
+      });
+      const cubeSQL = query.cubeSql('Order');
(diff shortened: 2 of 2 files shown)


sumeetgajjar:
Hi @KSDaemon, @RusovDmitriy, @paveltiunov - can you please review this PR?

Thanks in advance.

KSDaemon:
Hi @sumeetgajjar Thank you for contributing! I'll have a look!

sumeetgajjar:
3 tests under `postgres/pre-aggregations.test.js` are failing.

This is primarily because the `visitors` cube during the pre-aggregate generation uses the following SQL:
```
select * from visitors WHERE ((created_at >= undefined AND created_at <= undefined))\n' +
            '      AND (1 = 1)\n' +
```

Earlier (before this change) SQL was 
```
select * from visitors WHERE ((1 = 1))\n' +
            '      AND (1 = 1)\n' +
```

---

```js
cube(`visitors`, {
      sql: `
      select * from visitors WHERE ${FILTER_PARAMS.visitors.createdAt.filter('created_at')}
      AND ${FILTER_PARAMS.ReferenceOriginalSql.createdAt.filter('created_at')}
      `,
```

---

This makes sense as the time dimension filter values are unavailable during pre-aggregate generation.

sumeetgajjar:
> 3 tests under `postgres/pre-aggregations.test.js` are failing.
> 
> This is primarily because the `visitors` cube during the pre-aggregate generation uses the following SQL:
> 
> ```
> select * from visitors WHERE ((created_at >= undefined AND created_at <= undefined))\n' +
>             '      AND (1 = 1)\n' +
> ```
> 
> Earlier (before this change) SQL was
> 
> ```
> select * from visitors WHERE ((1 = 1))\n' +
>             '      AND (1 = 1)\n' +
> ```
> 
> ```js
> cube(`visitors`, {
>       sql: `
>       select * from visitors WHERE ${FILTER_PARAMS.visitors.createdAt.filter('created_at')}
>       AND ${FILTER_PARAMS.ReferenceOriginalSql.createdAt.filter('created_at')}
>       `,
> ```
> 
> This makes sense as the time dimension filter values ~are~ may be unavailable during pre-aggregate generation.

Given each of the DATE_OPERATORS_Where methods (i.e. inDateRangeWhere, notInDateRangeWhere, etc) exactly knows its required parameters i.e.
- for `inDateRangeWhere` both `from` and `to` are required
- for `beforeDateWhere` only `before` is required
- etc

One fix could be modifying each DATE_OPERATORS_Where method to return `1 = 1` when any required parameters are undefined.

@KSDaemon - Does the above proposal sound good to you?

sumeetgajjar:
> Hi @sumeetgajjar Thank you for contributing! I'll have a look!

Hi @KSDaemon - gentle ping, can you please take a look?

This bug is directly impacting our production workflows and is leading to incorrect data being shown to the users.

KSDaemon:
> Does the above proposal sound good to you?

That makes sense.

Seems to be good. I relaunched tests

KSDaemon (approved):
👍🏻 LGTM!

sumeetgajjar:
> 👍🏻 LGTM!

Thanks for the review @KSDaemon.

A couple of questions,
- can you please help us understand the release process and when will the above change be available for public consumption?
- can this fix be applied to the LTS branch?

sumeetgajjar:
Hi @KSDaemon, @paveltiunov, @igorlukanin - can you please help us merge this PR given it is already approved 😄?

KSDaemon:
Hi @sumeetgajjar, Sorry for the late answers. Yeah, everything looks right. I think the one missing piece is real integration tests. Please add some tests to `packages/cubejs-schema-compiler/test/integration/postgres/sql-generation.test.ts` to ensure that real queries are created and run successfully. Thanks!

sumeetgajjar:
Hi @KSDaemon - I've added the Postgres integration tests.
Can you please take a look?

KSDaemon:
Hi @sumeetgajjar Thank you for providing the tests! That's great! Merging!

PR #8945, opened 2024-11-12
feat(api-gateway): Meta - expose aliasMember for members in cube

ovr:
Motivation:

While I was working on the new layer for MDX, I discovered that all members in views are defined as View's members. It leads to a problem, that I am not able to separate it by cubes to align our model with Analysis Services.


KSDaemon (approved):

MazterQyou (approved):

aquariuslt:
hi,
is `aliasMember` visible in `@cubejs/client-core` TCubeDimension type?

PR #8948, opened 2024-11-12
Use the Trino client instead of inheriting from Presto

ndrluis:
Hello, I'm opening this PR, but I don't have expertise in TypeScript and JavaScript. I've tried my best to solve the problem, but I need some help to make the test suite work.

This PR aims to decouple the Trino driver from Presto, which uses an outdated client that is quite slow. I ran some tests comparing only the clients, and the Presto driver currently used by Cube (which is not the official one) takes about 3 seconds to execute a SELECT 1 query, while the Trino driver completes it in 300ms.

Additionally, there are features in the Presto driver that are not supported by Trino, such as unloading data through GCS. Therefore, we need to isolate each of these drivers.

**Check List**

### DIFF
--- a/.github/actions/smoke.sh
+++ b/.github/actions/smoke.sh
@@ -45,9 +45,9 @@ echo "::endgroup::"
 #yarn lerna run --concurrency 1 --stream --no-prefix smoke:prestodb
 #echo "::endgroup::"
 
-echo "::group::Trino"
-yarn lerna run --concurrency 1 --stream --no-prefix smoke:trino
-echo "::endgroup::"
+#echo "::group::Trino"
+#yarn lerna run --concurrency 1 --stream --no-prefix smoke:trino
+#echo "::endgroup::"
 
 echo "::group::MS SQL"
 yarn lerna run --concurrency 1 --stream --no-prefix smoke:mssql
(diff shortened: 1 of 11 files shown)


igorlukanin:
Hi @ndrluis 👋 I think this change was long awaited for, thanks for the contribution!

For the docs, if you could update this page with new env vars and check for other instances of "Trino", that would be fantastic: https://cube.dev/docs/reference/configuration/environment-variables

paveltiunov:
@ndrluis Hey Andre! Thanks for contributing! I feel it's right direction however in order to make this switch Trino should meet parity in terms of features. At the first sight at least unload support is missing. We need to have in order to land this change.

ndrluis:
@paveltiunov I understand the Presto implementation, but as a Trino user, I believe it makes more sense to wait for the [spooled client protocol](https://www.linkedin.com/posts/mateuszgajewski_trino-trino-performance-activity-7244658731991392258-0udu), which solves this problem for any catalog. The current solution only works for GCS and, as far as I know, only for the Hive catalog.

I fixed the query method and removed the incorrect stream method, but I'm unsure of the best way to implement and test it, since the driver works with just the query method. Which cube feature uses the stream method?

PR #8952, opened 2024-11-13
fix(cubesql): fix unhandled timestamp unwrapping in df/transform_response

KSDaemon:
This PR fixes timestamp parsing handling unwrapping that may lead to panic:
```
thread 'tokio-runtime-worker' panicked at cube/rust/cubesql/cubesql/src/compile/engine/df/scan.rs:1133:86:
called `Option::unwrap()` on a `None` value
```

**Check List**

### DIFF
--- a/rust/cubesql/cubesql/src/compile/engine/df/scan.rs
+++ b/rust/cubesql/cubesql/src/compile/engine/df/scan.rs
@@ -1135,10 +1135,16 @@ pub fn transform_response<V: ValueObject>(
                                     ))
                                 })?;
                             // TODO switch parsing to microseconds
-                            if timestamp.timestamp_millis() > (((1 as i64) << 62) / 1_000_000) {
+                            if timestamp.timestamp_millis() > (((1i64) << 62) / 1_000_000) {
                                 builder.append_null()?;
+                            } else if let Some(nanos) = timestamp.timestamp_nanos_opt() {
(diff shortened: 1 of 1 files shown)


rust/cubesql/cubesql/src/compile/engine/df/scan.rs
```diff
@@ -1135,10 +1135,12 @@ pub fn transform_response<V: ValueObject>(
...
                                 })?;
                             // TODO switch parsing to microseconds
-                            if timestamp.timestamp_millis() > (((1 as i64) << 62) / 1_000_000) {
+                            if timestamp.timestamp_millis() > (((1i64) << 62) / 1_000_000) {
                                 builder.append_null()?;
+                            } else if let Some(nanos) = timestamp.timestamp_nanos_opt() {
+                                builder.append_value(nanos)?;
                             } else {
-                                builder.append_value(timestamp.timestamp_nanos_opt().unwrap())?;
+                                builder.append_null()?;
```
ovr, line 1143: Another problem is keeping logs clean. We need to log it. Without that, it will be a silent surprise that some values in the column are null.

KSDaemon: Yeah, agree! Thnx for pointing! Added err logging!

PR #8956, opened 2024-11-14
feat(dremio-driver): Add Dremio Cloud Support

### LINKED ISSUE #8166: Creare possbility to connect to Dremio Cloud Driver
Dremio has evolved on past years, adopting project and catalog perspectives such as a lakehouse platform. This create a need for a new connector in cube that help to connect to Dremio Cloud, using the new and improved API's.

image

Dremio Cloud has a different documentation over Dremio Core and concepts such as `username` and `database` are not used in this solution. Also some ports are a bit different if compared to Dremio Core. 

There is a need for changing the Dremio Driver and Dremio Query components to adjust to new dremio capabilities. On older versions of Dremio Core there was a `username` that was used to authenticate on APIs. This information does not exists for Dremio Cloud, and the connection to Dremio SQL API and Dremio Jobs API is authenticated by project id and PAT.

Here is a snippet of code using python language for the newest jobs API:

```
def make_dremio_call(
        self,
        query: str,
        token: str
    ) -> DataFrame:
    '''
    This method makes a query on dremio's warehourse and and returns a DataFrame

    Args:
    - query (str): The query to be processed by dremio
    - token: the personal access token provided by dremio

    Returns:
    - df (DataFrame): a dataframe with the result of the query
    '''
    if self.dremio_project_id == '':
        raise ValueError('Please set dremio_project_id parameter on APICaller class.')
    #Trigger the job to run
    endpoint = f'v0/projects/{self.dremio_project_id}/sql'
    headers = {
        'Authorization': f'Bearer {token}',
        'Content-Type': 'application/json'
    }
    payload = {
        'sql':query
    }
    
    payload = json.dumps(payload)

    job_id = requests.post(
        url = self.dremio_uri + endpoint,
        headers = headers,
        data = payload,
    ).json()['id']
```

As you can see, the only keys used are `project_id` and `personal_access_token`.

The most interesting way to fix this problem is updating the code related to driver and query components adapting it to  the newest docuementation of Dremio Cloud.

**Additional context**
[Older API Documentation - Currently used in Cube Cloud]()
[Newest API Documentation for Dremio Cloud](https://docs.dremio.com/cloud/reference/api/job/)

maladroitthief:
**Check List**

closes #8166

- Adds support to cubejs-dremio-driver for Dremio Cloud
- Adds support for personal access tokens for Dremio Server


packages/cubejs-dremio-driver/README.md
```diff
@@ -9,11 +9,24 @@
...
+## Dremio Cloud
+
+To use this driver with [Dremio Cloud](https://docs.dremio.com/cloud/reference/api/), use the following setup:
+
+| Environment Variable | Value                                              |
+| -------------------- | -------------------------------------------------- |
+| CUBEJS_DB_TYPE       | dremio                                             |
+| CUBEJS_DB_URL        | https://api.dremio.cloud/v0/projects/${PROJECT_ID} |
+| CUBEJS_DB_NAME       | ${DB_NAME}                                         |
+| CUBEJS_DB_PASS       | Bearer ${PERSONAL_ACCESS_TOKEN}                    |
```
KSDaemon, line 21: Let's not mess everything in one place and introduce another special env var for token. You can do it in https://github.com/cube-js/cube/blob/master/packages/cubejs-backend-shared/src/env.ts 
Have a look at Pinot / Startree Auth Token for example.
Something like this:
```js
  /** ***************************************************************
   * Dremio Driver                                                  *
   **************************************************************** */

  /**
   * Dremio cloud Auth Token
   */
  dremioAuthToken: ({
    dataSource,
  }: {
    dataSource: string,
  }) => (
    process.env[
      keyByDataSource('CUBEJS_DB_DREMIO_AUTH_TOKEN', dataSource)
    ]
  ),

```

packages/cubejs-dremio-driver/README.md
```diff
@@ -9,11 +9,24 @@
...
+
+> It's important to note that "Bearer" is required for using personal access tokens.
```
KSDaemon, line 23: Having a special env just for the token itself also will make it more user-friendly and convenient for usage: just copy/paste the token without need to remember to add `Bearer ` prefix. You can add it in the code because you know exact flow when it's needed

packages/cubejs-dremio-driver/driver/DremioDriver.js
```diff
@@ -80,11 +84,24 @@ class DremioDriver extends BaseDriver {
...
+    } else {
+      const protocol = (this.config.ssl === true || this.config.ssl === 'true')
+        ? 'https'
+        : 'http';
+      this.config.url = `${protocol}://${this.config.host}:${this.config.port}`;
+      this.config.apiVersion = '/api/v3';
+    }
+
+    if (this.config.auth === 'PAT' || this.config.password.startsWith('Bearer ')) {
+      this.config.auth = 'PAT';
```
KSDaemon, line 101: `if (this.config.auth === 'PAT') then this.config.auth = 'PAT'` looks a bit weird. 
And what if `this.config.auth === 'PAT'` but there is no `password.startsWith('Bearer ')`? 

Ideally, if both `dbUrl` and `token` are required for the correct connection configuration, the code should check it and throw an error if one is missing.

KSDaemon: Also, readme should state that `dbUrl` takes precedence, and if it is set — then the driver will try to init using dbUrl+token flow.

packages/cubejs-dremio-driver/driver/DremioDriver.js
```diff
@@ -80,11 +84,24 @@ class DremioDriver extends BaseDriver {
...
+    } else {
+      const protocol = (this.config.ssl === true || this.config.ssl === 'true')
+        ? 'https'
+        : 'http';
+      this.config.url = `${protocol}://${this.config.host}:${this.config.port}`;
+      this.config.apiVersion = '/api/v3';
+    }
+
+    if (this.config.auth === 'PAT' || this.config.password.startsWith('Bearer ')) {
```
KSDaemon, line 100: These if-else blocks can be united into one with all the required checks and var sets.

KSDaemon (changes requested):
Hi @maladroitthief! Thanks for your contribution! 
I left some comments how to make it cleaner and more aligned with the rest codebase.

packages/cubejs-dremio-driver/README.md
maladroitthief: Good idea, change has been made to support this.

packages/cubejs-dremio-driver/driver/DremioDriver.js
maladroitthief: I added a note in the readme to point out the precedence and added the error in the case that the token is blank. Let me know if there is a more preferred way for throwing errors, I couldn't find any existing examples.

packages/cubejs-dremio-driver/driver/DremioDriver.js
maladroitthief: I ended up removing the `auth` variable and just check for the dremioAuthToken instead.

PR #9013, opened 2024-12-03
fix(api-gateway): support proxy when fetching jwk for token validation

johnsca:
**Check List**

When fetching the JWK data needed to validate the auth token, the environment proxy settings are not honored.


igorlukanin:
@johnsca Thanks for this contribution!

> the environment proxy settings are not honored

Just to double check, which proxy settings exactly are not honored? Also, which platform are we talking about here?

johnsca:
@igorlukanin
> Just to double check, which proxy settings exactly are not honored? Also, which platform are we talking about here?

No proxy settings were being honored when using JWT authentication (specifically, when retrieving the JWK keys to validate the token). This was encountered when deploying Cube in an environment with restricted outbound traffic which requires request to the public internet to go out via a proxy for security reasons.

This PR uses `ProxyAgent` because it supports the `HTTPS_PROXY` and `NO_PROXY` env vars, which the lower-level `HttpsProxyAgent` used by [getHttpAgentForProxySettings()](https://github.com/cube-js/cube/blob/d8d8de24a7f5f104c858ce9555a0231e7552e917/packages/cubejs-backend-shared/src/proxy.ts#L27-L31) doesn't appear to. It might be worth moving that code to `ProxyAgent` as well, but I wanted to keep this PR minimal.

KSDaemon:
Hi, @johnsca. Thanks for finding this issue and suggesting a fix. However, the right way to fix this, as you already mentioned,  is to patch [getProxySettings()](https://github.com/cube-js/cube/blob/d8d8de24a7f5f104c858ce9555a0231e7552e917/packages/cubejs-backend-shared/src/proxy.ts#L17), making it aware of common env vars like `HTTTP_PROXY` / `HTTPS_PROXY`. And then reuse it in `api-gateway`

packages/cubejs-api-gateway/src/jwk.ts
```diff
@@ -51,7 +52,7 @@ export type JWKsFetcherOptions = Pick<BackgroundMemoizeOptions<any, any>, 'onBac
 
 export const createJWKsFetcher = (jwtOptions: JWTOptions, options: JWKsFetcherOptions) => {
   const fetchJwkUrl = asyncMemoizeBackground(async (url: string) => {
-    const response = await asyncRetry(() => fetch(url), {
+    const response = await asyncRetry(() => fetch(url, { agent: new ProxyAgent() }), {
```
KSDaemon, line 55: We need to reuse `getHttpAgentForProxySettings()` from backend-shared here instead.

KSDaemon (changes requested):

johnsca:
@KSDaemon It doesn't seem like a good idea to reimplement what `proxy-from-env` does, which is what `ProxyAgent` uses, but `proxy-from-env` doesn't have an interface that can be used to directly convert `getProxySettings` to use (it unfortunately puts all of the proxy settings logic [directly into getProxyForUrl](https://github.com/Rob--W/proxy-from-env/blob/095d1c26902f37a12e22bea1b8c6b67bf13fe8cd/index.js#L24-L52) which is unfortunate). So I instead switched `getHttpAgentForProxySettings` to use `ProxyAgent` instead and just dropped the use of `getProxySettings`. That should be a drop-in replacement, and it also aligns better with the recommendation from the [proxy-agents monorepo](https://github.com/TooTallNate/proxy-agents?tab=readme-ov-file#nodejs-http-proxy-agents-monorepo) to prefer using `proxy-agent` rather than the lower-level implementations directly, such as `https-proxy-agent`.

I don't see anywhere that `getProxySettings` is used other than that one file but since it's exported, I didn't feel comfortable removing it even though it's unused now.

PR #9018, opened 2024-12-05
feat(schema-compiler): folders support, hierarchies improvements

vasilev-alex:


packages/cubejs-schema-compiler/test/unit/fixtures/folders.yml
```diff
@@ -0,0 +1,105 @@
...
+      - name: orders_hierarchy
+        levels:
+          - "{CUBE}.status"
+          - number
+          - users.city
+      - name: some_other_hierarchy
+        title: Some other hierarchy
+        levels:
+          - users.state
+          - users.city
```
igorlukanin, line 39: @paveltiunov Is it OK to include dimensions from other cubes in a hierarchy? Of course, it is possible to do so via a proxy dimension, I just ant to double check.

packages/cubejs-schema-compiler/test/unit/fixtures/folders.yml
```diff
@@ -0,0 +1,105 @@
...
+
+      - name: city
+        sql: city
+        type: string
+    hierarchies:
+      - name: orders_hierarchy
+        levels:
+          - "{CUBE}.status"
+          - number
+          - users.city
```
igorlukanin, line 34: @vasilev-alex Do we verify that only dimensions can be part of a hierarchy? (And measures/segments can't.)

packages/cubejs-schema-compiler/src/compiler/CubeValidator.ts
```diff
@@ -745,17 +745,25 @@ const baseSchema = {
...
   accessPolicy: Joi.array().items(RolePolicySchema.required()),
 };
 
 const cubeSchema = inherit(baseSchema, {
   sql: Joi.func(),
   sqlTable: Joi.func(),
+  hierarchies: Joi.array().items(Joi.object().keys({
+    name: Joi.string().required(),
+    title: Joi.string(),
+    public: Joi.boolean().strict(),
```
igorlukanin, line 764: @vasilev-alex Have we ensured that hierarchies are covered with RBAC?

PR #9021, opened 2024-12-06
fix(api-gateway): make sure DAP works with sql pushdown

bsod90:
fixes an issue when processing queries with member expressions

**Check List**

[For example #12]

[Description goes here]

### DIFF
--- a/packages/cubejs-api-gateway/src/gateway.ts
+++ b/packages/cubejs-api-gateway/src/gateway.ts
@@ -1185,7 +1185,8 @@ class ApiGateway {
     let normalizedQueries: NormalizedQuery[] = await Promise.all(
       queries.map(
         async (currentQuery) => {
-          const hasExpressionsInQuery = this.hasExpressionsInQuery(currentQuery);
+          const hasExpressionsInQuery =
+            this.hasExpressionsInQuery(currentQuery);
 
           if (hasExpressionsInQuery) {
             if (!memberExpressions) {
@@ -1195,7 +1196,15 @@ class ApiGateway {
             currentQuery = this.parseMemberExpressionsInQuery(currentQuery);
           }
 
-          const normalizedQuery = normalizeQuery(currentQuery, persistent);
+          let normalizedQuery = normalizeQuery(currentQuery, persistent);
+
+          if (hasExpressionsInQuery) {
+            // We need to parse/eval all member expressions early as applyRowLevelSecurity
+            // needs to access the full SQL query in order to evaluate rules
+            normalizedQuery =
+              this.evalMemberExpressionsInQuery(normalizedQuery);
+          }
+
           // First apply cube/view level security policies
           const queryWithRlsFilters = await compilerApi.applyRowLevelSecurity(
             normalizedQuery,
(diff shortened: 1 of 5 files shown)


packages/cubejs-api-gateway/src/gateway.ts
```diff
@@ -1207,8 +1212,8 @@ class ApiGateway {
             context,
           );
 
-          if (hasExpressionsInQuery) {
-            rewrittenQuery = this.evalMemberExpressionsInQuery(rewrittenQuery);
+          if (this.hasExpressionsInQuery(rewrittenQuery)) {
+            rewrittenQuery = this.parseMemberExpressionsInQuery(rewrittenQuery);
```
mcheshkov, line 1216: I've not yet checked yet what can `applyRowLevelSecurity` introduce

But `parseMemberExpressionsInQuery` is not enough anyway.
There's three states of member expressions:

1. `MemberExpression`
Object with `expression` method, that takes cubes as symbols and returns SQL

2. `ParsedMemberExpression`
Object with `expression` as array of strings. Essentially JSON-compatible way of storing `MemberExpression`

3. Just a string with JSON-serialzed `ParsedMemberExpression`.

Schema compiler expects only `MemberExpression`.
`parseMemberExpressionsInQuery` turns strings in proper places in query to `ParsedMemberExpression`.
`evalMemberExpressionsInQuery` turns `ParsedMemberExpression` to `MemberExpression` in those places.

So, depending on what `applyRowLevelSecurity` can add either only `evalMemberExpressionsInQuery`, or both of `parseMemberExpressionsInQuery` + `evalMemberExpressionsInQuery` are necessary.

packages/cubejs-api-gateway/src/gateway.ts
```diff
@@ -1195,7 +1195,12 @@ class ApiGateway {
             currentQuery = this.parseMemberExpressionsInQuery(currentQuery);
           }
 
-          const normalizedQuery = normalizeQuery(currentQuery, persistent);
+          let normalizedQuery = normalizeQuery(currentQuery, persistent);
+
+          if (hasExpressionsInQuery) {
```
mcheshkov, line 1200: Please, add comment about why early/second call is necessary. I'll probably forget about this whole issue next week :sweat_smile:

mcheshkov (changes requested):

packages/cubejs-api-gateway/src/gateway.ts
bsod90: `applyRowLevelSecurity` only adds `MemberExpression` objects and only in one special case: when the entire query should be denied, we add `WHERE 1 == 0`

bsod90: Correction: we actually add string member expressions: [
](https://github.com/cube-js/cube/blob/master/packages/cubejs-server-core/src/core/CompilerApi.js#L320)

bsod90: https://github.com/cube-js/cube/blob/master/packages/cubejs-server-core/src/core/CompilerApi.js#L320

mcheshkov: `expression: () => '1 = 0',`
That's `MemberExpression`, with `expression` as method, returnung SQL

String member expression would look like this: `query.segments.push('{"cubeName":"foo","expr": "1=0"}');`
So, for this case no need to add any additional calls after `applyRowLevelSecurity`

bsod90: Let me see if I can just add it as a parsed version and avoid this messy logic altogether

bsod90: Sorry, I've removed all my previous comment and just added the `eval`. Looks like it should be fine now

mcheshkov (approved):

PR #9055, opened 2024-12-17
feat(gateway): return usedPreAggregations even in prod

flochaz:
**Check List**

https://github.com/cube-js/cube/issues/9078

We are heavily relying on pre-aggregation and relying on job API to force rebuilt of pre aggregation. 
The API support cubes or preAggregation parameter but there is no way to know what cube or preAggregation is used by a given query


igorlukanin:
Hi @flochaz 👋 

Thanks for the contribution. Could you please explain what are you trying to achieve with this change? What is your use case?

packages/cubejs-docker/local.Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG DEV_BUILD_IMAGE=cubejs/cube:build
+ARG DEV_BUILD_IMAGE=cubejs/cube:dev
```
igorlukanin, line 1: I wonder why is this change needed.

KSDaemon: After quick search — it seems that we do not produce `cubejs/cube:build` tags anymore, so I think that one was just missed. So it is ok!

PR #9061, opened 2024-12-19
chore: Remove the length check for pre-aggregation table name

igorlukanin:
Remove the length check for pre-aggregation table name (including schema) as redundant since external pre-aggregations in Cube Store don't need a low length limit

**Check List**


igorlukanin:
We'll need to take another route here.

PR #9068, opened 2025-01-06
fix(api-gateway): Allow querying time dimensions with custom granularity

LFischerstrom:
**Check List**

It's possible to query time dimensions with granularity as a normal dimension but queries with `quarter` were rejected.


igorlukanin:
Thanks @LFischerstrom 🙌 

Honestly, I didn't even know this is possible to query dimensions like this:
Screenshot 2025-01-08 at 10 21 50

I feel like we'd at least need a test for this.

LFischerstrom:
@igorlukanin Ok, it's documented here https://cube.dev/docs/product/apis-integrations/rest-api/query-format

It also claims to support custom granularities which I think would rejected by the input validation.

igorlukanin:
Yeah, you're right:

> In the case of a dimension of the time type, a granularity could be optionally added to the name, in the following format: cube_name.time_dimension_name.granularity_name, e.g., stories.time.week. It can be one of the [default granularities](https://cube.dev/docs/product/data-modeling/concepts#time-dimensions) (e.g., year or week) or a [custom granularity](https://cube.dev/docs/reference/data-model/dimensions#granularities).

And I think you're also right wrt custom granularities. I wonder what is the best fix to that. A wildcard match? I bet @KSDaemon definitely has an opinion on this.

packages/cubejs-api-gateway/src/query.js
```diff
@@ -40,7 +40,7 @@ const getPivotQuery = (queryType, queries) => {
 
 const id = Joi.string().regex(/^[a-zA-Z0-9_]+\.[a-zA-Z0-9_]+$/);
 const idOrMemberExpressionName = Joi.string().regex(/^[a-zA-Z0-9_]+\.[a-zA-Z0-9_]+$|^[a-zA-Z0-9_]+$/);
-const dimensionWithTime = Joi.string().regex(/^[a-zA-Z0-9_]+\.[a-zA-Z0-9_]+(\.(second|minute|hour|day|week|month|year))?$/);
+const dimensionWithTime = Joi.string().regex(/^[a-zA-Z0-9_]+\.[a-zA-Z0-9_]+(\.(second|minute|hour|day|week|month|quarter|year))?$/);
```
KSDaemon, line 43: Yeah, unfortunately, this does not support custom granularities. 

Schema validations are kinda static, meaning defined once and not runtime-dynamic. So the only [easy] way to support it is to change it to simple id regex (actually we made it in other places):
```suggestion
const dimensionWithTime = Joi.string().regex(/^[a-zA-Z0-9_]+\.[a-zA-Z0-9_]+(\.[a-zA-Z0-9_]+)?$/);
```

KSDaemon (commented):
@LFischerstrom Let's update it like this. So it will support all granularities :)

It would be awesome if you can add a test for querying time dimension with some granulaties. It should not be hard. Have a look at https://github.com/cube-js/cube/blob/master/packages/cubejs-api-gateway/test/index.test.ts#L306-L339 as example. We just need to check that query is allowed by api qateway.

KSDaemon:
@LFischerstrom Hi! Hm.... Nothing special is required. What tests are you trying to run? What package?
Btw, I see lint fails, you can fix lint errors with `yarn run lint:fix` in root.

KSDaemon:
@LFischerstrom Try to follow the commands described in https://github.com/cube-js/cube/blob/master/CONTRIBUTING.md#cube-server

LFischerstrom:
Thanks, I tried to running `yarn build` instead of `yarn tsc`. After using `tsc` it worked.

LFischerstrom:
@KSDaemon Alright, the tests pass for me so if you think the new test looks good I think this is ready.

KSDaemon (approved):
👍🏻 I think this looks good! @LFischerstrom Thank you for adding the tests!

PR #9079, opened 2025-01-08
fix(dremio-driver): Fix stale version of testing-shared dependency

mcheshkov:
**Check List**

`dremio-driver` has old version of intra-repo dependency, and is stuck to it: lerna won't upgrade it, thinking that's intended. Also this is the reason of unnecessary diff in `yarn.lock`.
Strangely, we are using `yarn install --frozen-lockfile` in CI, but it finish successfully, and does not change lockfile.


KSDaemon (approved):
👍🏻

PR #9083, opened 2025-01-08
fix(schema-compiler): Handle member expressions in keyDimensions

mcheshkov:
**Check List**

If dimension is member expression, `d.dimension` would be object, and key would be `[object Object]` string, so separate dimensions would have same key, and all but one will be eliminated


KSDaemon (approved):
👍🏻  LGTM!

PR #9085, opened 2025-01-09
fix(clickhouse-driver): Support overriding Username & Password from Driver Config

danieljaniga:
Add the ability to override the username and password from the driver configuration. Currently, when using multiple data sources for different tenants ([Multitenancy vs Multiple Data Sources](https://cube.dev/docs/product/configuration/advanced/multitenancy#multitenancy-vs-multiple-data-sources)) it is not possible to use the username and password defined in the `ClickHouseDriver` config instead of the default ENV variables CUBEJS_DB_USER, CUBEJS_DB_PASS

**Check List**


KSDaemon:
Hey @danieljaniga! Thanks for the contribution!

PR #9124, opened 2025-01-22
fix(schema-compiler): Fix sql generation for rolling_window queries with multiple time dimensions"

KSDaemon:
This PR fixes the incorrect SQL generation for queries that reference measures with rolling_windows and a few (more than one) time timensions with granularities.

**Check List**


packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
```diff
@@ -1495,10 +1540,27 @@ export class BaseQuery {
...
+    const forSelect = this.overTimeSeriesForSelect(cumulativeMeasures, dateSeriesGranularity);
+    return `SELECT ${forSelect} FROM ${dateSeriesSql}` +
+      ` LEFT JOIN (${baseQuery}) ${this.asSyntaxJoin} ${baseQueryAlias} ON ${dateJoinConditionSql}` +
+      this.groupByClause();
+  }
+
+  overTimeSeriesForSelect(cumulativeMeasures, dateSeriesGranularity) {
+    return this.dimensions
+      .map(s => s.cumulativeSelectColumns())
+      .concat(this.timeDimensions.map(d => d.dateSeriesSelectColumn(null, dateSeriesGranularity)))
```
mcheshkov, line 1558: This call does not match signature in `BaseTimeDimension`, first argument can't be `null`

packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
```diff
@@ -1464,18 +1464,63 @@ export class BaseQuery {
...
   overTimeSeriesQuery(baseQueryFn, cumulativeMeasure, fromRollup) {
     const dateJoinCondition = cumulativeMeasure.dateJoinCondition();
+    const uniqDateJoinCondition = R.uniqBy(djc => djc[0].dimension, dateJoinCondition);
     const cumulativeMeasures = [cumulativeMeasure];
     if (!this.timeDimensions.find(d => d.granularity)) {
-      const filters = this.segments.concat(this.filters).concat(this.dateFromStartToEndConditionSql(dateJoinCondition, fromRollup, false));
+      const filters = this.segments
+        .concat(this.filters)
+        .concat(this.dateFromStartToEndConditionSql(
+          // If the same time dimension is passed more than once, no need to build the same
```
mcheshkov, line 1473: :+1: for comments like this

packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
```diff
@@ -1464,18 +1464,63 @@ export class BaseQuery {
...
     }
-    const dateSeriesSql = this.timeDimensions.map(d => this.dateSeriesSql(d)).join(', ');
-    const filters = this.segments.concat(this.filters).concat(this.dateFromStartToEndConditionSql(dateJoinCondition, fromRollup, true));
+
+    if (this.timeDimensions.filter(d => !d.dateRange && d.granularity).length > 0) {
+      throw new UserError('Time series queries without dateRange aren\'t supported');
+    }
+
+    // We can't do meaningful query if few time dimensions with different ranges passed,
+    // it won't be possible to join them together without loosing some rows.
```
mcheshkov, line 1488: Typo: `losing`

packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
```diff
@@ -1464,18 +1464,63 @@ export class BaseQuery {
...
+    const rangedTimeDimensions = this.timeDimensions.filter(d => d.dateRange && d.granularity);
+    const uniqTimeDimensionWithRanges = R.uniqBy(d => d.dateRange, rangedTimeDimensions);
```
mcheshkov, line 1490: `ramda` regularly find a new way to confuse me.
Just a note for myself: `d.dateRange` is a pair of strings, represented as array, so naive `Set` or `===` would not work here. `ramda` works just because it has this whole layer of compatibility starting from here, that results in `R.equals` call.
https://github.com/ramda/ramda/blob/v0.30.1/source/internal/_Set.js

So, essentially, `uniqBy` with array keys is quadratic, not `n*logn`

packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
```diff
@@ -1464,18 +1464,63 @@ export class BaseQuery {
...
+    if (uniqTimeDimensionWithRanges.length > 1) {
+      throw new Error('Can\'t build query for time dimensions with different date ranges');
+    }
+
+    // We need to generate time series table for the lowest granularity among all time dimensions
+    let dateSeriesDimension;
```
mcheshkov, line 1496: Feels like this should be combined to a single `reduce`

Something like this:

```
    const [dateSeriesDimension, dateSeriesGranularity] = this.timeDimensions.filter(d => d.granularity)
      .reduce(([prevDim, prevGran], d) => {
        const mg = this.minGranularity(prevGran, d.resolvedGranularity());
        if (mg === d.resolvedGranularity()) {
          return [dateSeriesDimension, mg] ;
        }
        return [prevDim, mg];
      }, [null, null]);
```

mcheshkov (approved):
LGTM feature-wise, commented on minor stuff

PR #9133, opened 2025-01-23
chore(deps): Update dependencies

KSDaemon:
- **upgrade url-parse**
- **upgrade semver**
- **upgrade json5**
- **upgrade cross-spawn**
- **update loader-utils**
- **update mssql**
- **update postcss**
- **update nanoid**
- **update engine.io**
- **update rollup**
- **update ua-parser-js**
- **update semver**
- **update webpack**


packages/cubejs-druid-driver/src/DruidClient.ts
```diff
@@ -64,7 +64,7 @@ export class DruidClient {
         });
 
         if (cancelled) {
-          await this.cancel(response.headers['x-druid-sql-query-id']);
+          await this.cancel(response.headers['x-druid-sql-query-id'] || '');
```
ovr, line 67: If there is no `x-druid-sql-query-id,`, we don't need to call the `cancel` method, + we need to log an error.

ovr (approved):

PR #9145, opened 2025-01-27
chore(ci): Fix post-release and cloud db tests ci workflows

KSDaemon:
- **fix post-release ci job**
- **fix cloud tests ci mb?**


mcheshkov (approved):
Let's merge it this way, to unlock releases, and deal with caches later.

PR #9154, opened 2025-01-29
fix(schema-compiler): Use member expression definition as measure key

mcheshkov:
**Check List**

Without this when measure was a member expression m.measure.name can be undefined, and different measures will have same key in collectRootMeasureToHieararchy


KSDaemon (approved):
👍🏻

PR #9170, opened 2025-02-01
chore!: Remove `running_total` measure type

igorlukanin:
BREAKING CHANGE: `running_total` measure type has long been deprecated. Use the `rolling_window` parameter instead.


igorlukanin:
Wow! It looks like supporting `running_total` required just a tiny bit of code.

KSDaemon:
@igorlukanin If I not missed somewhere else )))

PR #9203, opened 2025-02-07
Adding pricing info for AI API and AI Assistant

djayatillake:
**Check List**

Adding docs for Cube Cloud CCU consumption driven by AI API and AI Assistant.

Not sure how to do the links to other docs that the other pricing sections are doing.


igorlukanin:
@djayatillake I think it would be better to add these as new types of resources to the table. From this table, you can link feature names to the relevant sections of the docs. Also, since these new types of resources don't consume CCUs per hour but rather consume CCUs per request, I suggest adding a new column to this table for the latter. You can still add an additional section to explain that BYOLLM has different cost implications. 

Screenshot 2025-02-11 at 14 21 43

igorlukanin:
Also, here's how we usually refer to product tiers throughout the docs:

Screenshot 2025-02-11 at 14 27 43

djayatillake:
> @djayatillake I think it would be better to add these as new types of resources to the table. From this table, you can link feature names to the relevant sections of the docs. Also, since these new types of resources don't consume CCUs per hour but rather consume CCUs per request, I suggest adding a new column to this table for the latter. You can still add an additional section to explain that BYOLLM has different cost implications.
> 
> Screenshot 2025-02-11 at 14 21 43

But it isn't charged per hour, so it won't really fit in the table.

igorlukanin:
@djayatillake Exactly! That is why I suggested to introduce a new column to the table:

> Also, since these new types of resources don't consume CCUs per hour but rather consume CCUs per request, I suggest adding a new column to this table for the latter

Without having these resources in the table, I'm afraid it would be quite hard to discover that we do charge for AI features (too far to scroll).

djayatillake:
> @djayatillake Exactly! That is why I suggested to introduce a new column to the table:
> 
> > Also, since these new types of resources don't consume CCUs per hour but rather consume CCUs per request, I suggest adding a new column to this table for the latter
> 
> Without having these resources in the table, I'm afraid it would be quite hard to discover that we do charge for AI features (too far to scroll).

How does it look now?

PR #9204, opened 2025-02-07
 fix(cubejs-server-core): fix schema caching in SQL API with DAP on

bsod90:
**Check List**

[For example #12]

[Description goes here]

### DIFF
--- a/packages/cubejs-server-core/src/core/CompilerApi.js
+++ b/packages/cubejs-server-core/src/core/CompilerApi.js
@@ -1,7 +1,7 @@
 import crypto from 'crypto';
 import R from 'ramda';
 import { createQuery, compile, queryClass, PreAggregations, QueryFactory } from '@cubejs-backend/schema-compiler';
-import { v4 as uuidv4 } from 'uuid';
+import { v4 as uuidv4, parse as uuidParse, stringify as uuidStringify } from 'uuid';
 import { NativeInstance } from '@cubejs-backend/native';
 
 export class CompilerApi {
@@ -428,7 +428,7 @@ export class CompilerApi {
     const { cubeEvaluator } = compilers;
 
     if (!cubeEvaluator.isRbacEnabled()) {
-      return cubes;
+      return { cubes, visibilityMaskHash: null };
     }
 
     for (const cube of cubes) {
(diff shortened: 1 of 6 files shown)


packages/cubejs-server-core/src/core/CompilerApi.js
```diff
@@ -481,39 +481,59 @@ export class CompilerApi {
...
+    };
+  }
+
+  mixInVisibilityMaskHash(compilerId, visibilityMaskHash) {
+    const uuidBytes = uuidParse(compilerId);
+    const hashBytes = Buffer.from(visibilityMaskHash, 'hex');
+    // Mix hash into the first 4 bytes and the last 4 bytes in a way that doesn't change the UUID version
+    for (let i = 0; i < 4; i++) {
+      // eslint-disable-next-line no-bitwise
+      uuidBytes[i] ^= hashBytes[i]; // First 4 bytes
```
mcheshkov, line 508: Let's not use XOR mixing, and not do this patching manually, please.

1. One can provide 16 random bytes to `uuid.v4` with `random` option. See [link](https://github.com/uuidjs/uuid?tab=readme-ov-file#uuidv4options-buffer-offset)
2. If we want to derive new pseudorandom value deterministically, simplest and safest way would be to just concatenate and hash them again (because we know/expect the length of parts to be fixed): `crypto.createHash('sha256').update(uuidBytes).update(hashBytes).digest()`

So, why not just `uuidv4({random: c.createHash('sha256').update(uuidBytes).update(hashBytes).digest().slice(0,16)})`?

packages/cubejs-server-core/src/core/CompilerApi.js
```diff
@@ -481,39 +481,59 @@ export class CompilerApi {
...
-        config: {
-          ...cube.config,
-          measures: cube.config.measures?.map(visibilityPatcherForCube(cube)),
-          dimensions: cube.config.dimensions?.map(visibilityPatcherForCube(cube)),
-          segments: cube.config.segments?.map(visibilityPatcherForCube(cube)),
-          hierarchies: cube.config.hierarchies?.map(visibilityPatcherForCube(cube)),
-        },
-      }));
+    const visibiliyMask = JSON.stringify(isMemberVisibleInContext, Object.keys(isMemberVisibleInContext).sort());
+    const visibilityMaskHash = crypto.createHash('md5').update(visibiliyMask).digest('hex').slice(0, 8);
```
mcheshkov, line 485: Why slicing just 4 first bytes?

packages/cubejs-server-core/src/core/CompilerApi.js
bsod90: The native side expects `compilerId` to be a valid uuid (it tries to parse it)

packages/cubejs-server-core/src/core/CompilerApi.js
bsod90: to make it shorter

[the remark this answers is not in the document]
bsod90: md5 is already used in other places in the same file. For this use-case crypto strength isn't important so, any hash would do. Can switch it to sha256

packages/cubejs-server-core/src/core/CompilerApi.js
bsod90: Also important is that if DAP isn't applied, `compilerId` stays the same. This is to make sure we don't accidentally introduce any bug in non-DAP use-cases

mcheshkov (commented):
I forgot to click "Submit review" yesterday :disappointed:

packages/cubejs-server-core/src/core/CompilerApi.js
bsod90: Ah, ok. Now I see your point
updated the code

mcheshkov (approved):

PR #9211, opened 2025-02-11
chore(tesseract): Fix issue with FILTER_PARAMS and issue with multi-stage behavior of not multi-stage members

waralexrom:
**Check List**

[For example #12]

[Description goes here]

### DIFF
--- a/packages/cubejs-schema-compiler/test/integration/postgres/sql-generation.test.ts
+++ b/packages/cubejs-schema-compiler/test/integration/postgres/sql-generation.test.ts
@@ -1,3 +1,4 @@
+import { getEnv } from '@cubejs-backend/shared';
 import { UserError } from '../../../src/compiler/UserError';
 import { PostgresQuery } from '../../../src/adapter/PostgresQuery';
 import { BigqueryQuery } from '../../../src/adapter/BigqueryQuery';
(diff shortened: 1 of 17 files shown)


rust/cubenativeutils/src/wrappers/serializer/serializer.rs
```diff
@@ -138,7 +138,7 @@ impl<IT: InnerTypes> ser::Serializer for NativeSerdeSerializer<IT> {
     }
 
     fn serialize_none(self) -> Result<Self::Ok, Self::Error> {
-        Ok(self.context.undefined()?)
+        Ok(self.context.null()?)
```
KSDaemon, line 141: Are you sure `null` is correct? In JS it will result in smth like: `myVarOrProperty: null` instead of `undefined`.

packages/cubejs-schema-compiler/test/integration/postgres/sql-generation.test.ts
```diff
@@ -292,9 +293,12 @@ describe('SQL Generation', () => {
 
     cube('visitor_checkins', {
       sql: \`
-      select * from visitor_checkins WHERE
-      \${FILTER_PARAMS.visitor_checkins.created_at.filter('created_at')} AND
-      \${FILTER_GROUP(FILTER_PARAMS.visitor_checkins.created_at.filter("(created_at - INTERVAL '3 DAY')"), FILTER_PARAMS.visitor_checkins.source.filter('source'))}
+      select visitor_checkins.* from visitor_checkins left join visitors on visitor_checkins.visitor_id = visitors.id WHERE
```
KSDaemon, line 296: Can you explain the change? Could you tell me why you didn't create another cube/test or why this one doesn't suit your needs?

packages/cubejs-schema-compiler/test/integration/postgres/sql-generation.test.ts
```diff
@@ -2975,6 +3057,37 @@ describe('SQL Generation', () => {
     }]
   ));
 
+  if (getEnv('nativeSqlPlanner')) {
```
KSDaemon, line 3060: I think it would be more correct to mark the test as skipped in case of not using `nativeSqlPlanner` instead of just avoiding it all.

rust/cubesqlplanner/cubesqlplanner/src/planner/planners/multi_stage_query_planner.rs
```diff
@@ -248,7 +248,26 @@ impl MultiStageQueryPlanner {
             &time_dimension.get_granularity(),
         )?;
 
-        new_state.change_time_dimension_granularity(&time_dimension_name, result_granularity);
+        if time_dimension.get_date_range().is_some() && result_granularity.is_some() {
+            let granularity = time_dimension.get_granularity().unwrap(); //FIXME remove this unwrap
+            let date_range = time_dimension.get_date_range().unwrap(); //FIXME remove this unwrap
+            let seria = self
```
KSDaemon, line 254:
```suggestion
            let series = self
```

rust/cubesqlplanner/cubesqlplanner/src/planner/planners/multi_stage_query_planner.rs
```diff
@@ -376,7 +395,7 @@ impl MultiStageQueryPlanner {
 
         let childs = member_childs(&member)?;
 
-        let description = if childs.is_empty() {
+        let description = if childs.is_empty() || !has_multi_stage_members(&member, false)? {
             if has_multi_stage_members(&member, false)? {
```
KSDaemon, line 399: is it correct to check the same condition `has_multi_stage_members(&member, false)` twice?

rust/cubesqlplanner/cubesqlplanner/src/planner/sql_evaluator/sql_call.rs
```diff
@@ -217,16 +220,72 @@ impl SqlCall {
...
+        if let Some(filter) = filter {
+            let mut res = Vec::new();
+            for item in filter.items.iter() {
+                res.push(self.filters_to_native_filter_item_impl(item));
+            }
+            Some(res)
+        } else {
+            None
+        }
+    }
```
KSDaemon, line 259: I think this one may be rewritten in more rust-idiomatic style, smth like:
```suggestion
fn filters_to_native_filter_item(
    &self,
    filter: Option<Filter>,
) -> Option<Vec<NativeFilterItem>> {
    filter.map(|filter| 
        filter.items.iter()
            .map(|item| self.filters_to_native_filter_item_impl(item))
            .collect()
    )
}
```

rust/cubesqlplanner/cubesqlplanner/src/planner/sql_evaluator/sql_call.rs
```diff
@@ -217,16 +220,72 @@ impl SqlCall {
...
+
+    fn filters_to_native_filter_item_impl(&self, filter_item: &FilterItem) -> NativeFilterItem {
+        match filter_item {
+            FilterItem::Group(group) => {
+                let mut native_items = Vec::new();
+                for itm in group.items.iter() {
+                    native_items.push(self.filters_to_native_filter_item_impl(itm));
+                }
```
KSDaemon, line 267: Same here, more idiomatic style:
```suggestion
let native_items: Vec<_> = group.items
    .iter()
    .map(|itm| self.filters_to_native_filter_item_impl(itm))
    .collect();
```

KSDaemon (commented):
👍🏻  Left minor notes.

packages/cubejs-schema-compiler/test/integration/postgres/sql-generation.test.ts
waralexrom: This is an fix of an incorrect data model for tests. `SECURITY_CONTEXT.filter` should work regardless of whether the cube has built a joiner for a particular querier or not.  Therefore, this filter must be written directly in the sql of the cube. This is one of the reasons why SECURITY_CONTEXT is deprecated. In the case of tesseract it led to the fact that in one of the tests fails, because tesseract knows how to omit unnecessary joins in subqueries. But this is not the error of tesseract, but of incorrect use of SECURITY_CONTEXT in the model.

rust/cubenativeutils/src/wrappers/serializer/serializer.rs
waralexrom: Yes, I think it's justified here. In general, this change was made to normally pass parameters in filters, which can be either a string or null.

rust/cubesqlplanner/cubesqlplanner/src/planner/planners/multi_stage_query_planner.rs
waralexrom: done

rust/cubesqlplanner/cubesqlplanner/src/planner/planners/multi_stage_query_planner.rs
waralexrom: fixed

rust/cubesqlplanner/cubesqlplanner/src/planner/sql_evaluator/sql_call.rs
waralexrom: This is a very temporary solution that will be removed within a month, so I don't see the point in doing much styling here yet :)

rust/cubesqlplanner/cubesqlplanner/src/planner/sql_evaluator/sql_call.rs
waralexrom: Some as above

KSDaemon (approved):
👍🏻 LGTM!

PR #9225, opened 2025-02-14
feat(schema-compiler): Boost models transpilation 10-13x times (using SWC instead of Babel)

KSDaemon:
This PR brings further improvements into data model compilation.
Cube transpilers are rewritten in Rust and are called asynchronously, moving transpilation jobs to the internal native threads under the hood. More improvements are achieved by caching the transpilation metadata between files transpile calls within one compilation phase. 

This implementation supersedes other approaches taken in https://github.com/cube-js/cube/pull/9217 as it boosts performance more.

There were a bunch of iterations, research, and improvements. But long story short: now js files transpilation is 11−14 times faster than it was before!! 🚀

Here are a few numbers and charts for comparison:

**Input:** 300 js files with ~400 members each

We have 3 phases of compilation. That's why there are 3 messages with times for each phase.

Original master: synchronous sequential traversing in main node.js thread.

```js
Execution Time for transpiling all 301 files: 11657.56 ms
Execution Time for transpiling all 301 files: 11871.40 ms
Execution Time for transpiling all 301 files: 12725.75 ms
```

In total, it's ~36 seconds.
What is more important — we observe event loop delays up to 50,000 ms!

comp1 — Original master

And here is what we have now: asynchronous parallel processing in native worker threads. (defaults to 4 libuv threads)

```js
Execution Time for transpiling all 301 files: 222.43 ms
Execution Time for transpiling all 301 files: 190.51 ms
Execution Time for transpiling all 301 files: 2357.43 ms
```

In total, it's ~2,8 seconds.

Comparing to master: **36 / 2.8 = 12.8x!!!**

What's also important is that I can't get rid of event loop delays totally, but now there are smaller portions of delays, and they are way shorter (5,000 VS 50,000).

comp7 — SWC async backend-native threads-2

**Check List**


waralexrom (approved):

rust/cubetranspilers/src/error_reporter.rs
```diff
@@ -0,0 +1,45 @@
+use std::sync::{Arc, Mutex};
+
+use swc_core::common::errors::{DiagnosticBuilder, Emitter};
+
+pub struct ErrorReporter {
+    pub errors: Arc<Mutex<Vec<String>>>,
```
ovr, line 6: Do we need this `Arc<Mutex>`? I assume you can borrow `&mut` over `self`.  Why? Because you don't run transpilers in parallel. It goes sequentially.

Then you can do whatever you want. `std::mem::replace` to reset the state after transpilation and return errors or `errors: Option<Vec<String>>` and then `self.errors.take()`. WDYT?

Another question: what do you need, Arc? 

`ErrorReporter` is created only once and goes into `Handler`.

packages/cubejs-schema-compiler/src/compiler/ErrorReporter.ts
```diff
@@ -141,16 +141,16 @@ export class ErrorReporter {
     return this.rootReporter().errors;
   }
 
-  public addErrors(errors: CompilerErrorInterface[]) {
-    this.rootReporter().errors.push(...errors);
+  public addErrors(errors: any[]) {
```
ovr, line 144: `any` is not an option for a new code. Next engineer will reverse engineer code just to know the type. No validation.

rust/cubetranspilers/tests/cube_prop_ctx_test.rs
```diff
@@ -0,0 +1,947 @@
...
+        CONTEXT_SYMBOLS.clone(),
+        None,
+        &handler,
+    );
+
+    program.visit_mut_with(&mut visitor);
+
+    let output_code = generate_code(&program, &cm);
+
+    assert!(
```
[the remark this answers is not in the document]
KSDaemon: Resolved in https://github.com/cube-js/cube/pull/9234

packages/cubejs-schema-compiler/src/compiler/ErrorReporter.ts
KSDaemon: Resolved in https://github.com/cube-js/cube/pull/9234

packages/cubejs-schema-compiler/src/compiler/ErrorReporter.ts
```diff
@@ -141,16 +141,16 @@ export class ErrorReporter {
...
   }
 
   public getWarnings() {
     return this.rootReporter().warnings;
   }
 
-  public addWarnings(warnings: SyntaxErrorInterface[]) {
-    this.rootReporter().warnings.push(...warnings);
+  public addWarnings(warnings: any[]) {
+    warnings.forEach((w: any) => { this.warning(w); });
```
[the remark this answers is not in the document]
KSDaemon: same here

packages/cubejs-backend-native/src/transpilers.rs
```diff
@@ -0,0 +1,91 @@
...
+
+#[derive(Deserialize, Clone, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct TransformRequestConfig {
+    pub file_name: String,
+    pub transpilers: Vec<Transpilers>,
+    pub meta_data: Option<TransformMetaData>,
+}
+
+static METADATA_CACHE: RwLock<Option<TransformMetaData>> = RwLock::new(None);
```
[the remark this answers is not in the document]
KSDaemon: Resolved in https://github.com/cube-js/cube/pull/9234

rust/cubetranspilers/src/error_reporter.rs
KSDaemon: Yes, we need it because SWC requires thread safe types under the hood.

PR #9230, opened 2025-02-17
feat(tesseract): Pre-aggregation planning fallback

waralexrom:
**Check List**

[For example #12]

[Description goes here]

### DIFF
--- a/packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
+++ b/packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
@@ -286,11 +286,9 @@ export class BaseQuery {
       return dimension;
     }).filter(R.identity).map(this.newTimeDimension.bind(this));
     this.allFilters = this.timeDimensions.concat(this.segments).concat(this.filters);
+    this.useNativeSqlPlanner = this.options.useNativeSqlPlanner ?? getEnv('nativeSqlPlanner');
+    this.prebuildJoin();
 
-    if (!getEnv('nativeSqlPlanner')) {
-      // Tesseract doesn't require join to be prebuilt and there's a case where single join can't be built for multi-fact query
-      this.join = this.joinGraph.buildJoin(this.allJoinHints);
-    }
     this.cubeAliasPrefix = this.options.cubeAliasPrefix;
     this.preAggregationsSchemaOption = this.options.preAggregationsSchema ?? DEFAULT_PREAGGREGATIONS_SCHEMA;
     this.externalQueryClass = this.options.externalQueryClass;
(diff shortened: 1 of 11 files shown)


packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
```diff
@@ -278,11 +278,9 @@ export class BaseQuery {
       return dimension;
     }).filter(R.identity).map(this.newTimeDimension.bind(this));
     this.allFilters = this.timeDimensions.concat(this.segments).concat(this.filters);
+    this.useNativeSqlPlanner = this.options.useNativeSqlPlanner ?? getEnv('nativeSqlPlanner');
```
KSDaemon, line 281: It would be good to add `.default('false')` to this env so the useNativeSqlPlanner can be always bool.

packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
```diff
@@ -603,33 +609,61 @@ export class BaseQuery {
     return false;
   }
 
+  newQueryNotUseNative() {
```
KSDaemon, line 612:
```suggestion
  newQueryWithoutNative() {
```

packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
```diff
@@ -603,33 +609,61 @@ export class BaseQuery {
...
+          isRelatedToPreAggregation = true;
+        }
+      } else {
+        let preAggForQuery =
+          this.preAggregations.findPreAggregationForQuery();
+        if (this.options.disableExternalPreAggregations && preAggForQuery && preAggForQuery.preAggregation.external) {
+          preAggForQuery = undefined;
+        }
+        if (preAggForQuery) {
+          isRelatedToPreAggregation = true;
```
KSDaemon, line 644: It looks like these 2 ifs can be combined into one. Smth like:
```js
if (!(this.options.disableExternalPreAggregations && 
preAggForQuery && 
preAggForQuery.preAggregation.external)) {
  isRelatedToPreAggregation = true;
```

packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
```diff
@@ -2960,6 +2994,7 @@ export class BaseQuery {
   }
 
   newSubQueryForCube(cube, options) {
+    options = { ...options, useNativeSqlPlanner: false }; // We not use tesseract for pre-aggregations generation yet
```
KSDaemon, line 2997:
```suggestion
    options = { ...options, useNativeSqlPlanner: false }; // We don't use tesseract for pre-aggregations generation yet
```

packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
waralexrom: done

packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
waralexrom: done

packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
waralexrom: The current version seems more readable to me. In any case, this is a very temporary solution

packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
waralexrom: done

KSDaemon (approved):
👍🏻

PR #9242, opened 2025-02-19
fix(backend-native): Handle closed channel on Rust side

mcheshkov:
**Check List**

DataFusion can execute complex queries, containing multiple CubeScans, specifically multiple `CubeScanExecutionPlan` in physical plan
`CubeScanExecutionPlan`s can be executed concurrently, and as such concurrently trigger load calls on `TransportService`
Then, if one of several concurrent calls receives error whole stream with still-executing-plans can be dropped
For `NodeBridgeTransport` this would lead to dropping receiver side in `call_raw_js_with_channel_as_callback`
But sender side would continue to live in JS thread
Then JS side can call `writerOrChannel.resolve`, and it would find that channel is closed, and raise exception
But then this exception will be caught, leading to `writerOrChannel.reject` call
And because `writerOrChannel.resolve` already consumed channel it would raise its own exception
And, because `wrapNativeFunctionWithStream` uses async function, this exception would lead to promise rejection, will not get caught in native part on JS side (it does not care for return values), and can trigger unhandled rejection

### DIFF
--- a/packages/cubejs-backend-native/js/index.ts
+++ b/packages/cubejs-backend-native/js/index.ts
@@ -292,7 +292,23 @@ function wrapNativeFunctionWithStream(
       if (!!response && !!response.stream) {
         response.stream.destroy(e);
       }
-      writerOrChannel.reject(errorString(e));
+
+      try {
+        writerOrChannel.reject(errorString(e));
+      } catch (rejectError) {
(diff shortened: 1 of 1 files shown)


ovr (approved):

KSDaemon (approved):
👍🏻

PR #9270, opened 2025-02-25
fix(server-core): Handle empty query in getSqlGenerator

mcheshkov:
**Check List**

api-gateway calls getSqlGenerator with empty query and concrete data source to initialize SQL API But because query is empty `sqlGenerator.dataSource` can be undefined, and it would trigger re-creating query with new data source, which would be `default`


KSDaemon (approved):
👍🏻 LGTM!

ovr (approved):

PR #9289, opened 2025-02-28
feat(server-core): add fastReloadEnabled option to CompilerApi

asmisha:
**Check List**

[For example #12]

[Description goes here]

### DIFF
--- a/packages/cubejs-server-core/src/core/server.ts
+++ b/packages/cubejs-server-core/src/core/server.ts
@@ -712,6 +712,7 @@ export class CubejsServerCore {
         sqlCache: this.options.sqlCache,
         standalone: this.standalone,
         allowNodeRequire: options.allowNodeRequire,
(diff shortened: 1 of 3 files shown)


packages/cubejs-server-core/src/core/server.ts
```diff
@@ -712,6 +712,7 @@ export class CubejsServerCore {
         sqlCache: this.options.sqlCache,
         standalone: this.standalone,
         allowNodeRequire: options.allowNodeRequire,
+        fastReloadEnabled: getEnv('fastReloadEnabled'),
```
KSDaemon, line 715: Don't you want to do it like:
```suggestion
        fastReloadEnabled: options.fastReloadEnabled || getEnv('fastReloadEnabled'),
```
In case option from outside is set?

KSDaemon (approved):
👍🏻 LGTM!

PR #9294, opened 2025-03-03
chore(fast-reload): support fastReloadEnabled as a server option

asmisha:
**Check List**

[For example #12]

[Description goes here]

### DIFF
--- a/packages/cubejs-server-core/src/core/types.ts
+++ b/packages/cubejs-server-core/src/core/types.ts
@@ -227,7 +227,7 @@ export interface CreateOptions {
   serverless?: boolean;
   allowNodeRequire?: boolean;
   semanticLayerSync?: (context: RequestContext) => Promise<BiToolSyncConfig[]> | BiToolSyncConfig[];
-  fastReloadEnabled?: boolean;
+  fastReload?: boolean;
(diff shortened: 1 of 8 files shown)


packages/cubejs-server-core/src/core/types.ts
```diff
@@ -227,6 +227,7 @@ export interface CreateOptions {
   serverless?: boolean;
   allowNodeRequire?: boolean;
   semanticLayerSync?: (context: RequestContext) => Promise<BiToolSyncConfig[]> | BiToolSyncConfig[];
+  fastReloadEnabled?: boolean;
```
ovr, line 230: We don't use any prefixes/suffixes for `boolean` options. I belive it should `fastReload`

ovr:
@asmisha, you must define this option for Python configuration. Otherwise, it will not work from `cube.py`

https://github.com/cube-js/cube/blob/v1.2.14/packages/cubejs-backend-native/python/cube/src/__init__.py
https://github.com/cube-js/cube/blob/v1.2.14/packages/cubejs-backend-native/src/python/cube_config.rs#L18

ovr (approved):

PR #9301, opened 2025-03-04
feat: Add SQL queries support in /v1/sql endpoint

mcheshkov:
**Check List**

`/v1/sql` can now generate SQL for SQL API queries. It errors out for meta-only queries (like `CREATE TEMPORARY TABLE`) and post-processing queries (like `SELECT version();`), and works only when logical plan root is either `CubeScan` or `CubeScanWrappedSql`


packages/cubejs-api-gateway/src/gateway.ts
```diff
@@ -1272,6 +1295,25 @@ class ApiGateway {
     return [queryType, normalizedQueries, queryNormalizationResult.map((it) => remapToQueryAdapterFormat(it.normalizedQuery))];
   }
 
+  public async sql4sql({
```
KSDaemon, line 1298: Should this be private?

packages/cubejs-backend-native/src/node_export.rs
```diff
@@ -425,6 +631,60 @@ fn exec_sql(mut cx: FunctionContext) -> JsResult<JsValue> {
     Ok(promise.upcast::<JsValue>())
 }
 
+fn sql4sql(mut cx: FunctionContext) -> JsResult<JsValue> {
```
KSDaemon, line 634: Don't you mind moving all these functions to sub module (like it was done for orchestrator/planner/python)? - To keep this file clean and hold only kind of entry points, register interfaces, while having all the handlers in a separate sub modules.

packages/cubejs-api-gateway/src/gateway.ts
mcheshkov: TBH, I'm not sure.
I just copy-pasted `public async sql({`, but I see no reason to keep this public

MazterQyou (approved):

packages/cubejs-testing/test/__snapshots__/smoke-cubesql.test.ts.snap
```diff
@@ -1,5 +1,338 @@
...
+  UNION ALL
+  select 3 as id, 300 as amount, 'processed' status, '2024-01-03'::timestamptz created_at
+  UNION ALL
+  select 4 as id, 500 as amount, 'processed' status, '2024-01-04'::timestamptz created_at
+  UNION ALL
+  select 5 as id, 600 as amount, 'shipped' status, '2024-01-05'::timestamptz created_at
+  ) AS \\"orders\\"  WHERE (\\"orders\\".status = $1)",
+      "values": Array [
+        "foo",
+      ],
```
igorlukanin, line 153: @mcheshkov I think we're breaking the existing contract here. I thought that `sql.sql` will be a two element array, with first element being a string, second being an array of parameters.

 Here's what `/sql` currently returns:
```json
{
  "sql": {
    "sql": [
      "SELECT\n      `products`.product_category `orders__product_category`, date_trunc('year', from_utc_timestamp(`base_orders`.created_at::TIMESTAMP, 'UTC')) `orders__created_at_year`\n    FROM\n      hive_metastore.default.orders AS `base_orders`\nLEFT JOIN hive_metastore.default.line_items AS `line_items` ON `base_orders`.id = `line_items`.order_id\nLEFT JOIN hive_metastore.default.products AS `products` ON `line_items`.product_id = `products`.id  WHERE (`base_orders`.created_at::TIMESTAMP >= from_utc_timestamp(replace(replace(?, 'T', ' '), 'Z', ''), 'UTC') AND `base_orders`.created_at::TIMESTAMP <= from_utc_timestamp(replace(replace(?, 'T', ' '), 'Z', ''), 'UTC')) GROUP BY 1, 2 ORDER BY 2 ASC LIMIT 10000",
      [
        "2025-03-06T00:00:00.000Z",
        "2025-03-06T23:59:59.999Z"
      ]
    ],
// skipped
```

igorlukanin: The current way is definitely ugly but I bet we should not break it for no good reason.

PR #9309, opened 2025-03-06
feat(server): Add server config for headers and keep alive timeouts

marianore-muttdata:
**Check List**

This is a change that we've been using in our application to sync the timeouts with the LoadBalancer correctly.

Added two new environment variables to configure HTTP server options:
- `CUBEJS_SERVER_HEADERS_TIMEOUT`
- `CUBEJS_SERVER_KEEP_ALIVE_TIMEOUT`

Each default to the default values in the `Server` class found in the `HTTP` Node module.


KSDaemon:
Hey @marianore-muttdata looks good. Could you please rebase it on top of the latest master branch? GitHub shows that there are some conflicts. Thnx!

packages/cubejs-backend-shared/src/env.ts
```diff
@@ -144,6 +144,12 @@ const variables: Record<string, (...args: any) => any> = {
   webSockets: () => get('CUBEJS_WEB_SOCKETS')
     .default('false')
     .asBoolStrict(),
+  serverHeadersTimeout: () => get('CUBEJS_SERVER_HEADERS_TIMEOUT')
+    .default('60000')
```
KSDaemon, line 148: I think, it may be better to not provide these default values and instead rely on the ones from the node.js itself. So the value from your runtime node versions will be used. Because in other case - these values will stay forever because everyone forgot about it, but they might be changed in the next node versions.

KSDaemon: So if envs are not set - just do not touch `this.server.headersTimeout`.
And same for other env.

KSDaemon (changes requested):

marianore-muttdata:
Moved to https://github.com/cube-js/cube/pull/9684

packages/cubejs-backend-shared/src/env.ts
marianore-muttdata: Makes total sense. Thanks for the comment.
I'll updated the code, let me know if it's ok.
Also, some tests are failing but don't seem to be related to this addition or to the modules implicated in this PR.

KSDaemon: Cool! Thnx! I'll have a look!

marianore-muttdata:
> Hey @marianore-muttdata! Actually... The tests are failing because of:
> 
> ```
>     Error: Invalid cube-server-core options: "serverHeadersTimeout" is not 
>       allowed. "serverKeepAliveTimeout" is not allowed
> ```
> 
> We forgot to update the options scheme [here](https://github.com/cube-js/cube/blob/9ea0b9c606297a05b22f3156c30c73204bf89ab7/packages/cubejs-server-core/src/core/optionsValidate.ts#L48)

Got it! Let me know if there's anything else that you'd write differently.

PR #9331, opened 2025-03-11
feat: Implement disable_post_processing in /v1/sql

mcheshkov:
**Check List**

Second part of  #9301

### DIFF
--- a/packages/cubejs-api-gateway/src/gateway.ts
+++ b/packages/cubejs-api-gateway/src/gateway.ts
@@ -330,6 +330,7 @@ class ApiGateway {
       if (req.query.format === 'sql') {
         await this.sql4sql({
           query: req.query.query,
+          disablePostProcessing: req.query.disable_post_processing === 'true',
           context: req.context,
           res: this.resToResultFn(res)
         });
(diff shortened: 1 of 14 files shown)


MazterQyou (approved):

packages/cubejs-api-gateway/src/gateway.ts
```diff
@@ -330,6 +330,7 @@ class ApiGateway {
       if (req.query.format === 'sql') {
         await this.sql4sql({
           query: req.query.query,
+          disablePostProcessing: req.query.disable_post_processing === 'true',
```
KSDaemon, line 333: I see that we have `disable_post_processing` in snake-case defined in the docs... But in other places we have CamelCase.... Maybe we can review and align it?

KSDaemon (approved):
👍🏻 LGTM!

packages/cubejs-api-gateway/src/gateway.ts
ovr: +1 for camelCase. 

I assume we need to follow the style we used previously, and it should be camelCase.

cC @igorlukanin

ovr (approved):

PR #9335, opened 2025-03-12
fix(schema-compiler): Handle measures with dimension-only member expressions

mcheshkov:
**Check List**

`fullKeyQueryAggregate` is using measure.cube() a lot, like for grouping measures to key cube
This can be wrong for measure targeting view: we want to attach measure to subquery
for its definition cube, and there should be no subquery for view itself
It's expected that `fullKeyQueryAggregateMeasures` would resolve and prepare all measures from its original form in query to actual leaf measures in cubes, then build subqueries with those, and then `joinFullKeyQueryAggregate` would use `renderedReference` to point to these leaf measures

Hard case is a measure that:
* targeting view
* is a member expression
* references only dimensions from that view
Measure like that are "leaf" - there's nowhere to push it, member expression has to be directly in leaf subquery
If measure references dimension from a single cube it can be multiplied (think `SUM(${view.deep_dimension})`)
So, three points:
* it must be accounted for in `multipliedMeasures`
* it must be attached to a proper cube subquery
* it must use renderedReference correctly

`collectRootMeasureToHieararchy` will not drop such measure completely. Now it will check is there's 0 measures collected, try to collect all referenced members to gather used cubes and detect multiplication, and add new measure in hierarchy.

Because new returned measure is patched, it will be attached to correct cube subquery in `fullKeyQueryAggregate`

Then `outerMeasuresJoinFullKeyQueryAggregate` needs to generate proper alias for it, so outer unpatched measure would pick up alias from inner patched one

Finally, `checkShouldBuildJoinForMeasureSelect` has to be adjusted: it collected cube names and join hints, and for dimensions-only measures on views it would collect too many, and decide that measure is multiplied, when it should not. So now it first collects non-view members, and then collect cubes and join hints from those.

### DIFF
--- a/packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
+++ b/packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
@@ -1697,13 +1697,60 @@ export class BaseQuery {
       ${this.query()}`;
   }
 
+  dimensionOnlyMeasureToHierarchy(context, m) {
+    const measureName = typeof m.measure === 'string' ? m.measure : `${m.measure.cubeName}.${m.measure.name}`;
+    const memberNamesForMeasure = this.collectFrom(
+      [m],
+      this.collectMemberNamesFor.bind(this),
+      context ? ['collectMemberNamesFor', JSON.stringify(context)] : 'collectMemberNamesFor',
+      this.queryCache
+    );
+    const cubeNamesForMeasure = R.pipe(
+      R.map(member => this.memberInstanceByPath(member)),
+      // collectMemberNamesFor can return both view.dim and cube.dim
+      R.filter(member => member.definition().ownedByCube),
+      R.map(member => member.cube().name),
+      // Single member expression can reference multiple dimensions from same cube
+      R.uniq,
+    )(
+      memberNamesForMeasure
+    );
+
+    let cubeNameToAttach;
+    switch (cubeNamesForMeasure.length) {
+      case 0:
+        // For zero reference measure there's nothing to derive info about measure from
+        // So it assume that it's a regular measure, and it will be evaluated on top of join tree
+        return [measureName, [{
+          multiplied: false,
+          measure: m.measure,
+        }]];
+      case 1:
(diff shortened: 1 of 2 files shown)


packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
```diff
@@ -1709,7 +1716,64 @@ export class BaseQuery {
         const cubeName = m.expressionCubeName ? `\`${m.expressionCubeName}\` ` : '';
         throw new UserError(`The query contains \`COUNT(*)\` expression but cube/view ${cubeName}is missing \`count\` measure`);
       }
-      return [typeof m.measure === 'string' ? m.measure : `${m.measure.cubeName}.${m.measure.name}`, collectedMeasures];
+      if (collectedMeasures.length === 0 && m.isMemberExpression) {
```
waralexrom, line 1719: Maybe we should put this part related to the member expression over dimension into a separate function and just call it here? There is a lot of code inside the lambda function - it's quite hard to read.

packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
```diff
@@ -1976,14 +2040,37 @@ export class BaseQuery {
...
+        'collectMemberNamesFor',
+      );
+
+      const nonViewMembers = memberNamesForMeasure
+        .filter(mem => {
+          const cubeName = this.cubeEvaluator.parsePathAnyType(mem)[0];
+          const cubeDef = this.cubeEvaluator.getCubeDefinition(cubeName);
+          return !cubeDef.isView;
+        })
+        .map(m => this.memberInstanceByPath(m));
```
waralexrom, line 2064: Maybe it would be better to use `ownedByCube` here?
Something like
```
const nonViewMembers = memberNamesForMeasure
         .map(m => this.memberInstanceByPath(m))
        .filter(mem => {
          
          return mem.ownedByCube;
        });     
```

waralexrom (approved):

Issue #9340, opened 2025-03-13
Allow compression for ClickHouse client


Graphmaxer:
**Is your feature request related to a problem? Please describe.**
When querying large datasets with a big LIMIT (like 50 000 for example), the time to process the query inside Cube is very slow compared to the same query directly run via ClickHouse CLI in SQL.

Simple SELECT with 50 000 rows on Cube (3.8s):
![Image](https://github.com/user-attachments/assets/7e29598a-a7fe-4274-b1f3-14a1e56a08ee)

Same query directly on ClickHouse (instant):
![Image](https://github.com/user-attachments/assets/737c5364-c6c8-490b-a52e-9456f6284941)

**Describe the solution you'd like**
The compression option on ClickHouse JS client (https://clickhouse.com/docs/integrations/javascript#compression) can help with large data (https://clickhouse.com/docs/integrations/javascript#tips-for-performance-optimizations).

With compression enabled (1.1s):
![Image](https://github.com/user-attachments/assets/0c3ad788-c733-4037-a1ab-086449fdeeac)

**Describe alternatives you've considered**
I have tried to add more CPU/RAM to the Cube API service, but the issue is still present.

**Additional context**
Compression cannot be enabled with read only https://clickhouse.com/docs/integrations/javascript#read-only-users

Graphmaxer:
We also got really nice improvements directly in our clickhouse cluster 🥳 Maybe we should set this option to `true` by default 

![Image](https://github.com/user-attachments/assets/87db3885-4b3a-493f-9545-70548e228dfc)

PR #9341, opened 2025-03-13
feat(clickhouse-driver): allow to enable compression

Graphmaxer:
**Check List**

[**Issue Reference this PR resolves**](https://github.com/cube-js/cube/issues/9340)

Allow to enable compression via env variable for clickhouse driver

### DIFF
--- a/docs/pages/reference/configuration/environment-variables.mdx
+++ b/docs/pages/reference/configuration/environment-variables.mdx
@@ -217,6 +217,14 @@ Whether the ClickHouse user has read-only access or not.
 | --------------- | ---------------------- | --------------------- |
 | `true`, `false` | N/A                    | N/A                   |
 
+## `CUBEJS_DB_CLICKHOUSE_COMPRESSION`
+
+Whether the ClickHouse client has compression enabled or not.
+
+| Possible Values | Default in Development | Default in Production |
+| --------------- | ---------------------- | --------------------- |
+| `true`, `false` | N/A                    | N/A                   |
+
 ## `CUBEJS_DB_DATABRICKS_ACCEPT_POLICY`
 
 To accept the license terms for the Databricks JDBC driver, this must be set to
--- a/docs/pages/reference/configuration/environment-variables.mdx
+++ b/docs/pages/reference/configuration/environment-variables.mdx
@@ -223,7 +223,7 @@ Whether the ClickHouse client has compression enabled or not.
 
 | Possible Values | Default in Development | Default in Production |
 | --------------- | ---------------------- | --------------------- |
-| `true`, `false` | N/A                    | N/A                   |
+| `true`, `false` | `false`                | `false`               |
 
 ## `CUBEJS_DB_DATABRICKS_ACCEPT_POLICY`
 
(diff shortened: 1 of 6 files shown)


packages/cubejs-clickhouse-driver/src/ClickHouseDriver.ts
```diff
@@ -165,6 +166,11 @@ export class ClickHouseDriver extends BaseDriver implements DriverInterface {
       exportBucket: this.getExportBucket(dataSource),
       readOnly: !!config.readOnly,
       requestTimeout,
+      compression: {
+        // Response compression can't be enabled for a user with readonly=1, as ClickHouse will not allow settings modifications for such user.
+        response: this.readOnlyMode ? false : getEnv('clickhouseCompression', { dataSource }) === 'true',
+        request: getEnv('clickhouseCompression', { dataSource }) === 'true',
```
ovr, line 172: After usage of `env-var` we can remove `=== 'true'` here.

docs/pages/reference/configuration/environment-variables.mdx
```diff
@@ -217,6 +217,14 @@ Whether the ClickHouse user has read-only access or not.
 | --------------- | ---------------------- | --------------------- |
 | `true`, `false` | N/A                    | N/A                   |
 
+## `CUBEJS_DB_CLICKHOUSE_COMPRESSION`
+
+Whether the ClickHouse client has compression enabled or not.
+
+| Possible Values | Default in Development | Default in Production |
+| --------------- | ---------------------- | --------------------- |
+| `true`, `false` | N/A                    | N/A                   |
```
igorlukanin, line 226: Why `N/A`? I see that it's `false` by default, right?

igorlukanin:
Before we merge this—

@Graphmaxer Did you confirm that enabling the compression actually improves the performance for large result sets?

igorlukanin:
Also, if this is a non-breaking change (except for the nuance with read-only connections), should we consider enabling the compression by default in the next minor version of Cube?

docs/pages/reference/configuration/environment-variables.mdx
Graphmaxer: You are right, I updated the docs

Graphmaxer:
I did some testing documented in my issue https://github.com/cube-js/cube/issues/9340, I got positive impact in my use case. According to the Clickhouse documentation, this can have negative impacts depending on the queries.

> The client does not enable request or response compression by default. However, when selecting or inserting large datasets, you could consider enabling it via ClickHouseClientConfigOptions.compression (either for just request or response, or both).
Compression has significant performance penalty. Enabling it for request or response will negatively impact the speed of selects or inserts, respectively, but will reduce the amount of network traffic transferred by the application.

https://clickhouse.com/docs/integrations/javascript#tips-for-performance-optimizations

maximethebault:
Also facing the same issue when using PowerBI... Queries time out, but when replaying them in Clickhouse directly, they are almost instantaneous.

Graphmaxer:
Up :cry:

ovr:
Thank you @Graphmaxer for your contribution 👍 🍰 

Merged, it will be released in the next release.

PR #9365, opened 2025-03-19
fix(databricks-jdbc-driver): Allow paths in s3/azure export buckets

KSDaemon:
This allows to set path in the export bucket configuration for the databricks driver.

**Check List**


packages/cubejs-databricks-jdbc-driver/src/DatabricksDriver.ts
```diff
@@ -703,6 +703,13 @@ export class DatabricksDriver extends JDBCDriver {
     // wasbs://real-container-name@account.blob.core.windows.net
     // The extractors in BaseDriver expect just clean bucket name
     const url = new URL(this.config.exportBucket || '');
+    const prefix = url.pathname.slice(1);
+    let delimiter = '';
+    if (prefix && !prefix.endsWith('/')) {
```
ovr, line 708: I prefer to write such ifs using a ternary operator because it allows me to mark variables with a `const` modifier.

ovr (approved):
LGTM, minor nit.

packages/cubejs-databricks-jdbc-driver/src/DatabricksDriver.ts
KSDaemon: Funny, that I fix it to `obj?.smth` everywhere I faced it :) But missed in my own code)

PR #9366, opened 2025-03-19
fix(cubejs-schema-compiler): support unary operators in DAP filters

bsod90:
**Check List**

[For example #12]

[Description goes here]


packages/cubejs-schema-compiler/src/compiler/CubeValidator.ts
```diff
@@ -627,18 +627,24 @@ const PolicyFilterSchema = Joi.object().keys({
...
     'notInDateRange',
+    'onTheDate',
     'beforeDate',
     'beforeOrOnDate',
     'afterDate',
     'afterOrOnDate',
+    'measureFilter',
   ).required(),
-  values: Joi.func().required(),
+  values: Joi.func(),
```
KSDaemon, line 647: It's not great to make this optional, while most of the operators require values. So I suggest making an additional logic to mark values as optional only for set/notSet. Smth like:
```suggestion
  values: Joi.when('operator', {
    is: Joi.valid('set', 'notSet'),
    then: Joi.func().optional(),
    otherwise: Joi.func().required()
  })
```

bsod90: Oh, I didn't know joi can do this, thx :D

KSDaemon (approved):
👍🏻 LGTM!

PR #9373, opened 2025-03-21
fix(schema-compiler): case insensitive filter for ClickHouse

mironovich:
**Check List**
- [*] Tests have been run in packages where changes made if available
- [* ] Linter has been run for changed code
- [*] Tests for the changes have been added if not covered yet
- [* ] Docs have been added / updated if required

lower() => lowerUTF8()

### DIFF
--- a/packages/cubejs-schema-compiler/src/adapter/ClickHouseQuery.ts
+++ b/packages/cubejs-schema-compiler/src/adapter/ClickHouseQuery.ts
@@ -18,7 +18,7 @@ class ClickHouseFilter extends BaseFilter {
   public likeIgnoreCase(column, not, param, type) {
     const p = (!type || type === 'contains' || type === 'ends') ? '%' : '';
     const s = (!type || type === 'contains' || type === 'starts') ? '%' : '';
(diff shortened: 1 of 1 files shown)


packages/cubejs-schema-compiler/src/adapter/ClickHouseQuery.ts
```diff
@@ -18,7 +18,7 @@ class ClickHouseFilter extends BaseFilter {
   public likeIgnoreCase(column, not, param, type) {
     const p = (!type || type === 'contains' || type === 'ends') ? '%' : '';
     const s = (!type || type === 'contains' || type === 'starts') ? '%' : '';
-    return `lower(${column}) ${not ? 'NOT' : ''} LIKE CONCAT('${p}', lower(${this.allocateParam(param)}), '${s}')`;
+    return `lowerUTF8(${column}) ${not ? 'NOT' : ''} LIKE CONCAT('${p}', lowerUTF8(${this.allocateParam(param)}), '${s}')`;
```
KSDaemon, line 21: I think just using `ILIKE` would be enough here...
```suggestion
    return `${column} ${not ? 'NOT' : ''} ILIKE CONCAT('${p}', ${this.allocateParam(param)}, '${s}')`;
```

PR #9378, opened 2025-03-24
test(cubestore): Fix decimal_partition_pruning test in release mode

srh:
Makes the test wait for or induce compaction first.  In release mode the test had been failing.

Cherry-picked from the DF upgrade branch.

**Check List**


rust/cubestore/cubestore/src/sql/mod.rs
```diff
@@ -2812,21 +2812,34 @@ mod tests {
 
                 println!("All partitions: {:#?}", partitions);
 
-                let plans = service
-                    .plan_query("SELECT sum(num) from foo.numbers where num = 50")
-                    .await
-                    .unwrap();
+                // Semi-busy-wait for, or, seemingly, induce, compaction for 2000 ms.
+                let num_attempts = 100;
+                for i in 0..num_attempts {
```
ovr, line 2817: I believe it's better to use `JobResultListener` and `wait_for_job_results` here or metastore listener.

cC @waralexrom @waralex

srh: It is not clear to me what job to wait for.  Based on logs, some sort of PartitionCompaction job.  But I don't know the right way to see which ones I want to wait on.

PR #9381, opened 2025-03-24
chore(dev-env): Use referenced project for typescript code building

ovr:
https://cube-js.slack.com/archives/C01FU5AP9LJ/p1741252890815729

image

We need to use referenced project (building typescript code from the root), because it resolves dependency graph.

### DIFF
--- a/dev-env.sh
+++ b/dev-env.sh
@@ -18,14 +18,15 @@ install_root_dependencies() {
 build_packages() {
     echo "Building all packages..."
     cd "$SCRIPT_DIR"
+
+    # Building typescript packages first from root (it's referenced project)
+    yarn tsc
+
     for package in packages/*; do
         if [ -d "$package" ]; then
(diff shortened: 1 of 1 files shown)


dev-env.sh
```diff
@@ -18,14 +18,15 @@ install_root_dependencies() {
...
+
     for package in packages/*; do
         if [ -d "$package" ]; then
             echo "Building $package..."
             cd "$package"
-            if ! yarn build 2>/dev/null; then
-                #echo "yarn build failed for $package, trying yarn tsc..."
-                yarn tsc 2>/dev/null || true
-            fi
+            yarn build 2>/dev/null || true
```
ovr, line 29: I believe each typescript package defines a `build` script, which will lead to duplicate builds. But there is not other way.

KSDaemon (approved):
👍🏻 LGTM!

PR #9382, opened 2025-03-24
fix(tesseract): Support rolling window with multiple granularities

waralexrom:
**Check List**

[For example #12]

[Description goes here]


packages/cubejs-schema-compiler/test/integration/postgres/sql-generation.test.ts
```diff
@@ -976,12 +976,12 @@ SELECT 1 AS revenue,  cast('2024-01-01' AS timestamp) as time UNION ALL
     timeDimensions: [
       {
         dimension: 'visitors.created_at',
-        granularity: 'three_days',
```
KSDaemon, line 979: Originally, this test checks for custom granularities in rolling window, but you changed it to predefined one. Not a great deal. You can add another test with 2 predefined granularities, but please don't change existing tests to reduce the test cases.

rust/cubesqlplanner/cubesqlplanner/src/plan/join.rs
```diff
@@ -192,6 +215,7 @@ pub enum JoinCondition {
     BaseJoinCondition(Rc<dyn BaseJoinCondition>),
     RegularRollingWindowJoinCondition(RegularRollingWindowJoinCondition),
     ToDateRollingWindowJoinCondition(ToDateRollingWindowJoinCondition),
+    RollingTotalJoinCondition(RollingTotalJoinCondition),
```
KSDaemon, line 218: Shouldn't all join conditions implement some kind of trait with `to_sql()` func so you can just call it on any condition?

rust/cubesqlplanner/cubesqlplanner/src/plan/join.rs
```diff
@@ -247,6 +278,7 @@ impl JoinCondition {
             JoinCondition::ToDateRollingWindowJoinCondition(cond) => {
                 cond.to_sql(templates, context)
             }
+            JoinCondition::RollingTotalJoinCondition(cond) => cond.to_sql(templates, context),
```
KSDaemon, line 281: If all conditions implement common trait with `to_sql()` - you could just call it here....

rust/cubesqlplanner/cubesqlplanner/src/planner/planners/multi_stage/member_query_planner.rs
```diff
@@ -219,7 +226,29 @@ impl MultiStageMemberQueryPlanner {
...
         let mut select_builder = SelectBuilder::new(from.clone());
-        for dim in dimensions.iter() {
+
+        //We insert render reference for main time dimension (with the some granularity as in time series to avoid unnessesary date_tranc)
+        render_references.insert(
+            rolling_window_desc.time_dimension.full_name(),
+            QualifiedColumnName::new(Some(root_alias.clone()), format!("date_from")),
+        );
+
+        //We also insert render reference for the base dimension of time dimension (i.e. without `_granularit` prefix to let other time dimensions make date_tranc)
```
KSDaemon, line 236:
```suggestion
        //We also insert render reference for the base dimension of time dimension (i.e. without `_granularity` postfix to let other time dimensions make date_tranc)
```

packages/cubejs-schema-compiler/test/integration/postgres/sql-generation.test.ts
waralexrom: I agree it makes sense to bring it back. Although to be fair, that test was actually testing two time_dimensions, not custom granularities. I think the best approach would be to keep my change, and move the old test to custom-granularity. It’s generally better when a test focuses on a single piece of functionality instead of mixing multiple concerns.

KSDaemon (approved):
👍🏻 LGTM! 

Don't forget to add changed test! You promised me! :)

PR #9384, opened 2025-03-24
Update passing-dynamic-parameters-in-a-query.mdx

morgan-at-cube:
FILTER_PARAMS should always be used in this use case to avoid a massive cross join. I also added a simpler code snippet that is easier to adapt to any use case rather than the original verbose version. cc @paco-valdez who I consulted on this approach.

**Check List**

[For example #12]

[Description goes here]

### DIFF
--- a/docs/pages/guides/recipes/data-modeling/passing-dynamic-parameters-in-a-query.mdx
+++ b/docs/pages/guides/recipes/data-modeling/passing-dynamic-parameters-in-a-query.mdx
@@ -86,7 +86,7 @@ cubes:
         sql: id
         type: count
         filters:
-          - sql: "gender = 'female'
+          - sql: "gender = 'female'"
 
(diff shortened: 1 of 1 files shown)


docs/pages/guides/recipes/data-modeling/passing-dynamic-parameters-in-a-query.mdx
```diff
@@ -77,7 +86,7 @@ cubes:
         sql: id
         type: count
         filters:
-          - sql: "gender = 'female' and city = city_filter" 
+          - sql: "gender = 'female'
```
paco-valdez, line 89: it's missing the closing double quote:
```
 - sql: "gender = 'female'"
 ```

morgan-at-cube: Whoops! Thanks for catching!! Added it back

PR #9391, opened 2025-03-26
[WIP add tests] fix(schema-compiler): Use segments as-is in rollupPreAggregation

mcheshkov:
**Check List**

This is to support member expression segments, as BaseFilter does not expect member expressions in it, i.e. in `filterToWhere` is always call `dimensionSql`


KSDaemon (commented):
👍🏻

PR #9397, opened 2025-03-27
feat(server-core, query-orchestrator): Use the fastest xxhash for cache keys and model versions

KSDaemon:
Switch to using faster `xxhash` for cache keys and versions of data models instead of slow `crypto.createHash('md5')`.
Use the fastest node xxhash implementation, written in ... of course **rust** :)

> XXH3: generates 64 or 128-bit hashes, using vectorized arithmetic.

Here are some more info for those who is interested:
* https://xxhash.com
* https://github.com/Cyan4973/xxHash
* https://aras-p.info/blog/2016/08/09/More-Hash-Function-Tests/ - Some comparisons from 2016 (Xxh3 was not born at that moment)
* https://rurban.github.io/smhasher/ - Hash function quality and speed tests

Changes:
- **use xxh3 in server-core package**
- **use xxh3 in query-orchestrator package**
- **use xxh3 in backend-shared package**
- **use xxh3 in cubestore-driver package**
- **add @node-rs/xxhash package**

### DIFF
--- a/packages/cubejs-backend-shared/src/promises.ts
+++ b/packages/cubejs-backend-shared/src/promises.ts
@@ -1,6 +1,5 @@
 /* eslint-disable arrow-body-style,no-restricted-syntax */
-import crypto from 'crypto';
-
+import { xxh3 } from '@node-rs/xxhash';
 import { Optional } from './type-helpers';
 
 export type PromiseLock = {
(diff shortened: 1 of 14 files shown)


packages/cubejs-backend-shared/src/promises.ts
```diff
@@ -273,9 +272,7 @@ export const asyncDebounce = <Ret, Arguments>(
   const cache = new Map<string, Promise<Ret>>();
 
   return async (...args: Arguments[]) => {
-    const key = crypto.createHash('md5')
-      .update(args.map((v) => JSON.stringify(v)).join(','))
-      .digest('hex');
+    const key = xxh3.xxh64(args.map((v) => JSON.stringify(v)).join(',')).toString(16);
```
ovr, line 275: Let's make a function somewhere in `@cubejs-backend-shared,` with a name like `defaultHasherHex16` (or `defaultHasher` and a tiny interface over it) and use it everywhere in the code instead of defining and requiring `@node-rs/xxhash` in each package. 

This approach is how the Rust language does it: `DefaultHasher`. It will allow us to change it on the fly for testing via envs or replace the implementation whenever necessary.

KSDaemon: Absolutely agree! Great point!

PR #9407, opened 2025-03-31
feat(databricks-jdbc-driver): Add export bucket support for Google Cloud Storage

### LINKED ISSUE #9393: Databricks Driver: Export Bucket On GCS
Hi, team
I've tested the `create table using` statement on Databricks-GCP and it succeeds to export csv to a bucket on GCS, so i'm wondering if it's feasible to implement the GCS export feature based on this statement and the extract files function [here](https://github.com/cube-js/cube/blob/master/packages/cubejs-base-driver/src/BaseDriver.ts#L739C19-L739C38)

proof
Image

Image

qiao-x:
**Check List**

Closes #9393

- add `gcs` bucket type

### DIFF
--- a/packages/cubejs-databricks-jdbc-driver/src/DatabricksDriver.ts
+++ b/packages/cubejs-databricks-jdbc-driver/src/DatabricksDriver.ts
@@ -103,6 +103,11 @@ export type DatabricksDriverConfiguration = JDBCDriverConfiguration &
      * Azure service principal client secret
      */
     azureClientSecret?: string,
+
+    /**
+     * GCS credentials JSON content
+     */
+    gcsCredentials?: string,
   };
 
 type ShowTableRow = {
@@ -209,7 +214,7 @@ export class DatabricksDriver extends JDBCDriver {
       // common export bucket config
       bucketType:
         conf?.bucketType ||
-        getEnv('dbExportBucketType', { supported: ['s3', 'azure'], dataSource }),
+        getEnv('dbExportBucketType', { supported: ['s3', 'azure', 'gcs'], dataSource }),
       exportBucket:
         conf?.exportBucket ||
         getEnv('dbExportBucket', { dataSource }),
@@ -246,6 +251,10 @@ export class DatabricksDriver extends JDBCDriver {
       azureClientSecret:
         conf?.azureClientSecret ||
         getEnv('dbExportBucketAzureClientSecret', { dataSource }),
+      // GCS credentials
+      gcsCredentials:
+        conf?.gcsCredentials ||
+        getEnv('dbExportGCSCredentials', { dataSource }),
     };
     if (config.readOnly === undefined) {
       // we can set readonly to true if there is no bucket config provided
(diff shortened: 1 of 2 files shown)


KSDaemon:
@qiao-x I think yes, but let's make it in a separate PR.

packages/cubejs-databricks-jdbc-driver/src/DatabricksDriver.ts
```diff
@@ -733,6 +742,15 @@ export class DatabricksDriver extends JDBCDriver {
         url.host,
         objectSearchPrefix,
       );
+    } else if (this.config.bucketType === 'gcs') {
+      return this.extractFilesFromGCS(
+        { credentials: this.config.gcsCredentials },
+        url.host,
+        objectSearchPrefix+".csv",
```
KSDaemon, line 749: Is it important to add `.csv` here? Are there any other unrelated files that might be captured by `objectSearchPrefix` that should be excluded?

KSDaemon: Having this:
```js
  protected async extractFilesFromGCS(
    gcsConfig: GoogleStorageClientConfig,
    bucketName: string,
    tableName: string
  ): Promise<string[]> {
.......
const [files] = await bucket.getFiles({ prefix: `${tableName}/` });
```
Does Databricks create a folder for a table with a `csv` suffix? Like `my_exported_table.cvs/` with a bunch of real csv files?

packages/cubejs-databricks-jdbc-driver/src/DatabricksDriver.ts
```diff
@@ -733,6 +742,15 @@ export class DatabricksDriver extends JDBCDriver {
...
+      ).then(files => files.filter(file => 
+        decodeURIComponent(new URL(file).pathname).endsWith('.csv') || 
+        decodeURIComponent(new URL(file).pathname).endsWith('.csv.gz')
```
KSDaemon, line 752: The answers to the questions above might affect these lines. Are they needed?

packages/cubejs-databricks-jdbc-driver/src/DatabricksDriver.ts
```diff
@@ -209,7 +214,7 @@ export class DatabricksDriver extends JDBCDriver {
       // common export bucket config
       bucketType:
         conf?.bucketType ||
-        getEnv('dbExportBucketType', { supported: ['s3', 'azure'], dataSource }),
+        getEnv('dbExportBucketType', { supported: ['s3', 'azure', 'gcs'], dataSource }),
```
KSDaemon, line 217: It's better to define a const for this:
```js
const SUPPORTED_BUCKET_TYPES = ['s3', 'gcs', 'azure'];
```
And use it instead of copy-pasting it every time.

KSDaemon:
And I think `convert to quoted tableFullName` was not done in this PR :)

KSDaemon:
@qiao-x I'll pull this PR and update it with CI tests. Let's see how it goes.

PR #9424, opened 2025-04-04
feat(schema-compiler): Use LRUCache for js data models compiled vm.Scripts

KSDaemon:
This introduces an LRU cache for storing compiled versions of JS data model transpiled files. It improves compilation performance.

**Check List**


packages/cubejs-server-core/src/core/CompilerApi.js
```diff
@@ -29,6 +30,11 @@ export class CompilerApi {
     this.sqlCache = options.sqlCache;
     this.standalone = options.standalone;
     this.nativeInstance = this.createNativeInstance();
+    this.compiledScriptCache = new LRUCache({
+      max: options.compilerCacheSize || 250,
+      ttl: options.maxCompilerCacheKeepAlive,
```
ovr, line 35: AFAIR: We don't set up `maxCompilerCacheKeepAlive` by default. Could you check this moment? @KSDaemon

KSDaemon: And that is okay. It is optional. I just added it as we do in other places (for the similarity).

packages/cubejs-server-core/src/core/CompilerApi.js
```diff
@@ -29,6 +30,11 @@ export class CompilerApi {
     this.sqlCache = options.sqlCache;
     this.standalone = options.standalone;
     this.nativeInstance = this.createNativeInstance();
+    this.compiledScriptCache = new LRUCache({
+      max: options.compilerCacheSize || 250,
```
ovr, line 34: I didn't test, but I assume that `vm.Script` takes ~100kb for each "medium" complex model (file). Then 250 is ~25 MB. Is it correct?  But the biggest problem is that `lru-cache` doesn't proactively remove expired items, which can leave a lot of time.

For `CompilerApi` instance cache we are using:

```
    // proactively free up old cache values occasionally
    if (this.options.maxCompilerCacheKeepAlive) {
      this.maxCompilerCacheKeep = setInterval(
        () => this.compilerCache.prune(),
        this.options.maxCompilerCacheKeepAlive
      );
    }
```

I suggest do the same, but we need to enable `ttl` here by default.

KSDaemon: It uses the same options passed from the core/server: 
```
      max: this.options.compilerCacheSize || 250,
      ttl: this.options.maxCompilerCacheKeepAlive,
      updateAgeOnGet: this.options.updateCompilerCacheKeepAlive
```
So we should be fine here.

I can add the `setInterval` for pro-active cleanups. The question is when to clear it....

packages/cubejs-server-core/src/core/server.ts
```diff
@@ -204,7 +204,9 @@ export class CubejsServerCore {
     this.compilerCache = new LRUCache<string, CompilerApi>({
       max: this.options.compilerCacheSize || 250,
       ttl: this.options.maxCompilerCacheKeepAlive,
-      updateAgeOnGet: this.options.updateCompilerCacheKeepAlive
+      updateAgeOnGet: this.options.updateCompilerCacheKeepAlive,
+      // needed to clear the setInterval timer for proactive cache internal cleanups
+      dispose: (v) => v.dispose(),
```
ovr, line 209: 👍 

Then we need to reset `compilerCache` to trigger `dispose` on each `CompilerApi.` 

Right now, we don't clean it during the `shutdown`. As the last step, we should call `this.compilerCache.reset()` in the `shutdown` method.

KSDaemon: yep, sure! will do!

ovr (approved):

PR #9429, opened 2025-04-04
refactor(schema-compiler): More types for CubeEvaluator, CubeSymbols and DynamicReference

mcheshkov:
**Check List**

### DIFF
--- a/packages/cubejs-schema-compiler/src/compiler/CubeEvaluator.ts
+++ b/packages/cubejs-schema-compiler/src/compiler/CubeEvaluator.ts
@@ -1,7 +1,7 @@
 /* eslint-disable no-restricted-syntax */
 import R from 'ramda';
 
-import { CubeSymbols } from './CubeSymbols';
+import { CubeSymbols, type ToString } from './CubeSymbols';
 import { UserError } from './UserError';
 import { BaseQuery } from '../adapter';
 import type { CubeValidator } from './CubeValidator';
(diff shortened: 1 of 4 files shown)


packages/cubejs-schema-compiler/src/compiler/CubeEvaluator.ts
```diff
@@ -510,7 +544,7 @@ export class CubeEvaluator extends CubeSymbols {
           )
         )
       ))
-      .map(cube => {
+      .flatMap(cube => {
```
KSDaemon, line 547: 👍🏻

packages/cubejs-schema-compiler/src/compiler/CubeSymbols.ts
```diff
@@ -292,7 +292,8 @@ export class CubeSymbols {
         // If the hierarchy is included all members from it should be included as well
         // Extend `includes` with members from hierarchies that should be auto-included
         const cubes = type === 'dimensions' ? cube.cubes.map((it) => {
-          const fullPath = this.evaluateReferences(null, it.joinPath, { collectJoinHints: true });
+          // TODO recheck `it.joinPath` typing
+          const fullPath = this.evaluateReferences(null, it.joinPath as () => { toString(): string }, { collectJoinHints: true });
```
KSDaemon, line 296: Maybe it would be better to define an interface for `() => { toString(): string }` instead of copy-pasting many times it `as () => { toString(): string }` ? At least for now. wdyt?

KSDaemon (approved):
👍🏻 LGTM After the copy-paste removal :)

PR #9444, opened 2025-04-09
fix(schema-compiler): Do not drop join hints when view member is not itself owned

mcheshkov:
**Check List**

Without this when following from view to cube member join hints will be collected and ignored, then join hints would be collected for references from cube member to other cube members, and only those would be used for join tree


KSDaemon (approved):
👍🏻 LGTM! Really clear and self-explanatory tests with comments! Thank you!

PR #9447, opened 2025-04-10
docs: TPC-H benchmark and scalability insights

igorlukanin:


igorlukanin:
This has gone stale in a way a rebase can't fix. It adds `docs/pages/guides/benchmarks/`, and the two docs restructures since (#9578, then Nextra v4 in #10414) removed `docs/pages/` entirely — the content now lives under `docs/content/product/`, and `guides` isn't a section any more. Every other `/guides/*` page got an individual redirect into `/product/*` during that move; `/guides/benchmarks` has none, because this never landed.

So the conflict is really an open question about where TPC-H benchmark docs belong in the current IA, which I don't want to answer by picking a directory. Worth deciding alongside CORE-693, which is the live call on how the TPC-H numbers get published — the SF1–SF1000 latency charts here overlap that post directly, and these are 2025 numbers on a pre-Tesseract deployment.

Leaving it open and untouched pending that call.

igorlukanin:
Closing this: the partitioning mechanism it explains is now covered in depth by the Cube Store architecture docs (auto-partitioning, partition pruning, shared-nothing workers), and what's unique here is a set of 2025 latency numbers from a pre-Tesseract deployment, which belong in a post with a version attached rather than a docs page: https://docs.cube.dev/docs/pre-aggregations/cube-store-architecture

PR #9450, opened 2025-04-10
feat(databricks-jdbc-driver): Switch to the latest OSS Databricks JDBC driver

KSDaemon:
Switch to the brand new OSS JDBC Driver proposed by Databricks.
Read more [here](https://docs.databricks.com/aws/en/integrations/jdbc/oss).

**Check List**

### DIFF
--- a/packages/cubejs-backend-shared/src/http-utils.ts
+++ b/packages/cubejs-backend-shared/src/http-utils.ts
@@ -66,11 +66,11 @@ export async function streamWithProgress(
 type DownloadAndExtractFile = {
   showProgress: boolean;
   cwd: string;
-  noExtract?: boolean;
+  skipExtract?: boolean;
   dstFileName?: string;
 };
 
-export async function downloadAndExtractFile(url: string, { cwd, noExtract, dstFileName }: DownloadAndExtractFile) {
+export async function downloadAndExtractFile(url: string, { cwd, skipExtract, dstFileName }: DownloadAndExtractFile) {
   const request = new Request(url, {
     headers: new Headers({
       'Content-Type': 'application/octet-stream',
@@ -101,11 +101,14 @@ export async function downloadAndExtractFile(url: string, { cwd, noExtract, dstF
     });
   });
 
-  if (noExtract) {
+  if (skipExtract) {
     if (dstFileName) {
       fs.copyFileSync(savedFilePath, path.resolve(path.join(cwd, dstFileName)));
     } else {
-      fs.copyFileSync(savedFilePath, cwd);
+      // We still need some name for a file
+      const tmpFileName = path.basename(savedFilePath);
+      const destPath = path.join(cwd, tmpFileName);
+      fs.copyFileSync(savedFilePath, destPath);
     }
   } else {
     await decompress(savedFilePath, cwd);
--- a/packages/cubejs-backend-shared/src/http-utils.ts
+++ b/packages/cubejs-backend-shared/src/http-utils.ts
@@ -66,9 +66,11 @@ export async function streamWithProgress(
 type DownloadAndExtractFile = {
   showProgress: boolean;
   cwd: string;
+  noExtract?: boolean;
+  dstFileName?: string;
 };
 
-export async function downloadAndExtractFile(url: string, { cwd }: DownloadAndExtractFile) {
+export async function downloadAndExtractFile(url: string, { cwd, noExtract, dstFileName }: DownloadAndExtractFile) {
   const request = new Request(url, {
     headers: new Headers({
       'Content-Type': 'application/octet-stream',
@@ -99,7 +101,15 @@ export async function downloadAndExtractFile(url: string, { cwd }: DownloadAndEx
     });
   });
 
-  await decompress(savedFilePath, cwd);
+  if (noExtract) {
+    if (dstFileName) {
+      fs.copyFileSync(savedFilePath, path.resolve(path.join(cwd, dstFileName)));
+    } else {
+      fs.copyFileSync(savedFilePath, cwd);
+    }
+  } else {
+    await decompress(savedFilePath, cwd);
+  }
 
   try {
     fs.unlinkSync(savedFilePath);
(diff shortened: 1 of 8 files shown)


packages/cubejs-databricks-jdbc-driver/src/installer.ts
```diff
@@ -1,38 +1,32 @@
...
-
-  if (acceptStatus === false) {
-    console.log('You declined Terms & Conditions for JDBC driver from DataBricks by CUBEJS_DB_DATABRICKS_ACCEPT_POLICY');
-    console.log('Installation will be skipped');
-  }
-
-  return acceptStatus;
-}
-
+export const OSS_DRIVER_VERSION = '1.0.2';
```
mcheshkov, line 4: There's 1.0.3 already :wink:

ovr (approved):

packages/cubejs-databricks-jdbc-driver/src/helpers.ts
```diff
@@ -15,26 +15,37 @@ async function fileExistsOr(
...
-      path.join(__dirname, '..', 'download', 'DatabricksJDBC42.jar'),
+      path.join(__dirname, '..', 'download', `databricks-jdbc-${OSS_DRIVER_VERSION}-oss.jar`),
       async () => {
         const pathOrNull = await downloadJDBCDriver();
         if (pathOrNull) {
           return pathOrNull;
         }
         throw new Error(
-          'Please download and place DatabricksJDBC42.jar inside your ' +
+          `Please download and place databricks-jdbc-${OSS_DRIVER_VERSION}-oss.jar inside your ` +
```
mcheshkov, line 27: I'm not a software lawyer, but
Can we just put jar inside npm package? Download it during build, and distribute together.
IIUC, it is distributed under Apache license
It does contain some native code for lz4 in `com.databricks.internal.jpountz.util`, but, given that they distribute it as a single jar with all variants inside, I think it's gonna be fine regarding compatibility with target arch/os/...

packages/cubejs-databricks-jdbc-driver/src/helpers.ts
```diff
@@ -15,26 +15,37 @@ async function fileExistsOr(
...
-  const { pathname } = new URL(jdbcUrl);
-  const [_, ...params] = pathname.split(';');
-  const searchParams = new URLSearchParams(params.join('&'));
-  return searchParams.get('UID') || 'token';
+/**
+ * Extract if exist UID and PWD from URL and return UID, PWD and URL without these params
+ * @param jdbcUrl
+ */
+export function extractAndRemoveUidPwdFromJdbcUrl(jdbcUrl: string): [string, string, string] {
+  const uidMatch = jdbcUrl.match(/UID=([^;]*)/);
```
mcheshkov, line 40: Can you add link/reference/something to this drivers specific parsing?
I can't give you a link to GitHub, because my IP is banned in https://github.com/databricks/databricks-jdbc  repo
But there's `buildPropertiesMap` in com/databricks/jdbc/api/impl/DatabricksConnectionContext.java that does all the properties parsing.

mcheshkov, line 40: These regexps should be case-insensitive, at least for key part.
For everything but HTTP headers (which seems like a bug) driver calls `parametersBuilder.put(pair[0].toLowerCase(), pair[1]);` 

It's documented here
https://docs.databricks.com/aws/en/integrations/jdbc/configure-oss

> URL elements and properties are case insensitive

packages/cubejs-databricks-jdbc-driver/src/helpers.ts
```diff
@@ -15,26 +15,37 @@ async function fileExistsOr(
...
-export function extractUidFromJdbcUrl(jdbcUrl: string): string {
-  const { pathname } = new URL(jdbcUrl);
-  const [_, ...params] = pathname.split(';');
-  const searchParams = new URLSearchParams(params.join('&'));
-  return searchParams.get('UID') || 'token';
+/**
+ * Extract if exist UID and PWD from URL and return UID, PWD and URL without these params
+ * @param jdbcUrl
+ */
+export function extractAndRemoveUidPwdFromJdbcUrl(jdbcUrl: string): [string, string, string] {
```
mcheshkov, line 39: Please name positions in result tuple. Like this: `[a: string, b: number]`. This does not introduce actual names to fields, it's the same type, but at least can serve as a comment/documentation.
Or even better: turn it into struct with named fields.

packages/cubejs-databricks-jdbc-driver/src/helpers.ts
KSDaemon: I'm not sure what you want here... Cause this made sense for old proprietary driver, and is not actual for the OSS.
And you're not alone, my IP is also banned.

packages/cubejs-databricks-jdbc-driver/src/helpers.ts
KSDaemon: Yep, okay

packages/cubejs-backend-shared/src/http-utils.ts
```diff
@@ -66,9 +66,11 @@ export async function streamWithProgress(
 type DownloadAndExtractFile = {
   showProgress: boolean;
   cwd: string;
+  noExtract?: boolean;
+  dstFileName?: string;
 };
 
-export async function downloadAndExtractFile(url: string, { cwd }: DownloadAndExtractFile) {
+export async function downloadAndExtractFile(url: string, { cwd, noExtract, dstFileName }: DownloadAndExtractFile) {
```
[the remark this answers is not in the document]
KSDaemon: Agree :) Naming is hard, you know)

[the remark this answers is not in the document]
KSDaemon: It has wider use: not only for artifacts....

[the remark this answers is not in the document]
KSDaemon: So maybe I'll just rename `noExtract` → `skipExtract`... it seems to be better. But I don't want to make breaking changes to function signature

packages/cubejs-databricks-jdbc-driver/src/helpers.ts
KSDaemon: Got it. Well, as we can't paste link... At least I left a comment.

mcheshkov (approved):

PR #9458, opened 2025-04-11
fix(server): Fix incorrect native result wrapper processing via webSockets

KSDaemon:


mcheshkov (approved):
It certainly is better then before
But I hope we can get it to better state son

mcheshkov (approved):

paveltiunov (approved):

PR #9471, opened 2025-04-15
fix(schema-compiler): Use join paths from pre-aggregation declaration instead of building join tree from scratch

mcheshkov:
**Check List**

Take 2 for #9431, now with more tests and fixes in pre-aggregations matching

Without this pre-aggregatoin declaration would build a list of members, and build join tree from scratch. New join tree could be different, and pre-aggregation would contain incorrect data.

Add join hints to member adapters, to build join tree with proper members. This is more of a hack than proper solution: join tree would still disallow to have same cube twice with different join paths.
Just adding extra join hints on query level does not work because of FKQA and multiplication check in subqueries: we have to know join path for every member, so it should be either explicit or unambigous _for each member_.

Join tree from query is still not included in pre-aggregation matching, so user would have to manually ensure that views and pre-aggregations both match each other and does not produce false matches

For now, only pre-aggregation itself respects join paths, but pre-aggregation matching does same as before: strips join path from pre-aggregation references and compares members directly.


waralexrom (approved):

KSDaemon (approved):
👍🏻

PR #9479, opened 2025-04-16
fix(schema-compiler): Make RefreshKeySql timezone-aware

KSDaemon:
This PR makes RefreshKey SQL generation to use the query timezone. This is important for the case when there are a few scheduled refresh timezones (or even one but different from `UTC`) and there is a pre-aggregation with `day` partition granularity. So you want them to be updated at midnight in the provided timezone rather than at `UTC`.
This fixes it.

The change is also made in cron-based refresh keys, making them timezone-aware. If the pre-aggregation `refresh_key` does not define a `timezone`, then the `timezone` of the query will be used.

**Check List**


ovr (approved):
LGTM

Issue #8443, opened 2024-07-09
Allow to import Python files from `cube.py`


drummerwolli:
**Describe the bug**
I would like to move some code out of the cube.py file as it becomes very large and hard to maintain. But I fail to make it work in the cube docker container to import the outsourced files. also, unit testing my functions becomes impossible, because if i try to import the functions from `cube.py` the imports fail (there is no possibility to import this line: `from cube import config`).

**To Reproduce**
1. create the following files in one repo:

`cube.py`

```python
from cube import config
from cube_utils import test_function

@config('query_rewrite')
def query_rewrite(query: dict, ctx: dict) -> dict:
    query = test_function(query)
    return query
```

`cube_utils.py`

```python
def test_function(query: dict) -> dict:
    # do something
    return query
```

`Dockerfile`

```dockerfile
FROM cubejs/cube:v0.35.57

COPY cube.py /cube/conf/
COPY cube_utils.py /cube/conf/
```

2. build the docker file:

`docker build -f test.Dockerfile -t cube .`

3. run the docker container:

```bash
docker run -it cube                                         

Cube Error ---------------------------------------

Error: Python error: ModuleNotFoundError: No module named 'cube_utils'
Traceback (most recent call last):
  File "cube.py", line 2, in <module>
    from cube_utils import test_function
```

**Expected behavior**
Cube API should start without a problem with the above setup.

**Screenshots**
n/a, see above

**Minimally reproducible Cube Schema**

not needed, see above.

**Version:**
v0.35.57

drummerwolli:
one solution: add this line in your cube.py file before adding imports from other files:

`sys.path.append(os.path.dirname(os.path.abspath(__file__)))`

Aylr:
+1 vote for this feature. Thanks for the workaround in the meantime!

tbigelow-nomi:
A workaround I ended up with was setting this in my Dockerfile:

```dockerfile
ENV PYTHONPATH=/cube/conf

PR #9490, opened 2025-04-18
feat(backend-native): Allow importing modules from python files within `cube.py` and `globals.py`

KSDaemon:
This fix allows to import python modules from files relative to `cube.py`. See more info and issue description in https://github.com/cube-js/cube/issues/8443.

Closes https://github.com/cube-js/cube/issues/8443

**Check List**

### DIFF
--- a/packages/cubejs-backend-native/src/python/entry.rs
+++ b/packages/cubejs-backend-native/src/python/entry.rs
@@ -8,6 +8,18 @@ use pyo3::prelude::*;
 use pyo3::types::{PyDict, PyFunction, PyList, PyString, PyTuple};
 use std::path::Path;
 
+fn extend_sys_path(py: Python, file_name: &String) -> PyResult<()> {
+    let sys_path = py.import("sys")?.getattr("path")?.downcast::<PyList>()?;
+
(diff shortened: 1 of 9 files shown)


packages/cubejs-backend-native/src/python/entry.rs
```diff
@@ -20,6 +21,15 @@ fn python_load_config(mut cx: FunctionContext) -> JsResult<JsPromise> {
     py_runtime_init(&mut cx, channel.clone())?;
 
     let conf_res = Python::with_gil(|py| -> PyResult<CubeConfigPy> {
+        let sys_path = py.import("sys")?.getattr("path")?.downcast::<PyList>()?;
```
ovr, line 24: Let's add a small test for it. Inside `test/python.test.ts`. We can add an import in `config.py`. It should be enough to test it.

Thanks

ovr, line 24: Let's extract this code into a function, because it's used in 2 places.

ovr (approved):
LGTM with minor nit.

ovr (approved):

PR #9502, opened 2025-04-22
fix(schema-compiler): Preserve full time buckets in period-over-perio…

### LINKED ISSUE #9353: Rolling Window Calculations with Multiple Dimensions
**Describe the bug**
Topic: Rolling Window Calculations with Multiple Dimensions
Product Version:
•	Cube: 1.1.16
•	Source: Trino/Starburst
Configuration:
•	CUBESQL_SQL_PUSH_DOWN=false
•	CUBEJS_ALLOW_UNGROUPED_WITHOUT_PRIMARY_KEY=true

I am experiencing an issue with rolling window calculations when using multiple dimensions. Here is the relevant configuration:

# Rolling month calculations
```- 
- name: current_month_sum
  sql: totalprice
  type: sum
  rolling_window:
    trailing: 1 month
    offset: end

- name: previous_month_sum
  sql: totalprice
  type: sum
  rolling_window:
    trailing: 1 month
    offset: start

- name: month_over_month_ratio
  sql: "{total_price} / {order_cost}"
```
The dynamically generated query executed in Trino/Starburst uses an "INNER JOIN" between the current and previous month results:

(select …) as q_0
INNER JOIN 
(select …) as q_1 
ON (q_0."orders__clerk" = q_1."m__orders__clerk" OR 
    (q_0."m__orders__clerk" IS NULL AND q_1."m__orders__clerk" IS NULL))
AND 
(q_0."m__orders__orderdate_year" = q_1."m__orders__orderdate_year" OR 
    (q_0."m__orders__orderdate_year" IS NULL AND q_1."m__orders__orderdate_year" IS NULL))

The use of "INNER JOIN" results in output only when there is data for both the current and previous months. This becomes particularly problematic when additional dimensions are added, as it restricts the results to cases where data is present for both periods.

**Expected behavior**
I believe that using a "LEFT JOIN" or "FULL JOIN" instead of an "INNER JOIN" could resolve this issue. This would ensure that results are displayed even when one period has data, and the other does not, allowing for more accurate rolling calculations.

Example of Improved Query:
Copy Code
(select …) as q_0
LEFT JOIN 
(select …) as q_1 
ON (q_0."orders__clerk" = q_1."m__orders__clerk" OR 
    (q_0."m__orders__clerk" IS NULL AND q_1."m__orders__clerk" IS NULL))
AND 
(q_0."m__orders__orderdate_year" = q_1."m__orders__orderdate_year" OR 
    (q_0."m__orders__orderdate_year" IS NULL AND q_1."m__orders__orderdate_year" IS NULL))

By making this adjustment, we can ensure that rolling calculations are displayed even when one dimension has null values, providing a more complete and accurate representation of the data.

Request for Feedback:
Could you please review this approach and provide your thoughts on whether using a "LEFT JOIN" or "FULL JOIN" would be a suitable solution for this issue? Your guidance on this matter would be greatly appreciated.

nchristo1213:
## fix(schema-compiler): Preserve Full Time Buckets in Period-over-Period Queries

This PR replaces `INNER JOIN` with `LEFT JOIN` when stitching sub-queries in `outerMeasuresJoinFullKeyQueryAggregate`, ensuring all expected time buckets appear when calculating period-over-period (PoP) metrics.
### Problem

When PoP queries used `INNER JOIN`, buckets missing from auxiliary sub-queries (`q_1`, `q_2`, ...) were dropped causing incomplete results or incorrect ratios.
### Solution

Use `LEFT JOIN` instead, preserving all keys from the anchor query (`q_0`). Missing values in auxiliary queries yield `NULL` rather than removing the row entirely.

```js
LEFT JOIN ${this.wrapInParenthesis((q))} as q_${i + 1} ON ${this.dimensionsJoinCondition(`q_${i}`, `q_${i + 1}`)}
```
### Benefits

- Retains continuity for PoP metrics
- Allows ratio calculations to appear accurate when prior period data is missing
- Ensures accuracy without introducing gaps or dropped time buckets
### Tests

- **New fixture**: `sales` cube with `date`, `category`, `region`, and PoP measures
- **Adjusted**:
    - `CAGR`: Added `NULL` rows
    - `multi-stage graph` tests: Added secondary ordering
- **New tests**:
    - Time-only PoP
    - One-dimension PoP
    - Two-dimension PoP
    
These tests ensure the correct number of periods appear and that PoP measures return `NULL` instead of dropping rows when previous-period values are unavailable.

**Check List**

resolves #9353


nchristo1213:
Hi folks, just following up on this PR. I saw that a review was requested from the schema-compiler team. If there’s anything I can clarify or update, I’d be happy to help. Thanks in advance!

KSDaemon (approved):
👍🏻 LGTM!

packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
```diff
@@ -1083,7 +1083,7 @@ export class BaseQuery {
     const join = R.drop(1, toJoin)
       .map(
         (q, i) => (this.dimensionAliasNames().length ?
-          `INNER JOIN ${this.wrapInParenthesis((q))} as q_${i + 1} ON ${this.dimensionsJoinCondition(`q_${i}`, `q_${i + 1}`)}` :
+          `LEFT JOIN ${this.wrapInParenthesis((q))} as q_${i + 1} ON ${this.dimensionsJoinCondition(`q_${i}`, `q_${i + 1}`)}` :
```
KSDaemon, line 1086: What if there are few subjoins? And it's probably possible that the result set with gaps may occur on the right side too... So maybe `FULL OUTER JOIN` should be used here?

nchristo1213: You are right. I was thinking the root q_0 would always have the full time set, but `FULL OUTER JOIN` will work too. I can make the code and description updates.

nchristo1213: After more analysis, `FULL OUTER JOIN` won't work here because it seems PostgreSQL requires merge-joinable or hash-joinable join conditions and `dimensionsJoinCondition` includes other null-handling logic that prevents that (found by re-running tests after just changing to `FULL OUTER JOIN`). `LEFT JOIN` still seems appropriate because we can assume the first query `q_0` defines the full time range and that is what is preserved when data is missing from later subqueries. The same logic would apply in cases with few subjoins, `q_0` is treated as the time source subquery and `LEFT JOIN` ensures we retain those rows when there is missing data on the right.

PR #9517, opened 2025-04-25
fix(schema-compiler): Fix non-match for exact custom granularity pre-aggregation

KSDaemon:
This fixes cases when a query with custom granularity doesn't hit the pre-aggregation with the same custom granularity.

**Check List**


mcheshkov (approved):
LGTM, but I'm not the right person to ask for intricacies of granularities
I'd add one more test for query with preaggregation references alone, without date range

waralexrom (approved):

waralexrom (approved):

PR #9529, opened 2025-04-30
feat(databricks-jdbc-driver): Implement connection checking without waking up SQL warehouse

KSDaemon:
This updates Databricks test connection flow to use REST API instead of establishing real JDBC connection (which prevents SQL WH from suspending)

**Check List**


packages/cubejs-databricks-jdbc-driver/src/DatabricksDriver.ts
```diff
@@ -262,38 +271,46 @@ export class DatabricksDriver extends JDBCDriver {
...
+    });
+
+    if (!res.ok) {
+      throw new Error(`Databricks API error: ${res.statusText}`);
+    }
+
+    // const _data = await res.json();
+
+    // We don't need to check the actual status, because SQL WH will auto resume on real connection
+    // if (data.state !== 'RUNNING') {
```
mcheshkov, line 308: I think it should check anyway, at least for non `DELETING | DELETED` status

Also `health` field looks interesting, like when `health.status` is `FAILED` we can fail early this test.
https://docs.databricks.com/api/workspace/warehouses/get#health

KSDaemon: All right! You got it!) Let's make it good! :)

mcheshkov (approved):

PR #9537, opened 2025-05-01
feat(cubesql): Data source per member

mcheshkov:
**Check List**

This should fix SQL push-down targeting views.
Before this views would always have `default` datasource, and could use incorrect dialect in multi-datasource deployments.
Now JS-side would pass "member -> data source" mapping to SQL API, and this would reflect each view member properly. Using members as keys is necessary for multi-data-source views (even if they are not supported right now in SQL pushdown, they are valid as data model, and can be used with `rollupJoin` pre-aggregations)


MazterQyou (approved):

KSDaemon (commented):
👍🏻

PR #9538, opened 2025-05-01
chore(schema-compiler): move preAggregations class to typescript

KSDaemon:
Finally, move PreAggregations from js to ts with most of the types described! 🎉

Also switching to TS uncovered some incorrectness that is now fixed!

### DIFF
--- a/packages/cubejs-schema-compiler/src/adapter/PreAggregations.ts
+++ b/packages/cubejs-schema-compiler/src/adapter/PreAggregations.ts
@@ -1,4 +1,3 @@
-// @ts-nocheck
 import R from 'ramda';
 
 import { CubeSymbols } from '../compiler/CubeSymbols';
--- a/packages/cubejs-schema-compiler/src/adapter/PreAggregations.ts
+++ b/packages/cubejs-schema-compiler/src/adapter/PreAggregations.ts
@@ -1,15 +1,85 @@
+// @ts-nocheck
 import R from 'ramda';
 
 import { CubeSymbols } from '../compiler/CubeSymbols';
 import { UserError } from '../compiler/UserError';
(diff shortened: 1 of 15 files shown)


ovr (approved):
🔥 Awesome! Great to see that it finally happens and TS is here for pre-aggs!

packages/cubejs-schema-compiler/src/adapter/PreAggregations.ts
```diff
@@ -1,40 +1,109 @@
+// @ts-nocheck
```
mcheshkov, line 1: Is it necessary here?

packages/cubejs-schema-compiler/src/compiler/CubeEvaluator.ts
```diff
@@ -66,7 +66,33 @@ export type PreAggregationFilters = {
   scheduled?: boolean,
 };
 
-type PreAggregationDefinition = {
+export type EveryInterval = string;
```
mcheshkov, line 69: Does this need something like a `TODO: Write a real type` comment?

mcheshkov (approved):
I think we can land this for integration purposes, and improve on it later.

packages/cubejs-schema-compiler/src/adapter/PreAggregations.ts
KSDaemon: Nope! I was using that during debugging... Good catch! I'll remove it of course!

packages/cubejs-schema-compiler/src/compiler/CubeEvaluator.ts
KSDaemon: YES! Thxn! I'll add it!

Issue #9435, opened 2025-04-07
Filter by time measure doesn't work


Anger79:
**Problem**

I don't understand how to apply filter by time measure (`timestampMin` or `timestampMax`) in a cubejs playground and whether it's possible or not. I'm getting "`TypeError: Cannot read properties of null (reading 'split')`" when I'm trying to add a filter like `timestampMin` `in date range`/`not in date range`/`before date`/`after date`. No sql generated, "Run Query" button is inactive.

 I do able to filter by requests measure.

**Cube.js schema**
```javascript
cube(`Cube`, {
  ...
    measures: {
        requests: {
            sql: `count`,
            type: `sum`
        },
        timestampMin: {
            sql: `min(timestamp)`,
            type: `time`
        },
        timestampMax: {
            sql: `max(timestamp)`,
            type: `time`
        },
    },
    dimensions: {
        ....
        timestamp: {
            sql: `timestamp`,
            type: `time`
        },
        .....
})
```
**Cubejs version**: v1.2.30

**Data source**: Clickhouse

igorlukanin:
Hi @Anger79 👋 

You can add a filter by time dimension in the following way. Please check screenshots:

<img width="1392" alt="Image" src="https://github.com/user-attachments/assets/f4221db9-3868-4e8d-8974-dcb4a3569f4c" />

<img width="1392" alt="Image" src="https://github.com/user-attachments/assets/1256f182-9065-465b-ac24-2b72fdb902e6" />

That one will translate to the `dateRange` parameter in the `timeDimension` part of the query.

`inDateRange`, `notInDateRange`, `beforeDate`, `afterDate` filter types are currently not supported in Playground. Anyway, you'd probably be better using the `dateRange` parameter in `timeDimension`.

Anger79:
Hi @igorlukanin, Thanks for a quick reply!

But I'm interested in filtering exactly by the time measure, not a time dimension.
There is a difference between fitlering by dimension and by measure, right? Because filter by measure gets converted to the HAVING clause and applied after aggregation in sql query, while fitlering by dimension gets converted to a regular WHERE clause.

So I'm trying to understand if there is a way to apply filter by time measure in a cubejs query, and what is the proper way of doing that if that is supported.

Would be great if you provide query example.
Thanks!

igorlukanin:
> But I'm interested in filtering exactly by the time measure, not a time dimension.

[...]

PR #9544, opened 2025-05-06
fix(schema-compiler): Fix filtering by time measure

KSDaemon:
Fixes errors for queries with measures with type `time` used as date filters.

**Check List**

Closes https://github.com/cube-js/cube/issues/9435

### DIFF
--- a/packages/cubejs-schema-compiler/src/adapter/BaseFilter.ts
+++ b/packages/cubejs-schema-compiler/src/adapter/BaseFilter.ts
@@ -90,6 +90,17 @@ export class BaseFilter extends BaseDimension {
     }
   }
 
+  /**
+   * BaseFilter inherits from BaseDimension while Filter may be measure-based !!
+   */
+  public override dateFieldType() {
+    if (this.measure) {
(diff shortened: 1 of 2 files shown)


packages/cubejs-schema-compiler/src/adapter/BaseFilter.ts
```diff
@@ -90,6 +90,17 @@ export class BaseFilter extends BaseDimension {
     }
   }
 
+  /**
+   * BaseFilter inherits from BaseDimension while Filter may be measure-based !!
+   */
+  public override dateFieldType() {
+    if (this.measure) {
+      return this.measureDefinition().type; // There is no fieldType in measure, but it seems that it's enough
```
ovr, line 98: > Added line #L98 was not covered by tests

Can we add a test case for this?

KSDaemon: It's showing as `uncovered` because for now coverage is being collected only for `unit` tests. But this is an `integration` test

waralexrom (approved):

PR #9554, opened 2025-05-07
fix(schema-compiler): Collect join hints from subquery join conditions

mcheshkov:
**Check List**

This is necessary to support queries where output references only join subqueries.
Before, when query does have subqeury joins, but does not have any measures/dimensions/... join hints would be empty, and join tree would be `null`, which is almost okay, but to generate query with join subqueries we need to build join query, so join tree is more-or-less required.


ovr (approved):

KSDaemon (approved):
👍🏻  LGTM!

PR #9561, opened 2025-05-09
feat: Rewrite joins from SQL as query-level join hints

mcheshkov:
**Check List**

This allows to keep join structure in some cases. Consider query like this:
```
SELECT cube2.dim, cube3.dim
FROM cube1
JOIN cube2 ON cube1.__cubeJoinField = cube2.__cubeJoinField
JOIN cube3 ON cube1.__cubeJoinField = cube3.__cubeJoinField
GROUP BY 1, 2
```
It references members only from `cube2` and `cube3`, but join structure actually has `cube1` as root, and it is completely missing from query.

Now every join will generate separate join hint during rewrite, and then they will get to JS side as a query-level join hints.


waralexrom (approved):

igorlukanin (changes requested):
Let's disallow to run "external" queries that contain join hints. Join hints should only be accepted for queries that have an "internal" origin, i.e., initiated by Cube as part of the SQL API processing. Consequently, a REST API query with join hints should always fail with a clear error message.

paveltiunov (approved):

ovr (approved):

PR #9563, opened 2025-05-13
chore(client/ws-transport): Remove old manually created index.d.ts

ovr:
Previously, it was used for doc generation, but it's not used anymore.


packages/cubejs-client-ws-transport/package.json
```diff
@@ -26,7 +26,6 @@
     "ws": "^7.3.1"
   },
   "files": [
-    "src",
```
ovr, line 29: We don't need to ship source files.

KSDaemon (approved):

PR #9577, opened 2025-05-16
feat(tesseract): Initial BigQuery support

waralexrom:
**Check List**

### DIFF
--- a/packages/cubejs-schema-compiler/src/adapter/BigqueryQuery.ts
+++ b/packages/cubejs-schema-compiler/src/adapter/BigqueryQuery.ts
@@ -261,13 +261,24 @@ export class BigqueryQuery extends BaseQuery {
     templates.expressions.timestamp_literal = 'TIMESTAMP(\'{{ value }}\')';
     delete templates.expressions.ilike;
     delete templates.expressions.like_escape;
+    templates.filters.like_pattern = 'CONCAT({% if start_wild %}\'%\'{% else %}\'\'{% endif %}, LOWER({{ value }}), {% if end_wild %}\'%\'{% else %}\'\'{% endif %})';
+    templates.tesseract.ilike = 'LOWER({{ expr }}) {% if negated %}NOT {% endif %} LIKE {{ pattern }}';
     templates.types.boolean = 'BOOL';
     templates.types.float = 'FLOAT64';
     templates.types.double = 'FLOAT64';
     templates.types.decimal = 'BIGDECIMAL({{ precision }},{{ scale }})';
     templates.types.binary = 'BYTES';
+    templates.expressions.cast_to_string = 'CAST({{ expr }} AS STRING)';
     templates.operators.is_not_distinct_from = 'IS NOT DISTINCT FROM';
     templates.join_types.full = 'FULL';
+    templates.statements.time_series_select = 'SELECT DATETIME(TIMESTAMP(f)) date_from, DATETIME(TIMESTAMP(t)) date_to \n' +
+    'FROM (\n' +
+    '{% for time_item in seria  %}' +
+    '    select \'{{ time_item[0] }}\' f, \'{{ time_item[1] }}\' t \n' +
+    '{% if not loop.last %} UNION ALL\n{% endif %}' +
+    '{% endfor %}' +
+    ') AS dates';
+      
     return templates;
   }
 }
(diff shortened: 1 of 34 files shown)


paveltiunov (approved):

packages/cubejs-schema-compiler/src/adapter/BigqueryQuery.ts
```diff
@@ -261,13 +261,24 @@ export class BigqueryQuery extends BaseQuery {
...
+    templates.tesseract.ilike = 'LOWER({{ expr }}) {% if negated %}NOT {% endif %} LIKE {{ pattern }}';
     templates.types.boolean = 'BOOL';
     templates.types.float = 'FLOAT64';
     templates.types.double = 'FLOAT64';
     templates.types.decimal = 'BIGDECIMAL({{ precision }},{{ scale }})';
     templates.types.binary = 'BYTES';
+    templates.expressions.cast_to_string = 'CAST({{ expr }} AS STRING)';
     templates.operators.is_not_distinct_from = 'IS NOT DISTINCT FROM';
     templates.join_types.full = 'FULL';
+    templates.statements.time_series_select = 'SELECT DATETIME(TIMESTAMP(f)) date_from, DATETIME(TIMESTAMP(t)) date_to \n' +
```
KSDaemon, line 274: It would be much nicer to define it in backtick multiline.

packages/cubejs-testing-drivers/fixtures/bigquery.json
```diff
@@ -17,7 +17,8 @@
       "CUBESQL_SQL_PUSH_DOWN": "true",
 
       "CUBEJS_DB_EXPORT_BUCKET": "cube-open-source-export-bucket",
-      "CUBEJS_DB_EXPORT_BUCKET_TYPE": "gcp"
+      "CUBEJS_DB_EXPORT_BUCKET_TYPE": "gcp",
+      "_CUBEJS_TESSERACT_SQL_PLANNER": "true"
```
KSDaemon, line 21: Is it intentional to have an underscore prefix here?

packages/cubejs-testing-drivers/fixtures/postgres.json
```diff
@@ -12,7 +12,8 @@
       "CUBEJS_PG_SQL_PORT": "5656",
       "CUBEJS_SQL_USER": "admin",
       "CUBEJS_SQL_PASSWORD": "admin_password",
-      "CUBESQL_SQL_PUSH_DOWN": "true"
+      "CUBESQL_SQL_PUSH_DOWN": "true",
+      "_CUBEJS_TESSERACT_SQL_PLANNER": "true"
```
KSDaemon, line 16: Is it intentional to have an underscore prefix here?

rust/cubesqlplanner/cubesqlplanner/src/planner/sql_evaluator/sql_nodes/auto_prefix.rs
```diff
@@ -5,6 +5,8 @@ use crate::planner::sql_evaluator::MemberSymbol;
 use crate::planner::sql_evaluator::SqlEvaluatorVisitor;
 use crate::planner::sql_templates::PlanSqlTemplates;
 use cubenativeutils::CubeError;
+use lazy_static::lazy_static;
```
KSDaemon, line 8: Why not use smth like [OnceCell ](https://doc.rust-lang.org/std/cell/struct.OnceCell.html) or [OnceLock](https://doc.rust-lang.org/std/sync/struct.OnceLock.html) (if thread-safe is needed) from the std library rather than depend on an external crate?

KSDaemon (commented):
Looks good! Left a few minor questions and suggestions.

rust/cubesqlplanner/cubesqlplanner/src/planner/sql_evaluator/sql_nodes/auto_prefix.rs
waralexrom: Because to do that, I’d need to rewrite them across the whole Tesseract. I’ll do it, but as a separate task. For now, it’s better to use lazy_static! everywhere than end up with a zoo.

packages/cubejs-schema-compiler/src/adapter/BigqueryQuery.ts
waralexrom: That’s a debatable suggestion. Manual line concatenation helps keep both the Jinja logic and the rendered result readable.
```
'{% for time_item in seria  %}' +
    '    select \'{{ time_item[0] }}\' f, \'{{ time_item[1] }}\' t \n' +
'{% if not loop.last %} UNION ALL\n{% endif %}' +
'{% endfor %}' +
```
Converting it to a backtick multiline would cause an extra newline to be inserted before every select in the output.

And if I were to collapse the for and select into the same line, the template would become much harder to read — and the rendered SQL would also lose formatting by placing all SELECTs on one line.

KSDaemon (approved):

Issue #9619, opened 2025-05-26
Misconfiguration of "main" fields in package.json of cube clients


Graphmaxer:
**Describe the bug**

The cube client are not pointing to an existing file in the "main" field of package.json. Resulting in a import error "Error: Failed to resolve entry for package "@cubejs-client/vue3". The package may have incorrect main/module/exports specified in its package.json."

**To Reproduce**

- Install the 1.3.16 cube client for Vue 3
- Setup a vite project with vitest
- When running tests the error occurs

**Expected behavior**

The configuration of the client should be good and not generating errors.

**Screenshots**

Error generated due to bad configuration (in Vite):

![Image](https://github.com/user-attachments/assets/39ab51bf-e1c4-407d-ac7b-99c57e03e711)

In the new 1.3.16 version (not working):

https://www.npmjs.com/package/@cubejs-client/vue3/v/1.3.16?activeTab=code

![Image](https://github.com/user-attachments/assets/10a5b789-39d9-4e52-b312-62f6679d0e51)

In the old 1.3.15 verrsion (working):

https://www.npmjs.com/package/@cubejs-client/vue3/v/1.3.15?activeTab=code

![Image](https://github.com/user-attachments/assets/007fdccb-bfc2-4f3e-a9c5-457fb1fd5823)

**Version:**

1.3.16

**Additional context**

This new error is caused by the refactor of the cube core client to TS https://github.com/cube-js/cube/pull/9562

johnhunter:
Same issue with @cubejs-client/react

johnhunter:
1.3.18 does not fix the issue with @cubejs-client/react

PR #9620, opened 2025-05-26
fix(client): update main fields in package.json to cjs files

Graphmaxer:

https://github.com/cube-js/cube/issues/9619


Graphmaxer:
Yes, just changing this field in the file `node_modules/@cubejs-client/vue3/package.json` in my project fixes the issue.

Issue #9303, opened 2025-03-05
Trino user may not have permissions to access `/v1/node`


Priyansh121096:
**Describe the bug**
#9109 seems to have broken the Trino driver. This is because the Trino driver internally uses the PrestoDB driver and Trino does not have a `/v1/node` REST API similar to PrestoDB. 

The same works with v1.1.16 which was the latest release which didn't include #9109.

**To Reproduce**
1. Try hitting a Trino connection from the Playground with v1.2.16.
2. See error:
```
Error: node list api returns error:<html>
<head>
<meta http-equiv="Content-Type" content="text/html;charset=ISO-8859-1"/>
<title>Error 403 Forbidden</title>
</head>
<body>
<h2>HTTP ERROR 403 Forbidden</h2>
<table>
<tr><th>URI:</th><td>https://your.trino.host.com/v1/node</td></tr>
<tr><th>STATUS:</th><td>403</td></tr>
<tr><th>MESSAGE:</th><td>Forbidden</td></tr>
</table>

</body>
</html>
```

**Expected behavior**
Trino driver should work as expected.

**Version:**
[e.g. 1.2.16]

igorlukanin:
Hi @Priyansh121096 👋 

> Trino does not have a /v1/node REST API similar to PrestoDB.

Are you sure? I see `/v1/node` mentioned here in Trino docs: https://trino.io/docs/current/release/release-0.54.html

Priyansh121096:
> Hi [@Priyansh121096](https://github.com/Priyansh121096) 👋
> 
> > Trino does not have a /v1/node REST API similar to PrestoDB.
> 
> Are you sure? I see `/v1/node` mentioned here in Trino docs: https://trino.io/docs/current/release/release-0.54.html

Ah, thanks for pointing this out @igorlukanin. It may be that the trino instance I'm using does not allow me to hit this endpoint.

allekai:
This issue affects us as well. In our case, it seems we won't get the permission on the `/v1/node` endpoint. 

@igorlukanin As I understand, this request is made to test the connection. To me it seems that most of the other data source drivers use a `SELECT 1` to check the connection. Wouldn't that make sense for the Trino driver as well?

KSDaemon:
[...]

PR #9634, opened 2025-05-30
feat(trino-driver): Add special testConnection for Trino

KSDaemon:
This PR includes a special `testConnection()` implementation for Trino, as the one for PrestoDB is not working for Trino Server.

Also updated the "presto-client" to the latest "^1.1.0".

This fixes https://github.com/cube-js/cube/issues/9303

**Check List**

### DIFF
--- a/packages/cubejs-trino-driver/src/TrinoDriver.ts
+++ b/packages/cubejs-trino-driver/src/TrinoDriver.ts
@@ -1,5 +1,6 @@
+import fetch from 'node-fetch';
 import { PrestoDriver } from '@cubejs-backend/prestodb-driver';
-import { PrestodbQuery } from '@cubejs-backend/schema-compiler/dist/src/adapter/PrestodbQuery';
+import { PrestodbQuery } from '@cubejs-backend/schema-compiler';
 
 export class TrinoDriver extends PrestoDriver {
   public constructor(options: any) {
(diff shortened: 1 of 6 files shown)


packages/cubejs-trino-driver/src/TrinoDriver.ts
```diff
@@ -9,4 +10,24 @@ export class TrinoDriver extends PrestoDriver {
...
+  public override async testConnection(): Promise<void> {
+    const { host, port, ssl, basic_auth: basicAuth } = this.config;
+    const protocol = ssl ? 'https' : 'http';
+    const url = `${protocol}://${host}:${port}/v1/info`;
+    const headers: Record<string, string> = {};
+
+    if (basicAuth) {
+      const { user, password } = basicAuth;
+      const encoded = Buffer.from(`${user}:${password}`).toString('base64');
+      headers.Authorization = `Basic ${encoded}`;
```
Madisonw, line 23: I'm not sure how the auth works through the BaseDriver stuff, but wouldnt this preclude authenticating with anything but basic auth? As i understand,  you can't auth with Trino or Presto via jwt for example, but i'd imagine you'd want to be futureproof for that change.

KSDaemon: @Madisonw you're right! But I look at the `PrestoDriverConfiguration` options, and I see only basicAuth is supported there... no other options are passed to the underlying driver. So whenever we have a need for some other auth flows - they need to be added to options and implemented accordingly. At that time we may update our `testConnection()` flow too.

Also `testConnection()` uses just common `fetch` over HTTP(s) protocol, so nothing special about auth or BaseDriver here...

KSDaemon: I'll implement JWTs auth in a separate PR.

PR #9659, opened 2025-06-05
feat(schema-compiler,api-gateway): Nested folders support

KSDaemon:
This PR brings nested folder support into views.
So now you can define this:

```yaml
  - name: test_view
    folders:
      - name: folder1
        includes:
          - users_city
          - name: inner folder 2
            includes:
              - renamed_orders_status
          - name: inner folder 3
            includes: "*"
```

`/meta` endpoint was extended with additional cube property `nestedFolders`, which holds the recursive nested folders structure. Original `folders` property is kept untouched for backward compatibility.

Additional env flag `CUBEJS_NESTED_FOLDERS_DELIMITER` was introduced. It controls how to form the flatten `folders` property in meta data. If it is not set - all nested folders members will be merged to the root folder members. If it is set to some string (e.g. "/"), then nested folders are bubbled to the root level with their names being constructed as `root folder/nested folder/deep nested` simulating a nested structure for those APIs (probably some BI tools) that don't support nested folders out of the box.

**Check List**

### DIFF
--- a/docs/pages/product/apis-integrations.mdx
+++ b/docs/pages/product/apis-integrations.mdx
@@ -45,10 +45,11 @@ for an unofficial, community-maintained [client library for Python](https://gith
 Support for data modeling features differ across APIs, integrations, and [visualization
 tools][ref-viz-tools]. Some of the features with partial support are listed below:
 
-| Feature | ✅ Supported in | ❌ Not supported in |
-| --- | --- | --- |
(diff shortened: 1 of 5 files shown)


packages/cubejs-api-gateway/openspec.yml
```diff
@@ -173,7 +173,9 @@ components:
         members:
           type: array
           items:
-            type: "string"
+            oneOf:
+              - type: string
+              - $ref: "#/components/schemas/V1CubeMetaFolder"
```
ovr, line 178: We need to regenerate the client, because otherwise, defining nested folders will break SQL API, MDX & DAX API, while it's not possible to deserialize such a meta response.

Thanks

ovr (changes requested):
Please regenerate `/rust/cubeclient` client. Thanks

MazterQyou (approved):

ovr (approved):

PR #9665, opened 2025-06-06
fix(cubesql): Propagate errors from SqlAuthService to the user

KSDaemon:
When using cube.py it’s fundamental to see errors and even better the stack trace of errors generated from python code, currently the logs doesn’t show any errors generated from python check_sql_auth.

This PR propagates such errors to the log.

Imaging this cube.py config:
```py
# Validate the username and password of the user submitting the API request
@config('check_sql_auth')
def check_sql_auth(query: dict, username: str, password: str) -> dict:
  raise Exception('You should be able to see this error in the logs')
```

And let's try to connect to Cube via psql:
```sh
❯ psql -h 127.0.0.1 -p 15432 -U cube
psql: error: connection to server at "127.0.0.1", port 15432 failed: FATAL:  password authentication failed for user "cube"
```

And here is what we now can see in logs:
```
cubejs-server
🔥 Cube Store (1.3.19) is assigned to 3030 port.
🔓 Authentication checks are disabled in developer mode. Please use NODE_ENV=production to enable it.
🦅 Dev environment available at http://localhost:4000
🔗 Cube SQL (pg) is listening on 0.0.0.0:15432
🚀 Cube API server (1.3.19) is listening on 4000
Auth Error: undefined
{}
Error: Python error: Exception: You should be able to see this error in the logs
Traceback (most recent call last):
  File "cube.py", line 14, in check_sql_auth
    raise Exception('You should be able to see this error in the logs')

```

**Check List**

### DIFF
--- a/packages/cubejs-api-gateway/src/sql-server.ts
+++ b/packages/cubejs-api-gateway/src/sql-server.ts
@@ -128,14 +128,25 @@ export class SQLServer {
         };
       },
       checkSqlAuth: async ({ request, user, password }) => {
-        const { password: returnedPassword, superuser, securityContext, skipPasswordCheck } = await checkSqlAuth(request, user, password);
+        try {
+          const { password: returnedPassword, superuser, securityContext, skipPasswordCheck } = await checkSqlAuth(request, user, password);
 
-        return {
-          password: returnedPassword,
-          superuser: superuser || false,
-          securityContext,
-          skipPasswordCheck,
-        };
+          return {
+            password: returnedPassword,
+            superuser: superuser || false,
+            securityContext,
+            skipPasswordCheck,
+          };
+        } catch (e) {
+          this.apiGateway.log({
+            type: 'Auth Error',
+            protocol: (request as any).protocol,
+            method: (request as any).method,
+            apiType: 'sql',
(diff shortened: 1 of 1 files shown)


rust/cubesql/cubesql/src/sql/postgres/pg_auth_service.rs
```diff
@@ -87,15 +87,14 @@ impl PostgresAuthService for PostgresAuthServiceDefaultImpl {
...
-            ))
-        };
-
-        let Ok(authenticate_response) = authenticate_response else {
-            return auth_fail();
+        let authenticate_response = match authenticate_response {
+            Ok(r) => r,
+            Err(err) => {
+                return AuthenticationStatus::Failed(format!(
+                    "password authentication failed for user \"{}\". \nError: {}",
```
ovr, line 94: It's not safe to expose all error messages to end user. For example:

user can define `check_sql_auth` with HTTP client and it can throw something secret in exception's message. Like URL of JWT keys or database URL, which contains all parameters.

I belive, we should change message, that error is happening, please take a look at logs. And print it to logs as error.

cC @igorlukanin

PR #9673, opened 2025-06-11
fix(tesseract): Fix rolling window with few time dimensions, filter_group in segments and member expressions

waralexrom:
**Check List**

### DIFF
--- a/rust/cubesqlplanner/cubesqlplanner/src/planner/base_measure.rs
+++ b/rust/cubesqlplanner/cubesqlplanner/src/planner/base_measure.rs
@@ -152,7 +152,7 @@ impl BaseMeasure {
         }))
     }
 
-    pub fn can_used_as_addictive_in_multplied(&self) -> bool {
+    pub fn can_be_used_as_additive_in_multplied(&self) -> bool {
         if let Ok(measure_symbol) = self.member_evaluator.as_measure() {
             measure_symbol.can_used_as_addictive_in_multplied()
         } else {
--- a/rust/cubesqlplanner/cubesqlplanner/src/planner/query_properties.rs
+++ b/rust/cubesqlplanner/cubesqlplanner/src/planner/query_properties.rs
@@ -896,7 +896,7 @@ impl QueryProperties {
                             .rendered_as_multiplied_measures
                             .insert(item.measure.full_name());
                     }
-                    if item.multiplied && !item.measure.can_used_as_addictive_in_multplied() {
+                    if item.multiplied && !item.measure.can_be_used_as_additive_in_multplied() {
                         result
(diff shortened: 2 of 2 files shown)


rust/cubesqlplanner/cubesqlplanner/src/planner/planners/multiplied_measures_query_planner.rs
```diff
@@ -299,7 +299,7 @@ impl MultipliedMeasuresQueryPlanner {
             schema,
             filter: logical_filter,
             offset: self.query_properties.offset(),
```
UsmanYasin, line 301: Nice catch there. I noticed the offset is still present here and wanted to share some thoughts (though I'm still familiarizing myself with the codebase, so please correct me if I'm missing something).

From what I understand, there might be a few considerations regarding offset in subqueries:
1. **Database compatibility:** Different databases have varying requirements for offset usage:
MSSQL requires an ORDER BY clause when using OFFSET
MySQL requires OFFSET to be paired with a LIMIT clause
2. **Pagination concerns:** I'm wondering if pagination logic should be restricted to the outer query only? My understanding is that applying pagination to inner queries could potentially lead to data inconsistencies, especially when the API request includes result pagination.

Would it make sense to remove the offset from this subquery as well? I'd appreciate your thoughts on this, as you likely have more context about the intended behavior here.

rust/cubesqlplanner/cubesqlplanner/src/planner/base_measure.rs
```diff
@@ -138,36 +136,28 @@ impl BaseMeasure {
...
-        let measure_type = self.measure_type();
-        let res = if measure_type == "countDistinct" || measure_type == "countDistinctApprox" {
-            true
-        } else if measure_type == "count" {
-            if let Some(definition) = &self.definition {
-                !definition.has_sql()?
-            } else {
-                false
-            }
+    pub fn can_used_as_addictive_in_multplied(&self) -> bool {
```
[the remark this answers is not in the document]
KSDaemon: Measures are not about drugs :)

rust/cubesqlplanner/cubesqlplanner/src/planner/base_measure.rs
```diff
@@ -138,36 +136,28 @@ impl BaseMeasure {
...
+        if let Ok(measure_symbol) = self.member_evaluator.as_measure() {
+            measure_symbol.can_used_as_addictive_in_multplied()
```
KSDaemon, line 157: Would be nice to rename this too:
```suggestion
            measure_symbol.can_be_used_as_additive_in_multplied()
```

PR #9674, opened 2025-06-12
fix(schema-compiler): Avoid ambiguous dimension column mappings for ClickHouse queries 

### LINKED ISSUE #9383: Error mapping Clickhouse result set when using joins
**Describe the bug**
An error is reported while executing queries involving cubes with joins. This happens only for Clickhouse datasources (the same query works for Mysql datasources). The generated SQL query is valid and can be executed successfully, so the issue seems to occur while mapping the result set values. The reported error message is `Error: Member name not found for alias: 'q_0.acube__id'` (no stacktrace is present).

**To Reproduce**
Execute the following query via the REST API:
```
{
  "dimensions": [
    "Acube.id"
  ],
  "measures": [
    "Acube.AcubeTotalValue",
    "Acube.BcubeTotalValue",
    "Acube.CcubeTotalAmount"
  ]
}
```

Equivalent SQL API query:
```
SELECT
  Acube.id,
  MEASURE(Acube.AcubeTotalValue),
  MEASURE(Acube.BcubeTotalValue),
  MEASURE(Acube.CcubeTotalAmount)
FROM
  Acube
GROUP BY
  1
LIMIT
  10000;
```

To test the failing query, a minimal set of 3 joined cubes was used:

```javascript
cube(`Acube`, {
    sql: `SELECT 1 as id, 'Category A' as category, 10 as value UNION ALL 
         SELECT 2, 'Category B', 20 UNION ALL 
         SELECT 3, 'Category C', 30`,
  
    joins: {
      Bcube: {
        relationship: `one_to_many`,
        sql: `${Acube}.id = ${Bcube}.id`
      },
      Ccube: {
        relationship: `one_to_many`,
        sql: `${Acube}.id = ${Ccube}.id`
      },      
    },
  
    measures: {
      AcubeCount: {
        type: `count`,
      },
      AcubeTotalValue: {
        type: `sum`,
        sql: `value`
      },
      BcubeTotalValue: {
        sql: `${Bcube.totalValue}`,
        type: `number`
      },
      CcubeTotalAmount: {
        sql: `${Ccube.totalAmount}`,
        type: `number`
      },
    },
  
    dimensions: {
      id: {
        sql: `id`,
        type: `number`,
        primaryKey: true,
        shown: true
      },
      category: {
        sql: `category`,
        type: `string`
      },
      appId: {
        sql: `'fmgutp'`,
        type: `string`
      }
    }
  });
  
cube(`Bcube`, {
    sql: `SELECT 1 as id, 'Bgroup1' as groupName, 50 as value UNION ALL 
         SELECT 2, 'Bgroup2', 60 UNION ALL 
         SELECT 3, 'Bgroup3', 70`,
  
    measures: {
      count: {
        type: `count`
      },
      totalValue: {
        type: `sum`,
        sql: `value`
      }
    },
  
    dimensions: {
      id: {
        sql: `id`,
        type: `number`,
        primaryKey: true
      },
      groupName: {
        sql: `groupName`,
        type: `string`
      }
    }
  });
  
cube(`Ccube`, {
    sql: `SELECT 1 as id, 'Ctype1' as type, 15 as amount UNION ALL 
         SELECT 2, 'Ctype2', 25 UNION ALL 
         SELECT 3, 'Ctype3', 35`,
  
    measures: {
      count: {
        type: `count`
      },
      totalAmount: {
        type: `sum`,
        sql: `amount`
      }
    },
  
    dimensions: {
      id: {
        sql: `id`,
        type: `number`,
        primaryKey: true,
        shown: true
      },
      type: {
        sql: `type`,
        type: `string`
      }
    }
  });
```

**Version:**
v1.2.26 (Docker image)

dragosrep:
**Check List**

This fixes #9383

The root cause of this issue is the ClickHouse server that sometimes includes the table prefix in the column name when returning the result set. For example, the query:
```
SELECT
    q_0.id,
    q_0.cat,
    value_0,
    value_1
FROM  (
    SELECT 1 as id, 'a' as cat, 100 as value_0
) q_0
INNER JOIN (
    SELECT 1 as id, 'b' as cat, 200 as value_1
) q_1 ON q_0.id=q_1.id
```
returns:
image
,while:
```
SELECT
    q_0.id,
    q_0.cat,
    value_0,
    value_1
FROM  (
    SELECT 1 as id, 'a' as cat, 100 as value_0
) q_0
INNER JOIN (
    SELECT 1 as id, 'b' as cat, 200 as value_1
) q_1 ON q_0.id=q_1.id
INNER JOIN (
    SELECT 1 as id, 'c' as cat, 300 as value_2
) q_2 ON q_0.id=q_2.id
```
has the following result set:
image

Similarly, SQL result sets from complex analytics queries with cube joins sometimes end up with column names having the `q_0` prefix, leading to an error when trying to map them to a dimension.

The overriden `dimensionColumns` explicitly sets the column name to avoid this problem. Without it, one of the two provided integration tests fails.

### DIFF
--- a/packages/cubejs-schema-compiler/src/adapter/ClickHouseQuery.ts
+++ b/packages/cubejs-schema-compiler/src/adapter/ClickHouseQuery.ts
@@ -266,6 +266,17 @@ export class ClickHouseQuery extends BaseQuery {
     return `ALTER TABLE ${tableName} ADD INDEX ${indexName} (${escapedColumns.join(', ')}) TYPE minmax GRANULARITY 1`;
   }
 
+  public dimensionColumns(cubeAlias) {
+    // For the top-level SELECT statement, explicitly set the column alias.
+    // Clickhouse sometimes includes the "q_0" prefix in the column name, and this
+    // leads to errors during the result mapping.
+    if (cubeAlias === 'q_0') {
+      return this.dimensionAliasNames().map(alias => `${cubeAlias && `${cubeAlias}.` || ''}${alias} ${alias}`);
+    } else {
+      return super.dimensionColumns(cubeAlias);
+    }
+  }
+
   public sqlTemplates() {
     const templates = super.sqlTemplates();
     templates.functions.DATETRUNC = 'DATE_TRUNC({{ args_concat }})';
--- a/packages/cubejs-schema-compiler/test/integration/clickhouse/complex-joins.test.ts
+++ b/packages/cubejs-schema-compiler/test/integration/clickhouse/complex-joins.test.ts
@@ -1,0 +1,210 @@
+import { prepareJsCompiler } from '../../unit/PrepareCompiler';
+import { ClickHouseDbRunner } from './ClickHouseDbRunner';
+
+describe('ClickHouse complex joins', () => {
+  jest.setTimeout(20 * 1000);
+
+  const dbRunner = new ClickHouseDbRunner();
+
+  afterAll(async () => {
+    await dbRunner.tearDown();
+  });
+
+  const { compiler, joinGraph, cubeEvaluator } = prepareJsCompiler(`
+    cube(\`Acube\`, {
+      sql: \`SELECT * FROM (SELECT 1 as id, 'Category A' as category, 10 as value UNION ALL
+           SELECT 2, 'Category B', 20 UNION ALL
+           SELECT 3, 'Category C', 30) as t\`,
+
+      joins: {
+        Bcube: {
+          relationship: \`one_to_many\`,
+          sql: \`\${Acube}.id = \${Bcube}.id\`
+        },
+        Ccube: {
+          relationship: \`one_to_many\`,
+          sql: \`\${Acube}.id = \${Ccube}.id\`
+        },
+      },
+
+      measures: {
+        AcubeCount: {
+          type: \`count\`,
+        },
+        AcubeTotalValue: {
+          type: \`sum\`,
+          sql: \`value\`
+        },
+        BcubeTotalValue: {
+          sql: \`\${Bcube.totalValue}\`,
+          type: \`number\`
+        },
+        CcubeTotalAmount: {
+          sql: \`\${Ccube.totalAmount}\`,
+          type: \`number\`
+        },
+      },
+
+      dimensions: {
+        id: {
+          sql: \`id\`,
+          type: \`number\`,
+          primaryKey: true,
+          shown: true,
+          title: \`id\`
+        },
+        category: {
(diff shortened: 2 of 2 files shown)


packages/cubejs-schema-compiler/src/adapter/ClickHouseQuery.ts
```diff
@@ -266,6 +266,17 @@ export class ClickHouseQuery extends BaseQuery {
     return `ALTER TABLE ${tableName} ADD INDEX ${indexName} (${escapedColumns.join(', ')}) TYPE minmax GRANULARITY 1`;
   }
 
+  public dimensionColumns(cubeAlias) {
+    // For the top-level SELECT statement, explicitly set the column alias.
+    // Clickhouse sometimes includes the "q_0" prefix in the column name, and this
+    // leads to errors during the result mapping.
+    if (cubeAlias === 'q_0') {
+      return this.dimensionAliasNames().map(alias => `${cubeAlias && `${cubeAlias}.` || ''}${alias} ${alias}`);
```
KSDaemon, line 274: I think checking for `cubeAlias` inside template string (`cubeAlias && `${cubeAlias}.` || ''`) is not necessary, as you can get here only if `cubeAlias === 'q_0'`
```suggestion
      return this.dimensionAliasNames().map(alias => `${cubeAlias}.${alias} ${alias}`);
```

KSDaemon: I'm trying to get what the difference between this and impl in baseQuery... Okay, I see.. You add another plain alias without cubeAlias... What if there are more query members from other subqueries with the same plain alias? It seems that it leads to ambiguity, and the query will fail... Nope?

dragosrep: Thank you for the analysis.

The committed fix should ensure that the result set mapping rules generated by `aliasNameToMember` (`BaseQuery.js`) can be correctly applied in `get_members` (`query_result_transform.rs`) when the Clickhouse backend is being used. The mapping rules never include the table alias, and given how the Clickhouse server sometimes includes the table name in the column's name in the result set, we updated the `dimensionColumns` function to force an alias for dimension columns when called from `outerMeasuresJoinFullKeyQueryAggregate` (the `q_0` condition). 

Complex queries that end up triggering this issue in Clickhouse are generated by `outerMeasuresJoinFullKeyQueryAggregate`. (`BaseQuery.js`) In the top-level "SELECT" statement all measure columns have an explicit alias, and dimension columns don't. In this case we can't end up with a column name conflict.

PR #9713, opened 2025-06-25
fix: Possible race condition on Orchestrator API initialization

ovr:


packages/cubejs-server-core/src/core/server.ts
```diff
@@ -569,88 +569,49 @@ export class CubejsServerCore {
...
-    const driverPromise: Record<string, Promise<BaseDriver>> = {};
-
-    let externalPreAggregationsDriverPromise: Promise<BaseDriver> | null = null;
-
-    const contextToDbType: DbTypeAsyncFn = this.contextToDbType.bind(this);
-    const externalDbType = this.contextToExternalDbType(context);
-
-    // orchestrator options can be empty, if user didn't define it.
-    // so we are adding default and configuring queues concurrency.
-    const orchestratorOptions =
```
ovr, line 590: As you can see, there is an await point here at `(await this.orchestratorOptions(context)) || {},`. With this, it's possible to have two orchestrator API instances, if it's called immediately at one time.

KSDaemon (approved):
👍🏻 LGTM!

PR #9717, opened 2025-06-26
fix(native): Handle null correctly as JsNull, not as JsString

ovr:


KSDaemon (approved):
👍🏻 LGTM just fix cargo fmt...

PR #9762, opened 2025-07-08
feat(tesseract): Rolling window with custom granularities without date range

waralexrom:
**Check List**


rust/cubesqlplanner/cubesqlplanner/src/planner/sql_templates/plan.rs
```diff
@@ -244,12 +244,8 @@ impl PlanSqlTemplates {
     }
 
     pub fn cast_to_string(&self, expr: &str) -> Result<String, CubeError> {
-        self.render.render_template(
-            "expressions/cast_to_string",
```
KSDaemon, line 248: If it's not needed anymore - let's remove it from the dialect Query classes templates too

KSDaemon (approved):
👍🏻 LGTM!

PR #9766, opened 2025-07-09
fix(cubestore): Make projection_above_limit optimization behave deter…

srh:
…ministically

This orders columns in plan generation based on the order they're seen instead of using hash table ordering.  Note that this affects internal plan nodes and does not change the output of any correctly-running queries.  This has the effect of making query behavior deterministic and reproducible when investigating other bugs in query evaluation.

**Check List**

### DIFF
--- a/rust/cubestore/cubestore/src/queryplanner/projection_above_limit.rs
+++ b/rust/cubestore/cubestore/src/queryplanner/projection_above_limit.rs
@@ -37,7 +37,7 @@ fn projection_above_limit(plan: &LogicalPlan) -> Result<LogicalPlan> {
         LogicalPlan::Limit { n, input } => {
             let schema: &Arc<DFSchema> = input.schema();
 
-            let lift_up_result = lift_up_expensive_projections(input, HashSet::new());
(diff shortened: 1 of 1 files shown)


rust/cubestore/cubestore/src/queryplanner/projection_above_limit.rs
```diff
@@ -107,15 +107,38 @@ fn projection_above_limit(plan: &LogicalPlan) -> Result<LogicalPlan> {
...
+
+impl ColumnCollector for Vec<Column> {
+    fn push(&mut self, column: &Column) {
+        self.push(column.clone());
+    }
+}
+
+#[derive(Default)]
+struct ColumnRecorder<T: ColumnCollector> {
+    column_hash: HashSet<Column>,
```
ovr, line 127: @srh, what do you think about using [`indexmap`](https://github.com/indexmap-rs/indexmap)?

waralexrom (approved):

rust/cubestore/cubestore/src/queryplanner/projection_above_limit.rs
srh: Sure.

PR #9776, opened 2025-07-14
feat(cubesql): Push Limit-Sort down Projection

MazterQyou:
**Check List**

This PR adds Limit-Sort to Projection push down rule which helps queries that could be planned as CubeScan but contain complex Projections that cannot be pushed down, and as such get planned using SQL push down.
It also fixes an issue with Limit-Sort combinations not being prioritized lower than SQL push down when Limit-Sort is not at the top of the plan.

Related test is included.


KSDaemon (approved):
👍🏻 Cool!

PR #9787, opened 2025-07-16
fix(cubesql): Fix sort push down projection on complex expressions

MazterQyou:
**Check List**

This PR workarounds an issue with recently introduced Limit-Sort push down Projection rules, throwing an error when sort column has been replaced with a complex expression and SQL push down is involved. This is a bug in SQL push down generation, so for now only expressions that resolve to columns are being pushed down.


KSDaemon (approved):
👍🏻

PR #9800, opened 2025-07-21
chore(schema-compiler): Make cube joins to be an array instead of hashmap

KSDaemon:
- **more types in CubeSymbols**
- **change joins schema in Validator**
- **specify type for accessPolicy**
- **make prepareJoins() aware of arrays**
- **make joins in CubeSymbols / CubeEvaluator as array instead of hashmap**
- **update JoinGraph to treat joins as array instead of hashmap**
- **fix CubePropContextTranspiler to convert joins hashmap into array**
- **add @types/js-yaml**
- **make YamlCompiler aware of joins as array**
- **fix/update snapshots for schema tests**
- **fix transform yaml joins**
- **add schema tests for joins as array/object**

**Check List**

### DIFF
--- a/packages/cubejs-schema-compiler/src/compiler/transpilers/CubePropContextTranspiler.ts
+++ b/packages/cubejs-schema-compiler/src/compiler/transpilers/CubePropContextTranspiler.ts
@@ -100,6 +100,13 @@ export class CubePropContextTranspiler implements TranspilerInterface {
 
     return {
       ObjectProperty: (path) => {
+        if (path.node.key.type === 'Identifier' && path.node.key.name === 'joins' && t.isObjectExpression(path.node.value)) {
+          const fullPath = this.fullPath(path);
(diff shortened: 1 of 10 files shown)


packages/cubejs-schema-compiler/src/compiler/transpilers/CubePropContextTranspiler.ts
```diff
@@ -100,6 +100,13 @@ export class CubePropContextTranspiler implements TranspilerInterface {
 
     return {
       ObjectProperty: (path) => {
+        if (path.node.key.type === 'Identifier' && path.node.key.name === 'joins' && t.isObjectExpression(path.node.value)) {
```
ovr, line 103: I am considering the case when some defined `joins` dynamically with `asyncModule`. It will be a breaking change for him. Theoretically, it can be done this way, but I don't know if this is used in practice.

PR #9809, opened 2025-07-23
fix(query-orchestrator): Reduce number of refresh key queries

ovr:
Previously, Cube generated a burst load of queries for refresh keys. We don't use a queue for cube store queries, but we use in memory cache inside `QueryQueue`. Cache miss leads to a problem that we start executing queries without debouncing.

Cube loads/refresh refresh keys in parallel, that DDoS-ing Cube Store.

To demonstrate that, I've added `console.log` in `asyncDebounce`, just to showcase it:

image

### DIFF
--- a/packages/cubejs-query-orchestrator/src/orchestrator/QueryCache.ts
+++ b/packages/cubejs-query-orchestrator/src/orchestrator/QueryCache.ts
@@ -2,7 +2,7 @@ import crypto from 'crypto';
 import csvWriter from 'csv-write-stream';
 import { LRUCache } from 'lru-cache';
 import { pipeline } from 'stream';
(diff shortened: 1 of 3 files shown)


packages/cubejs-backend-shared/src/promises.ts
```diff
@@ -265,7 +265,7 @@ export const retryWithTimeout = <T>(
   );
 
 /**
- * High order function that makes to debounce multi async calls to single call at one time
+ * Debunks multiple asynchronous calls at once.
```
KSDaemon, line 268: I believe you mean `debounce` and not `debunks` :) 

```suggestion
 * Creates a debounced version of an asynchronous function.
```

PR #9815, opened 2025-07-24
refactor: Remove deprecated queue events bus functionality

ovr:
The queue event bus was deprecated and is no longer needed.

This completes the removal of the deprecated queue events bus system across the entire codebase.

https://www.youtube.com/watch?v=A8Fnmy05xqQ


.gitignore
```diff
@@ -24,3 +24,4 @@ rust/cubesql/profile.json
 .cubestore
 .env
 .vimspector.json
+.claude/settings.local.json
```
KSDaemon, line 27: :) Should this be just the directory `.claude`?

KSDaemon (approved):
👍🏻

.gitignore
ovr: @KSDaemon not, because it can contain commands. For example, @igorlukanin defined 1 in `docs`.

https://github.com/cube-js/cube/tree/master/docs/.claude/commands

KSDaemon: Got it! thnx!)

Issue #7969, opened 2024-03-20
API endpoints return 404 when basePath is set to '/'


slatkovic:
**Describe the bug**
Setting `basePath` in cube.js to `/` results in 404 across all API endpoints

**To Reproduce**
1. Create a cube.js fils
2. Set `basePath: '/'`
3. Try accessing the API at [base host]/v1/meta
4. Receive a 404

**Expected behavior**
The API should be accessible at [base host]/v1/meta

**Version:**
v0.35.1

**Additional context**
Setting the `basePath` to a different value, e.g. `/test` results in proper behavior ([base host]/test/v1/meta returns 200)

igorlukanin:
I was able to confirm this behavior.

@slatkovic Could you please describe your use case? Why would you like to have an empty base path?

BTW, as a workaround, you can (and probably should) [put Cube behind a reverse proxy](https://cube.dev/docs/product/deployment/core#set-up-reverse-proxy) if you're self-hosting.

PR #7816, opened 2024-02-26
feat(playground): vite build uses relative asset paths instead of absolute

benswinburne:
When attempting to run Cube behind a reverse proxy, it is only possible to set the base path for the API endpoints of Cube. This renders the playground unusable inside a non root base path.

For example when using /cube as a base path behind a reverse proxy, I can get the index.html to load, from /cube/index.html but then requests for assets are made to /some-asset.js instead of /cube/some-asset.js

**Check List**
- [x] Tests has been run in packages where changes made if available
- [x] Linter has been run for changed code
- [ ] Tests for the changes have been added if not covered yet
- [ ] Docs have been added / updated if required

**Description of Changes Made (if issue reference is not provided)**

Changes the path of bundled assets from absolute to relative. 

For example

    <script type="module" crossorigin src="/assets/index-4bed1a0c.js"></script>
 
Becomes

    <script type="module" crossorigin src="./assets/index-4bed1a0c.js"></script>

This allows Cube playground to be run inside a subdirectory (behind a reverse proxy, for example).

Adding `base: './'` to the Vite config is the only change required for this.

https://vitejs.dev/config/shared-options.html#base

As you can see in the docs, it's suggested for "embedded" applications, which I suppose encompasses those bundled in  containers such as Cube's playground.

In terms of backwards compatibility, it should be fine as currently you must run it on the root path, so applications running on the root path will go from `/asset...` to `./asset...` which resolves to `/asset...`.

Running `yarn build` in the  packages/cubejs-playground directory yields the following

```diff
diff --git a/build/index.html b/index-built.html
index 3b100ab4e..07b9dd75b 100644
--- a/build/index.html
+++ b/relative-build/index.html
@@ -2,17 +2,17 @@
 <html lang="en">
   <head>
     <meta charset="utf-8" />
-    <link rel="apple-touch-icon" sizes="180x180" href="/apple-touch-icon.png">
-    <link rel="icon" type="image/png" href="/favicon-32x32.png" sizes="32x32">
-    <link rel="icon" type="image/png" href="/favicon-16x16.png" sizes="16x16">
+    <link rel="apple-touch-icon" sizes="180x180" href="./apple-touch-icon.png">
+    <link rel="icon" type="image/png" href="./favicon-32x32.png" sizes="32x32">
+    <link rel="icon" type="image/png" href="./favicon-16x16.png" sizes="16x16">
     <link rel="mask-icon" href="/safari-pinned-tab.svg" color="#5bbad5">
[...]

Issue #3028, opened 2021-06-29
Allow Users to set URL Path Prefix for Developer Playground


Alex-Yakubovsky:
**Is your feature request related to a problem? Please describe.**
During local development we have many services running on a local k8s cluster. To prevent URL clashing, each service has its own URL prefix (e.g, `http://1.2.3.4/api/...`, `http://1.2.3.4/frontend-1/...`). I'm prefixing `cubejs` api calls with `http://1.2.3.4/cubejs/...` but the developer playground frontend has no ways to add a prefix to the api calls it makes.

**Describe the solution you'd like**
An environment variable like `CUBEJS_PLAYGROUND_URL_PREFIX` that can be set. 

**Describe alternatives you've considered**
Temporarily allowing `http://1.2.3.4` to point to cubejs. It's rather cumbersome since I have to comment a bunch of code and then remember to uncomment before finishing a feature I'm working on.

**Additional context**

nacmonad:
Is there an alternative way to do this?  
Running into reverse-proxy issue as a result of not being able to set a url path prefix.  

Best

diogosilva30:
I'm unable to deploy cubejs playground behind nginx reverse proxy. My nginx:
```
location /cubeplayground/ {
    proxy_pass          http://cube:4000/;
    proxy_set_header    Host $host;
    proxy_set_header    X-Real-IP $remote_addr;
    proxy_set_header    X-Forwarded-For $proxy_add_x_forwarded_for;
    proxy_set_header    X-Forwarded-Proto https;
    proxy_read_timeout  600s;
}
location /cube/cubejs-api/ {
    proxy_pass          http://cube:4000/cubejs-api/;
    proxy_set_header    Host $host;
    proxy_set_header    X-Real-IP $remote_addr;
    proxy_set_header    X-Forwarded-For $proxy_add_x_forwarded_for;
    proxy_set_header    X-Forwarded-Proto https;
    proxy_read_timeout  600s;
}
```
The API is working fine. However, the playground just shows a blank white page, with the console error:
```
Failed to load module script: Expected a JavaScript module script but the server responded with a MIME type of "text/html". Strict MIME type checking is enforced for module scripts per HTML spec.
```
Any idea how to make this work? I have "dev" mode enabled

ommore1523:
> I'm unable to deploy cubejs playground behind nginx reverse proxy. My nginx:
> 
> ```
> location /cubeplayground/ {
[...]

PR #9850, opened 2025-08-04
feat(playground): Use relative paths for assets, playground requests and fonts

dacosta444:
Based on the work in PR https://github.com/cube-js/cube/pull/7816

"When attempting to run Cube behind a reverse proxy, it is only possible to set the base path for the API endpoints of Cube. This renders the playground unusable inside a non root base path.

For example when using /cube as a base path behind a reverse proxy, I can get the index.html to load, from /cube/index.html but then requests for assets are made to /some-asset.js instead of /cube/some-asset.js"

In addition to the changes made in that PR the requests to playground were still using the absolute path. 
Fonts were also using an absolute path and the fix for that was forcing a relative path.

**Check List**

#3028 #7969

### DIFF
--- a/packages/cubejs-playground/src/App.tsx
+++ b/packages/cubejs-playground/src/App.tsx
@@ -63,7 +63,7 @@ class App extends Component<PropsWithChildren<RouteComponentProps>, AppState> {
       });
     });
 
-    const res = await fetch('/playground/context');
+    const res = await fetch('playground/context');
     const context = await res.json();
 
     setTelemetry(context.telemetry);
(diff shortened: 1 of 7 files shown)


dacosta444:
Would it be possible to trigger the action to build and publish a Docker image for this PR so I can use it while the review is still pending?

It would be super helpful to have a published image using the GitHub Actions flow — something similar to the [build and publish step](https://github.com/cube-js/cube/actions/runs/16756280506/job/47440252343)
cc @vasilev-alex, @ovr 
Thanks a lot in advance!

tenphi:
@dacosta444  Sorry for the long wait.

PR #9869, opened 2025-08-11
fix(client-core): Remove default value in table pivot

Emile-Filteau:
**Check List**

Back in May when release 1.3.6 came out, [the client package was migrated to typescript](https://github.com/cube-js/cube/pull/9562). While trying to upgrade from previous versions to the newer releases, i ended up seeing discrepancies when running tests against some cube models. 

After some digging, it seems like the default value was added during the migration process. The previous behaviour being that if you are joining on a table that does not have a matching row for your model, the requested columns were `null` and not `""`.

I'm wondering if that behaviour is expected, and if so why?


KSDaemon (approved):
👍🏻 LGTM!

PR #9881, opened 2025-08-14
fix(api-gateway): Handle array format for joins in /meta?extended endpoint

igorlukanin:
After PR #9800, joins are stored as arrays instead of objects in the schema compiler. The transformJoins function in the API gateway was still using Object.entries() which, when applied to arrays, returns numeric indices as keys instead of actual join names.

This fix:
- Checks if joins is an array and preserves the name property
- Maintains backward compatibility with object format
- Adds test coverage for both array and object formats

Fixes the issue where join names appear as "0", "1", "2" instead of their actual names in the /meta?extended endpoint response.

**Check List**


KSDaemon (approved):
👍🏻 LGTM! thnx for fixing!

Issue #9566, opened 2025-05-13
pipeline function in QueryCache fails with ECONNRESET for lambda rollup pre-aggregations


viktordebulat:
**Describe the bug**
We are encountering an issue with [pipeline in QueryCache.ts](https://github.com/cube-js/cube/blob/master/packages/cubejs-query-orchestrator/src/orchestrator/QueryCache.ts#L501) when processing lambda rollup pre-aggregations for ClickHouse. Specifically, the error occurs during the streaming of data from `tableData.rowStream` to the `writer`. The error is intermittent but consistently happens for certain request types after approximately 2 seconds:

```
Error: aborted
    at TLSSocket.socketCloseListener (node:_http_client:478:19)
    at TLSSocket.emit (node:events:530:35)
    at node:net:351:12
    at TCP.done (node:_tls_wrap:650:7) {
  code: 'ECONNRESET'
```

The problem goes away when replacing affected code with direct iterator processing:

```
const iterator = tableData.rowStream[Symbol.asyncIterator]();
let result = await iterator.next();
while (!result.done) {
  writer.write(result.value);
  result = await iterator.next();
}
writer.end();
```

This workaround resolves the issue, but it bypasses the pipeline utility, which is designed to handle stream piping and error propagation.

**To Reproduce**
1. Create cube with ClickHouse as datasource, declare rollups and lambda rollups in preaggregations section.
2. Trigger a request that processes a large dataset or involves a lambda running.

**Expected behavior**
The pipeline function should handle the streaming of data without prematurely aborting due to ECONNRESET.

**Minimally reproducible Cube Schema**

```
cubes:
  - name: cube1
    sql: >
      some query here

  - name: cube2
    sql: >
      another query here

  - name: cube_total
    sql: >
      { cube1.sql() }
      UNION ALL
      { cube2.sql() }

    measures:
      - name: total_transactions
        sql: transaction_id
        type: count

      - name: total_amount
        sql: amount
        type: sum

      - name: total_payout
        sql: payout
        type: sum

    dimensions:
      - name: user_id
        sql: user_id
        type: string

      - name: currency
        sql: currency
        type: string

      - name: at
        sql: at
        type: time

    pre_aggregations:
      - name: cube_total_rollup_lambda
        type: rollup_lambda
        union_with_source_data: true
        rollups:
          - CUBE.cube_total_rollup

      - name: cube_total_rollup
        type: rollup
        measures:
          - cube_total.total_transactions
          - cube_total.total_amount
[...]

PR #9891, opened 2025-08-17
fix(query-orchestrator): QueryCache - propagate error later for streams

viktordebulat:
If the source (rowStream) emits an error or closes before all data is read, pipeline will reject.

This manually consumes the async iterator from the stream, catching errors at each step. If the stream ends early or throws, you catch the error and can handle it gracefully. This avoids some of the internal state issues that can occur with pipeline when the source stream is an async generator that may throw.

#9566


viktordebulat:
Hi @igorlukanin, i'm not sure that this approach is the best way. Could you summon someone to review this?
I've been using this for 3 month in my project (ClickHouse rollups with lambdas) and would like to add this fix to the upstream

Issue #10010, opened 2025-09-29
snowflake-driver: Support IAM roles with IRSA for S3 export buckets


morford-brex:
**Is your feature request related to a problem? Please describe.**
Currently, the Snowflake driver requires explicit AWS credentials (`CUBEJS_DB_EXPORT_BUCKET_AWS_KEY` and `CUBEJS_DB_EXPORT_BUCKET_AWS_SECRET`) for S3 export buckets, even when using IAM roles. This creates a problem in cloud-native environments (e.g. k8s with IRSA).

**Describe the solution you'd like**
When no explicit credentials or storage integration is provided, the driver should:
- Skip credential validation for `keyId`/`secretKey`
- Omit the `CREDENTIALS` clause from Snowflake `COPY INTO` commands
- Let the AWS SDK use the default credential chain (IRSA → instance profile → environment)

**Describe alternatives you've considered**
The current alternative of using hardcoded credentials requires additional setup and doesn't leverage native cloud IAM capabilities.

**Additional context**
The AWS SDK already supports this pattern out-of-the-box - we just need the Snowflake driver to allow it.

PR #10011, opened 2025-09-29
fix(snowflake-driver): Add support for IAM roles with IRSA for S3 export buckets

morford-brex:
**Check List**

Fixes https://github.com/cube-js/cube/issues/10010


morford-brex:
@KSDaemon - apologies i missed adding the new driver tests to the relevant worklow config

docs/pages/product/configuration/data-sources/snowflake.mdx
```diff
@@ -134,10 +134,11 @@ Storage][google-cloud-storage] for export bucket functionality.
...
 
 Ensure the AWS credentials are correctly configured in IAM to allow reads and
-writes to the export bucket in S3 if you are not using storage integration.
-If you are using storage integration then you still need to configure access keys
-for Cube Store to be able to read from the export bucket.
-It's possible to authenticate with IAM roles instead of access keys for Cube Store.
+writes to the export bucket in S3. You can authenticate using:
+
+- **IAM user credentials** - Explicit AWS access keys and secrets
+- **Storage integration** - Snowflake-managed AWS integration (may still require credentials for Cube Store)
```
KSDaemon, line 140: > (may still require credentials for Cube Store)

This is a bit incorrect. Cube Store doesn't use credentials at all. It operates only with the signed URLs. The credentials might be required by the Cube.js server to be able to sign the URLs for exported CSVs. 

So I suggest rephrasing it to something about that Cube might still require credentials for exported files post-processing... or smth like that (without much tech details here).

packages/cubejs-snowflake-driver/src/SnowflakeDriver.ts
```diff
@@ -771,14 +793,18 @@ export class SnowflakeDriver extends BaseDriver implements DriverInterface {
       const { bucketName, path } = this.parseBucketUrl(this.config.exportBucket!.bucketName);
       const exportPrefix = path ? `${path}/${tableName}` : tableName;
 
+      const s3Config: any = { region };
```
KSDaemon, line 796: Don't you mind changing `any` to something more explicit? Trying to avoid it when possible.

KSDaemon (commented):
Looks good! Left minor comments.

docs/pages/product/configuration/data-sources/snowflake.mdx
morford-brex: updated in https://github.com/cube-js/cube/pull/10011/commits/fad29b1bc0519ff121b66bf397fa1252952fa493

packages/cubejs-snowflake-driver/src/SnowflakeDriver.ts
morford-brex: ah good catch - updated to use the proper aws type here https://github.com/cube-js/cube/pull/10011/commits/fad29b1bc0519ff121b66bf397fa1252952fa493

docs/pages/product/configuration/data-sources/snowflake.mdx
morford-brex: i actually decided to revert the docs changes as i realized the current state is sufficient

we'll still need to use snowflake integration here for the write side so ultimately the change here is just removing the broken logic to enable read side iam usage when the key/secret params are excluded which is what the docs suggest as-is

KSDaemon: Great! thnx! I think it is better now!

morford-brex: sweet - much appreciated! should be good to go from my end whenever you're ready to merge

dmeshkov-brex:
@KSDaemon possible to release this PR today/tomorrow? Thanks!

PR #10014, opened 2025-09-30
fix(schema-compiler): Do not collect disabled external pre-aggregations

MazterQyou:
**Check List**

This PR resolves an issue with SQL push down queries having pre-aggregation description which let it bypass rollup-only mode and get registered as a pre-aggregated query.


KSDaemon (approved):
👍🏻

PR #10031, opened 2025-10-06
fix(sqlite-driver) Update tablesSchema

LFischerstrom:
**Check List**

The current implementation of `tableSchema` doesn't pass the tests added in this PR.

To pass the tests (and improve on readability) we can use `pragma_table_info(<tableName>)` to list the column names and their type. This results in N+1 queries but this is not an issue with SQLite https://www.sqlite.org/np1queryprob.html

### DIFF
--- a/packages/cubejs-sqlite-driver/package.json
+++ b/packages/cubejs-sqlite-driver/package.json
@@ -15,7 +15,7 @@
   "typings": "driver/index.d.ts",
   "scripts": {
     "lint": "eslint **/*.js",
-    "unit": "jest ./**/*.test.js"
+    "unit": "jest"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


packages/cubejs-sqlite-driver/package.json
```diff
@@ -14,7 +14,8 @@
   "main": "driver/SqliteDriver.js",
   "typings": "driver/index.d.ts",
   "scripts": {
-    "lint": "eslint **/*.js"
+    "lint": "eslint **/*.js",
+    "test": "jest ./**/*.test.js"
```
KSDaemon, line 18: Let's rename (or just add an alias) `unit` so this test could run in CI unit tests with `lerna run unit` - that would be useful.

PR #10066, opened 2025-10-19
fix(schema-compiler): add support for time filters and rolling windows and fix subquery aliasing in Oracle dialect

Tankilevitch:
# Fix Oracle Query Generation Issues

## Summary

This PR fixes three critical Oracle database compatibility issues that prevented queries with rolling windows, time dimension filters, and subqueries from executing. All fixes maintain backward compatibility with other database adapters.

---

### Problem
Oracle does not support the `AS` keyword before table/subquery aliases. Queries requiring subqueries would fail with:
```
ORA-00933: SQL command not properly ended
```

### Why It Happened
The query builder was hardcoding the `as` keyword when generating subquery aliases, which works for PostgreSQL/MySQL but is invalid in Oracle SQL syntax.

### What We Did
Modified the query builder to use a database-specific property (`asSyntaxJoin`) that each adapter can override. Oracle sets this to an empty string, while other databases use `"AS"`.

### Example

**PostgreSQL (unchanged):**
```sql
FROM (SELECT ...) AS q_0
INNER JOIN (SELECT ...) AS q_1 ON ...
```

**Oracle (fixed):**
```sql
FROM (SELECT ...) q_0
INNER JOIN (SELECT ...) q_1 ON ...
```

---

### Problem
When using time dimensions for filtering only (without specifying granularity), queries would crash with:
```
TypeError: Cannot read properties of undefined (reading 'isNaturalAligned')
```

### Example Query
```json
{
  "measures": ["visitors.count"],
  "timeDimensions": [{
    "dimension": "visitors.createdAt",
    "dateRange": ["2020-01-01", "2020-12-31"]
    // No granularity - just filtering
  }]
}
```

### Why It Happened
The code assumed granularity was always present and tried to access properties on an undefined object. Time dimensions used only for filtering don't require granularity.

### What We Did
Added a null check before accessing granularity properties. When granularity is not specified, the dimension is used as-is for filtering without grouping.

### Result
```sql
-- Query now generates correctly:
SELECT count(*) FROM visitors
WHERE created_at >= to_date(...) AND created_at <= to_date(...)
-- No GROUP BY (correct for filtering without granularity)
```

---

### Problem
Queries with rolling windows would fail with:
```
ORA-17041: Missing IN or OUT parameter at index: 1
```

### Example Query
```json
{
  "measures": ["visitors.unboundedCount"],
  "timeDimensions": [{
    "dimension": "visitors.createdAt",
    "granularity": "year",
    "dateRange": ["2020-01-01", "2022-12-31"]
  }]
}
```
*(where `unboundedCount` has a `rollingWindow` configuration)*

### Why It Happened
The default date arithmetic used PostgreSQL-style interval syntax:
```sql
-- PostgreSQL syntax (doesn't work in Oracle)
date_field >= some_date - interval '1 year'
```

Oracle requires specific functions for date arithmetic instead of the `INTERVAL` keyword.

### What We Did
Implemented Oracle-specific `addInterval` and `subtractInterval` methods that use:
- **`ADD_MONTHS(date, n)`** for year/month/quarter intervals
- **`NUMTODSINTERVAL(n, unit)`** for day/hour/minute/second intervals

### Transformation Examples

**Before (PostgreSQL syntax):**
```sql
WHERE date_field >= to_date(:"?", ...) - interval '1 year'
```

**After (Oracle syntax):**
```sql
WHERE date_field >= ADD_MONTHS(to_date(:"?", ...), -12)
```

### Interval Conversion

| Interval Type | Oracle Function |
|--------------|-----------------|
| Years | `ADD_MONTHS(date, ±12)` |
| Quarters | `ADD_MONTHS(date, ±3)` |
| Months | `ADD_MONTHS(date, ±1)` |
| Days | `date ± NUMTODSINTERVAL(n, 'DAY')` |
| Hours | `date ± NUMTODSINTERVAL(n, 'HOUR')` |

---

### New Test Coverage
Created comprehensive test suite in `oracle-query.test.ts` with 10 tests covering:
- Basic query generation
- Subquery aliases without `AS` keyword (multiple scenarios)
- `FETCH NEXT` syntax instead of `LIMIT`
- Group by dimensions (not indexes)
- **Time dimensions without granularity** ✅
- **Oracle-specific interval arithmetic** ✅

### Regression Protection
Added test in `postgres-query.test.ts` to ensure PostgreSQL continues using `AS` keyword correctly.

---

## Backward Compatibility

✅ **All changes are backward compatible:**
- PostgreSQL, MySQL, BigQuery, etc. continue using `AS` keyword
- The granularity null check works for all database adapters
- Only Oracle uses the new interval arithmetic methods
- All existing queries continue to work as before

---

## Impact

These fixes enable Oracle users to:
- Use time dimensions for filtering without specifying granularity
- Execute queries with rolling windows (trailing, leading, offset)
- Perform time-based comparisons with proper date arithmetic
- Use any query pattern that requires subqueries or date calculations

### DIFF
--- a/packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
+++ b/packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
@@ -3857,11 +3857,6 @@ export class BaseQuery {
    * @return {string}
    */
   dimensionTimeGroupedColumn(dimension, granularity) {
-    // Handle case when granularity is not specified (e.g., time dimension used only for filtering)
-    if (!granularity) {
-      return this.timeGroupedColumn(null, dimension);
-    }
-
     let dtDate;
 
     // Interval is aligned with natural calendar, so we can use DATE_TRUNC
--- a/packages/cubejs-schema-compiler/src/adapter/OracleQuery.ts
+++ b/packages/cubejs-schema-compiler/src/adapter/OracleQuery.ts
@@ -56,7 +56,11 @@ export class OracleQuery extends BaseQuery {
    * using forSelect dimensions for grouping
    */
   public groupByClause() {
-    const dimensions = this.forSelect().filter((item: any) => !!item.dimension) as BaseDimension[];
+    // Only include dimensions that have select columns
+    // Time dimensions without granularity return null from selectColumns()
+    const dimensions = this.forSelect().filter((item: any) => (
+      !!item.dimension && item.selectColumns && item.selectColumns()
+    )) as BaseDimension[];
     if (!dimensions.length) {
       return '';
     }
--- a/packages/cubejs-schema-compiler/test/unit/oracle-query.test.ts
+++ b/packages/cubejs-schema-compiler/test/unit/oracle-query.test.ts
@@ -324,8 +324,36 @@ describe('OracleQuery', () => {
     const queryAndParams = query.buildSqlAndParams();
     const sql = queryAndParams[0];
 
-    // Key test: no GROUP BY on time dimension when granularity is missing
+    // Time dimensions without granularity should not appear in GROUP BY
     expect(sql).not.toMatch(/GROUP BY.*created_at/i);
+  });
+
+  it('handles time dimension with granularity in SELECT and GROUP BY', async () => {
+    await compiler.compile();
+
+    const query = new OracleQuery({ joinGraph, cubeEvaluator, compiler }, {
+      measures: [
+        'visitors.count'
+      ],
+      timeDimensions: [{
+        dimension: 'visitors.createdAt',
+        granularity: 'day',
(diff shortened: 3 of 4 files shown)


packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
```diff
@@ -3857,6 +3857,11 @@ export class BaseQuery {
    * @return {string}
    */
   dimensionTimeGroupedColumn(dimension, granularity) {
+    // Handle case when granularity is not specified (e.g., time dimension used only for filtering)
+    if (!granularity) {
```
KSDaemon: Can you show the example when this function is called without `granularity` object?

KSDaemon (approved):
👍🏻  LGTM!

PR #10070, opened 2025-10-20
feat(cubesql): Filter push down for date_part(?upper) AND date_part(?lower)

ovr:
Allowing filters to be pushed down with `DATE_PART('year', ?column) = ?year AND DATE_PART('quarter', ?column) = ?quarter` by pushing it as inDateRange. Now, such queries will be accelerated with pre-aggregations.

I am re-using code that we have for tablea, but without usage of `Trunc` as wrapper. I've added ability to merge date range with `quarter`.

### DIFF
--- a/rust/cubesql/cubesql/src/compile/rewrite/rules/filters.rs
+++ b/rust/cubesql/cubesql/src/compile/rewrite/rules/filters.rs
@@ -4057,8 +4057,12 @@ impl FilterRules {
                         return false;
                     }
 
+                    // Preserves existing constraints, for example:
+                    // inDataRange: order_date >= '2019-02-15' AND order_date < '2019-03-10'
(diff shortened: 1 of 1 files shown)


rust/cubesql/cubesql/src/compile/rewrite/rules/filters.rs
```diff
@@ -4022,6 +4064,48 @@ impl FilterRules {
...
+                    };
+
+                    let Some(new_end_date) = new_start_date
+                        .checked_add_months(Months::new(3))
+                        .and_then(|date| date.checked_sub_days(Days::new(1)))
+                    else {
+                        return false;
+                    };
+
+                    // Paranoid check, If the resulting range is outside of the original range, we can't merge
```
ovr: Someone can combine timestamp comparision and quarter. Let's add these check to protect such cases.

PR #10086, opened 2025-10-27
fix(cube-orchestrator): Deterministic column sequence in output

### LINKED ISSUE #10080: Random data column sequence in API response (endpoint: /v1/load)
**Problem**

Cube.dev load api call is returning a response, but the sequence of the columns changes for every request. It is a challenge in case of pagination requests in compact mode, as the data array may have a different column sequence for every page, which has to be realigned to combine multiple pages together. A similar issue is observed in the case of default responseFormat, where the property fields in every single JSON object in the result have a different sequence. I remember some older version(maybe before v0.36.x) didn't have this issue. So, the question is whether this is a regression or intended behaviour. 

**Expected Behaviour**
In my understanding, the expected behaviour is to have the columns sequence in the same order as the Time dimension first, followed by the dimensions in the provided sequence, followed by the measures in their provided sequence. 

**Probable Root Cause**
I have had a brief look into the code and believe the behaviour is happening because of using HashMap instead of IndexMap in Rust (rust/cubeorchestrator/src/query_result_transform.rs). [IndexMap](https://docs.rs/indexmap/latest/indexmap/) is drop-in compatible with HashMap, so do you think this can be a way to add deterministic behaviour to cube.dev responses? Happy to contribute a PR if you think this is the right approach.

**Related Cube.js schema**
Used the Demo deployment. I can share a simple schema as well if that helps.

**Screenshots:**
Default Mode:
```json
{
    "query": {
        "dimensions": [
            "orders.users_city",
            "orders.users_state",
            "orders.users_age",
            "orders.status"
        ],
        "measures": [
            "orders.total",
            "orders.count",
            "orders.completed_count",
            "orders.completed_percentage"
        ],
        "limit": 10
    }
}
```

1st Run:
Image

2nd Run:
Image

Compact mode:
```json
{
    "query": {
        "responseFormat": "compact",
        "dimensions": [
            "orders.users_city",
            "orders.users_state",
            "orders.users_age",
            "orders.status"
        ],
        "measures": [
            "orders.total",
            "orders.count",
            "orders.completed_count",
            "orders.completed_percentage"
        ],
        "limit": 10
    }
}
```

1st Run:
Image

2nd Run:
Image

UsmanYasin:
**Check List**

closes #10080 

Replaced HashMap with IndexMap, which ensures deterministic order. The idea is to maintain the column order consistency based on the sequence in the API request.


KSDaemon (approved):
👍🏻 LGTM!

ovr (approved):

PR #10110, opened 2025-11-03
fix(clickhouse-driver): Respect timezone for origin in custom granula…

ovr:
## Problem

When using custom granularities with an `origin` timestamp, ClickHouse's `dateBin` implementation was not respecting time zones properly:

ClickHouse aligns both timestamps internally to the same timezone before computing the difference, causing an unintended offset.  As a result, the date difference and the final rounded time are incorrect, and the returned values are effectively in UTC.

### DIFF
--- a/packages/cubejs-schema-compiler/test/integration/mssql/mssql-pre-aggregations.test.ts
+++ b/packages/cubejs-schema-compiler/test/integration/mssql/mssql-pre-aggregations.test.ts
@@ -270,7 +270,7 @@ describe('MSSqlPreAggregations', () => {
 
       expect(preAggregationsDescription[0].invalidateKeyQueries[0][0].replace(/(\r\n|\n|\r)/gm, '')
         .replace(/\s+/g, ' '))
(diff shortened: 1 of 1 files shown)


packages/cubejs-schema-compiler/src/adapter/ClickHouseQuery.ts
```diff
@@ -105,11 +111,17 @@ export class ClickHouseQuery extends BaseQuery {
     return this.dateTimeCast(value);
   }
 
-  public dateTimeCast(value: string): string {
+  public dateTimeCast(value: string, timezone?: string): string {
+    // If a timezone is specified, use toDateTime64 to parse the string AS IF it's in that timezone
+    // This is critical for custom granularity, because timezone should be aligned between origin and source column
+    if (timezone) {
+      // Use precision 3 for milliseconds to match the format 'YYYY-MM-DDTHH:mm:ss.SSS'
+      return `toDateTime64(${value}, 3, '${timezone}')`;
```
ovr: We prefer to cast datetime on the source database side.

waralexrom (approved):

waralexrom (approved):

PR #10192, opened 2025-11-27
fix(query-orchestrator): Table used cache should not be used for unions

ovr:
We don't need to cache them or mark as used in cache store, because it's single tables. Caching such keys result in huge memory usage, we can call it as memory leak.


KSDaemon (approved):
👍🏻 Cool!

PR #10206, opened 2025-12-01
feat(backend-native): expose lastRefreshTime via /cubesql API

KSDaemon:
- **add lastRefreshTime to load result in openapi spec**
- **feat(backend-native): expose lastRefreshTime via /cubesql API**

**Check List**

### DIFF
--- a/rust/cubesql/cubesql/src/transport/service.rs
+++ b/rust/cubesql/cubesql/src/transport/service.rs
@@ -113,12 +113,6 @@ impl SpanId {
     }
 }
 
-#[derive(Clone)]
-pub struct TransportServiceLoadResponse {
-    pub last_refresh_time: Option<String>,
-    pub results_batch: RecordBatch,
-}
-
 #[async_trait]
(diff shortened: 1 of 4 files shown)


rust/cubesql/cubesql/src/transport/service.rs
```diff
@@ -113,6 +113,12 @@ impl SpanId {
     }
 }
 
+#[derive(Clone)]
+pub struct TransportServiceLoadResponse {
```
ovr: At least, it's incorrect name for such structure in term of the usage, because it wraps RecordBatch, but result of the `load` method is `Vec<TransportServiceLoadResponse>`. It just missleading, let's name `TransportServiceLoadRecordBatch`

Annother question, what is the reason to wrap RecordBatch instead of storing `last_refresh_time` at top level?

```
TransportServiceLoadResponse {
     pub last_refresh_time: Option<String>,
     pub batches: Vec<RecordBatch>,
}
```

PR #10262, opened 2025-12-16
fix(databricks-driver): exclude partition information metadata

blturan:
**Check List**

N/A

Databricks `DESCRIBE` output for partitioned tables includes metadata rows where the column name is `# Partition Information` and the `data_type` field is empty ([docs.databricks](https://docs.databricks.com/aws/en/sql/language-manual/sql-ref-syntax-aux-describe-table)). These rows were previously parsed as columns.

This change stops column parsing for the partition metadata section, ensuring that only actual table columns are returned. The check requires both the column name to match `# Partition Information` and*the `data_type` to be empty, as a real table column could legitimately have this name but would always include a non-empty `data_type`.

**Issue Explanation**

Selecting YAML for `Generate Data Model` causes a playground error because the generated YAML becomes invalid when partition metadata rows are included as columns.
image

Playground error.
image

Selecting `JavaScript` does not trigger a playground error, but the generated schema still includes `col_name` and `# Partition Information`, which should not be treated as table columns.
image

### DIFF
--- a/docs/pages/product/embedding/signed-embedding.mdx
+++ b/docs/pages/product/embedding/signed-embedding.mdx
@@ -152,6 +152,37 @@ Here's a complete HTML example that demonstrates the full flow for embedding a d
 </html>
 ```
 
+## Customizing the theme
+
+You can customize the embedded application theme by setting it to `light`, `dark`, or `auto` (follows system preference).
+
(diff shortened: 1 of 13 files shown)


blturan:
Could you please take a look at this PR, @KSDaemon? Thank you.

PR #10264, opened 2025-12-17
feat(schema-compiler): view's cubes multiple join path validation

KSDaemon:
This PR extends CubeValidator with the ability to check if view's included cubes have multiple join paths.

**Check List**
- [x] Tests have been run in packages where changes have been made, if available

### DIFF
--- a/packages/cubejs-schema-compiler/src/compiler/CubeValidator.ts
+++ b/packages/cubejs-schema-compiler/src/compiler/CubeValidator.ts
@@ -1161,20 +1161,28 @@ export class CubeValidator implements CompilerInterface {
   }
 
   /**
-   * Validates that each leaf node (last segment of joinPath) appears only once across all cubes in a view.
-   * This ensures that there is only one path to reach each cube in the view's join graph.
+   * Validates that each cube appears only once within each root path in a view.
+   * This ensures that there is only one path to reach each cube within each root's join graph.
+   * For example: 'a.b.c' + 'd.b.c' is valid (different roots), but 'a.b.c' + 'a.d.c' is invalid (same root 'a').
    */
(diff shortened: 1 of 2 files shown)


packages/cubejs-schema-compiler/src/compiler/CubeValidator.ts
```diff
@@ -1155,6 +1160,45 @@ export class CubeValidator implements CompilerInterface {
     return result;
   }
 
+  /**
+   * Validates that each leaf node (last segment of joinPath) appears only once across all cubes in a view.
+   * This ensures that there is only one path to reach each cube in the view's join graph.
+   */
```
igorlukanin: @KSDaemon Would it catch a case with these join paths?
- `a.b.c`
- `a.x.b`

KSDaemon: yes. Please, have a look at the tests below.

packages/cubejs-schema-compiler/test/unit/cube-validator.test.ts
```diff
@@ -1739,4 +1739,231 @@ describe('Cube Validation', () => {
...
+      const cubeValidator = new CubeValidator(new CubeSymbols());
+      const cube = {
+        name: 'test_view',
+        isView: true,
+        fileName: 'fileName',
+        cubes: [
+          { joinPath: () => 'A.B', prefix: false, includes: ['*'] },
+          { joinPath: () => 'C.D.E.B', prefix: false, includes: ['*'] }
+        ]
+      };
```
igorlukanin: @paveltiunov Should this be actually allowed? Isn't this a multi-fact use case?

ovr (approved):

PR #10286, opened 2026-01-06
feat(api-gateway): Introduce Query format Convertion API

KSDaemon:
This PR introduces new API endpoint `/v1/convert` that allows you to convert queries between different formats. Right now only `sql→rest` conversion is implemented, but this might be extended in the future.

- **feat(api-gateway): Introduce /v1/convert API for query format conversion**
- **prepare backend native**
- **add tests**

**Check List**
- [x] Tests have been run in packages where changes have been made, if available

### DIFF
--- a/packages/cubejs-api-gateway/src/gateway.ts
+++ b/packages/cubejs-api-gateway/src/gateway.ts
@@ -419,7 +419,7 @@ class ApiGateway {
       });
     }));
 
-    app.post(`${this.basePath}/v1/convert`, jsonParser, userMiddlewares, userAsyncHandler(async (req, res) => {
+    app.post(`${this.basePath}/v1/convert-query`, jsonParser, userMiddlewares, userAsyncHandler(async (req, res) => {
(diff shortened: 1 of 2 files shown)


packages/cubejs-backend-native/src/rest4sql.rs
```diff
@@ -0,0 +1,182 @@
...
+
+    let native_auth_ctx = Arc::new(NativeSQLAuthContext {
+        user: Some(String::from("unknown")),
+        superuser: false,
+        security_context: NonDebugInRelease::from(security_context),
+    });
+
+    let (deferred, promise) = cx.promise();
+
+    // In case spawned task panics or gets aborted before settle call it will leave permanently pending Promise in JS land
```
ovr: @claude Could you verify this comment? AFAIK Unhandle promises in Neon are handled with Drop implementation and rejects on JS side with Error. If it's true, please suggest code suggestion to remove/change this comment.

PR #10290, opened 2026-01-07
feat(schema-compiler): support string, time, and boolean measure types in pre-aggregations

### LINKED ISSUE #7896: Using Pre-Aggregation using calculated measures of types: string/time
**Problem**

I have some calculated measures that either aggregate `string values` to an array using `array_agg` or finding a `max timestamp` using `MAX(time_column).

When trying to use pre-aggregations on those measures, I noticed that the pre-agg table is created correctly, however, the select clause is using the `SUM` function on those fields (string/timestamp). 

Below are the Cube Schema, and both queries Im getting from cubejs,cubestore.

Was wandering if thats even possible, Thank you!

**Related Cube.js schema**
```javascript
cube(`Cube`, {
  sql: `SELECT * FROM dummy`,
  preAggregations: {
    main: {
      dimensions: [CUBE.name],
      measures: [CUBE.idsWithThatName]
    }
  },
  measures: {
    idsWithThatName: {
      type: `string`,
      sql: `array_agg(${CUBE.id})`
    }
  },
  dimensions: {
     name: {
       sql: `name`,
       type: `string`
     },
    id: {
      sql: `id`,
      type: `string`
    }
  }
})
```

**Related Cube.js generated SQL**
Regular query, without pre-aggregation: `SELECT name, array_agg(id) FROM dummy group by 1;`
Pre-Aggregation Query:  ```
SELECT `cube__name` `cube__name`,
       sum(`cube__ids_with_that_name`) `cube__ids_with_that_name`
FROM local_cubejs_pre_agg.pre_agg_table_with_uniq_id AS `pre_agg_table`
GROUP BY 1
ORDER BY 2 DESC
LIMIT 20000
```

### PR DESCRIPTION
**Pull request by zubairov**
## Description
This PR fixes issue #7896 by adding support for string, time, and boolean measure types in pre-aggregations.

## Problem
When using pre-aggregations with calculated measures of non-numeric types (string, time, or boolean), the pre-aggregation table is created correctly, but the SELECT clause incorrectly uses the SUM function on these non-numeric fields, which is invalid SQL.

## Solution
- Modified aggregateOnGroupedColumn in BaseQuery.js to use MAX for string/time/boolean types
- Updated aggregationsColumns in PreAggregations.ts with mappings for these types
- Added comprehensive unit tests covering all three new measure types

## Changes
- Modified: packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
- Modified: packages/cubejs-schema-compiler/src/adapter/PreAggregations.ts
- Added: packages/cubejs-schema-compiler/test/unit/pre-agg-string-time-measures.test.ts

## Testing
Added 5 comprehensive unit tests:
1. String measure aggregation test
2. Time measure aggregation test
3. Boolean measure aggregation test
4. String SQL generation test
5. Boolean SQL generation test

## Backward Compatibility
✓ Fully backward compatible
✓ No breaking changes
✓ Existing numeric measure types continue to work as before

Fixes #7896

### CODE REVIEW

File: packages/cubejs-schema-compiler/test/unit/pre-agg-string-time-measures.test.ts [hunk 1]
```diff
@@ -0,0 +1,156 @@
...
+        having: [],
+        orders: [],
+        limit: null,
+        offset: null,
+        responseFormat: 'compact'
+      }
+    );
+
+    const [sql, params] = query.buildSqlAndParams();
+    console.log('Cumulative string measure SQL:', sql);
```
pnedelko:
```suggestion
```

PR #10318, opened 2026-01-20
fix(server-core): Support config dict return from externalDriverFactory

### LINKED ISSUE #10317: `external_driver_factory` is unusable in Python (`cube.py`)
**Describe the bug**
When using `external_driver_factory` in Python configuration (`cube.py`), CubeJS does not convert the returned configuration dictionary to a driver instance. The function is called, but CubeJS expects a fully instantiated `BaseDriver` object with methods like `release(). It makes `external_driver_factory` effectively broken in Python.

**To Reproduce**
1. Create `cube.py` with `external_driver_factory` returning a config dict:

```python
from cube import config

@config("external_driver_factory")
def external_driver_factory(ctx: dict | None) -> dict:
    return {
        "type": "postgres",
        "host": "localhost",
        "port": 5432,
        "database": "cube_external",
        "user": "postgres",
        "password": "secret",
    }
```

2. Configure a primary driver with pre-aggregations enabled
3. Trigger a pre-aggregation build
4. See error:

```
TypeError: driver.release is not a function
    at /cube/node_modules/@cubejs-backend/server-core/src/core/server.ts:671:30
```

**Expected behavior**
`external_driver_factory` in Python should accept configuration dictionaries and automatically convert them to driver instances, matching the behavior of `driver_factory`.

**Screenshots**
N/A

N/A

**Version:**
latest

**Additional context**
Looks like `driverFactory` has conversion logic in `resolveDriver()`, but `externalDriverFactory` does not.

UgoM:
It is my first PR on your repo, so don’t hesitate to help me improve it if I’ve done things wrong.

**Check List**

**Issue**

Fixes #10317

**Description**

`external_driver_factory` in Python (`cube.py`) fails because it expects a `BaseDriver` instance but Python can only return config dicts. This works fine for `driver_factory` which has conversion logic in `resolveDriver()`, but `externalDriverFactory` was missing the same treatment.

This PR adds the same config-to-driver conversion pattern used by `driverFactory`:

```typescript
if (isDriver(val)) {
  driver = val;
} else if (val.type && typeof val.type === 'string') {
  const { type, ...rest } = val;
  driver = CubejsServerCore.createDriver(type, rest);
}
```

Now both factories behave consistently and Python users can use `external_driver_factory` with config dicts.

### DIFF
--- a/packages/cubejs-server-core/src/core/server.ts
+++ b/packages/cubejs-server-core/src/core/server.ts
@@ -653,20 +653,29 @@ export class CubejsServerCore {
             let driver: BaseDriver | null = null;
 
             try {
-              driver = await this.options.externalDriverFactory(context);
-              if (typeof driver === 'object' && driver != null) {
-                if (driver.setLogger) {
-                  driver.setLogger(this.logger);
-                }
+              const val = await this.options.externalDriverFactory(context);
 
-                await driver.testConnection();
-
-                return driver;
+              if (isDriver(val)) {
+                driver = <BaseDriver>val;
+              } else if (val && (<DriverConfig>val).type && typeof (<DriverConfig>val).type === 'string') {
+                const { type, ...rest } = <DriverConfig>val;
+                driver = CubejsServerCore.createDriver(type, rest);
+              } else if (val && typeof val === 'object') {
+                // Accept driver-like objects for backward compatibility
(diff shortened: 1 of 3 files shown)


SaMnCo:
Any chance to get a review for this? It is blocking us. Thanks in advance.

Issue #9488, opened 2025-04-18
Time dimension retuns [object Object] values from mssql


LzbnSam:
**Problem**
I have a problem with time type with MSSQL source database. In playground I got [object Object] result instead of real dates values.
I am using:
- MSSQL 2019
- Latests Cube Core Docker Image

![Image](https://github.com/user-attachments/assets/992558bf-d060-45e7-9f2f-30d4daae49dc)

The generated script works fine and returns values. 

![Image](https://github.com/user-attachments/assets/85296715-8cba-4ef9-8649-4a7a5c128301)

My cube yaml:
```yaml
cubes:
  - name: cube_mes_supply
    sql_table: cube.v_mes_supply
    data_source: dwh_reporting_development

    joins: 
      - name: dim_document_status
        relationship: many_to_one
        sql: "{cube_mes_supply}.ATTB = {dim_document_status.operation_id}"

    dimensions:
      - name: id
        sql: id
        type: number
        primary_key: true      

      - name: company_guid
        sql: COMPANY
        type: string

      - name: mars_reason
        sql: mars_reason
        type: string

      - name: is_cross_docking
        sql: IS_CROSS_DOCKING
        type: boolean

      - name: create_date
        sql: create_datetime
        type: time
      
      - name: test_datetimeoffset
        sql: test_datetimeoffset
        type: time

    measures:
      - name: count
        type: count

      - name: quantity
        sql: Quantity
        type: sum

      - name: amount_rub
        sql: Amount_Rub
        type: sum

    pre_aggregations:
      - name: marsReasonIsCrossDockingMeasure
        measures:
          - amount_rub
          - quantity
        dimensions:
          - mars_reason
          - is_cross_docking
        refreshKey:
          every: 10 minute
```
I think that the problem is that MSSQL returns DATETIMEOFFSET type with nanoseconds.
How can I fix that?

BubuDavid:
Yeap, this is super annoying, I ended up changing to postgress :c, I hope this can be checked for someone..

prabhusiva:
I am also facing the same issue for Date columns property. i am using a MySql database.

igorlukanin:
@zafir83 If you can provide answers to the questions I raise above—that would be helpful.

b2berry:
@vladik-vitre  Latest cube & cubestore images 1.3.63

burma69:
[...]

PR #10411, opened 2026-02-19
fix(mssql-driver): Use UTC timestamps in responses

ovr:
Improve fix for https://github.com/cube-js/cube/issues/9488


packages/cubejs-mssql-driver/src/MSSqlDriver.ts
```diff
@@ -119,22 +121,26 @@ export class MSSqlDriver extends BaseDriver implements DriverInterface {
       requestTimeout: getEnv('dbQueryTimeout') * 1000,
       options: {
         encrypt: getEnv('dbSsl', { dataSource }),
-        useUTC: false
+        useUTC: true
```
ovr: It's a fix, before that, time in responses was local, with `transformRow` in returs an incorrect timestamp

PR #10461, opened 2026-03-05
feat(client-core): Use `throwContinueWait` parameter in `cubeSql` method

MazterQyou:
**Check List**

This PR updates `cubeSql` method in `cubejs-client-core` to make use of the new `throwContinueWait` parameter.

### DIFF
--- a/packages/cubejs-client-core/src/index.ts
+++ b/packages/cubejs-client-core/src/index.ts
@@ -379,7 +379,7 @@ class CubeApi {
         body.error = text;
       }
 
-      if (body.error === 'Continue wait') {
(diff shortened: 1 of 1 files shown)


packages/cubejs-client-core/src/index.ts
```diff
@@ -379,7 +379,7 @@ class CubeApi {
         body.error = text;
       }
 
-      if (body.error === 'Continue wait') {
+      if (body.error?.includes('Continue wait')) {
```
ovr: Let's not do that, because potential it can be a part of error message that are not continue wait.

Issue #9241, opened 2025-02-19
Unexpected GROUP BY when using multi_stage and referencing dimensions with {}


victormorenoi:
**Describe the bug**
When using {} to reference a dimension in Cube with `multi_stage: true`, the generated query introduces an unnecessary GROUP BY, which alters the final aggregation results.

**To Reproduce**
I have the following dimension definitions:
```
- name: count  
  type: sum  
  sql: "{multibrand_weight}"  

- name: count_total  
  type: sum  
  multi_stage: true  
  sql: "{count}"  
  group_by:  
    - retailer_name  
    - category_name
```
If I define count as `sql: multibrand_weight` (without `{}`), referencing the table column directly, the query works correctly, applying the aggregation directly at the table level.
However, when using `sql: "{multibrand_weight}"`,  referencing the Cube dimensions, Cube first executes a SELECT on the dimension and introduces an unnecessary GROUP BY that changes the final result.

**Expected behavior**
I would expect the same behavior when calling the Cube dimension, {}, as when directly referencing the table column.

**Example: beginning of the Queries:**
:white_check_mark: Correct (without {}):
```
WITH cte_0 AS (  
  SELECT  
    "test_cube"."BRAND_NAME" "test_cube__brand_name",  
    sum("test_cube".multibrand_weight) "test_cube__count"  
  FROM "SH_CMI"."MART_MEDIA_HASHTAG" AS "test_cube"  
  WHERE ...  
  GROUP BY 1  
  ORDER BY 2 DESC  
)
```
:x: Incorrect (with {}):
```
WITH cte_0 AS (  
  SELECT  
    "test_cube"."BRAND_NAME" "test_cube__brand_name",  
    "test_cube"."MULTIBRAND_WEIGHT" "test_cube__multibrand_weight"  
  FROM "SH_CMI"."MART_MEDIA_HASHTAG" AS "test_cube"  
  WHERE ...  
  GROUP BY 1, 2  
  ORDER BY 1 ASC  
),  
cte_1 AS (  
  SELECT  
    "test_cube__brand_name",  
    sum("test_cube__multibrand_weight") "test_cube__count"  
  FROM cte_0  
  GROUP BY 1  
  ORDER BY 2 DESC  
)
```

The `GROUP BY` performed in the initial `SELECT` causes the data to be grouped as if applying a `DISTINCT`, leading to incorrect final results.
I would expect the same behavior when calling the Cube dimension, `{}`, as when directly referencing the table column.
[...]

PR #10550, opened 2026-03-25
fix(schema-compiler): multi_stage leaf measure missing FROM clause

### LINKED ISSUE #10549: multi_stage leaf measure throws "lacks FROM clause" when sql references raw column
## Describe the bug

When a `multi_stage` measure's `sql` references a raw column (no `{}` measure dependency), and another `multi_stage` measure references it via `{}`, the query throws:

```
Multi stage member 'cube.measure' lacks FROM clause in sub query
```

## To Reproduce

```yaml
cubes:
  - name: orders
    sql_table: orders

    dimensions:
      - name: id
        sql: id
        type: number
        primary_key: true

      - name: status
        sql: status
        type: string

    measures:
      - name: raw_sum
        sql: amount
        type: sum
        multi_stage: true

      - name: computed
        sql: "{raw_sum}"
        type: number
        multi_stage: true
```

Query `orders.computed` with any dimension — the leaf measure `raw_sum` causes the error.

## Root cause

In `renderWithQuery` (`BaseQuery.js`), the leaf measure has no children in the member dependency graph, so `memberFrom` is null. This makes `fromMeasures` null → `fromSubQuery` null → `fromSql` null. The error check at the end of `renderWithQuery` then throws because `subQuery.from` is undefined.

## Expected behavior

The query should build successfully, falling back to querying the cube's own table when `fromSubQuery` was never built.

## Related

- #9241 — different symptom (unexpected GROUP BY) in the same code path (`renderWithQuery`), but different root cause (non-leaf case with `{}` references)

TheRoot-1:
## Summary

- Fixes a crash in `renderWithQuery` when a `multi_stage` measure's `sql` has no `{}` measure dependencies (leaf node in the dependency graph)
- `memberFrom` is null for leaves, causing `fromSubQuery` → `fromSql` to be null, which triggers the error check: `"Multi stage member '...' lacks FROM clause in sub query"`
- Adds a fallback: when `fromSubQuery` was never built and the withQuery has measures, build a subquery directly from the cube's own table

Fixes: https://github.com/cube-js/cube/issues/10549
Related: https://github.com/cube-js/cube/issues/9241

## Test plan

- [x] Added unit tests in `base-query.test.ts` covering:
  - Leaf `sum` measure does not throw
  - CTE query is generated (multi_stage path is taken)
  - Filters propagate through the leaf fallback subquery
  - `avg` type works as well as `sum`
- [x] All existing unit tests pass (466 tests, 464 pass, 2 pre-existing failures in `error-reporter.test.ts`)

### DIFF
--- a/packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
+++ b/packages/cubejs-schema-compiler/src/adapter/BaseQuery.js
@@ -1780,7 +1780,25 @@ export class BaseQuery {
     // TODO not working yet
     const membersToSelect = measures?.concat(multiStageDimensions).concat(multiStageTimeDimensions);
     const select = fromSubQuery && fromSubQuery.outerMeasuresJoinFullKeyQueryAggregate(membersToSelect, membersToSelect, withQuery.memberFrom.map(f => f.alias));
-    const fromSql = select && this.wrapInParenthesis(select);
+    // Leaf multi_stage measure: sql references a raw column with no measure dependencies,
+    // so fromMeasures is null, fromSubQuery was never built, and select is null.
+    // Fall back to querying the cube's own table directly instead of producing an empty FROM.
+    // See: https://github.com/cube-js/cube/issues/9241
+    let fromSql = select && this.wrapInParenthesis(select);
+    if (!fromSql && !fromSubQuery && withQuery.measures?.length) {
+      const leafFromQuery = this.newSubQuery({
+        measures: withQuery.measures,
+        dimensions: withQuery.dimensions,
+        timeDimensions: withQuery.timeDimensions,
+        multiStageDimensions: withQuery.multiStageDimensions,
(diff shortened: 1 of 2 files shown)


igorlukanin:
Thanks for putting this together, @TheRoot-1.

After looking into it, we've concluded the underlying data model is incorrect rather than this being a bug to fix: a `multi_stage` measure is expected to reference at least one other measure via `{}`, and a `multi_stage` leaf measure pointing at a raw column isn't a supported pattern. The Tesseract SQL planner already rejects it with a clear error (`Multi stage member '...' doesn't reference other members.`), and the correct modeling is to keep the leaf as a regular measure with `multi_stage` only on the derived measure.

Because of that we don't want to add a fallback that accepts the invalid model in the JS planner, so I'm closing the linked issue (#10549) as not planned and this PR along with it. Really appreciate the investigation and the detailed write-up though — thank you!

PR #10573, opened 2026-03-27
feat(cubestore): Add Azure Blob Storage support for remote filesystem

### LINKED ISSUE #8288: Does Cube Core support Azure Blob Storage?
https://cube.dev/docs/product/caching/running-in-production#persistent-storage

> AWS S3, Google Cloud Storage, and Azure Blob Storage (Cube Cloud only) are the only known implementations that provide strong consistency.

I'm considering deploying Cube Core on Azure App Services and attach Azure Blob Storage as the volumes needed for the various services. The documentation wording seems to imply this is not possible and is only available for Cube Cloud.

Could this be clarified?

### LINKED ISSUE #10332: Want to use azure blob storage as external  preaggregation storage
I want to use azure blob storage as external  preaggregation storage
I have preaggregation types rollup and original sql
Please reply with a sample docker compose file and a env file and also share links or resources which might help me

tete17:
## Issue Reference

Resolves #10332
Resolves #8288

## Description of Changes Made

Add Azure Blob Storage as a remote filesystem backend for Cube Store, on top of the `object_store` crate (already a dependency) with its `azure` feature enabled. Credentials are resolved in this order:

- **Account key**: `CUBESTORE_AZURE_ACCESS_KEY`
- **SAS token**: `CUBESTORE_AZURE_SAS_TOKEN`
- **Workload identity**: the standard `AZURE_CLIENT_ID`, `AZURE_TENANT_ID` and `AZURE_FEDERATED_TOKEN_FILE` variables (e.g. Azure AD Workload Identity on AKS) — no extra Cube Store env vars needed
- **Managed identity**: otherwise `object_store` falls back to the host's managed identity via IMDS

### Implementation

- New `AzureBlobRemoteFs` struct implementing `RemoteFs` and `ExtendedRemoteFs`
- `FileStoreProvider::AzureBlob` variant added to config; `CUBESTORE_AZURE_CONTAINER` selects the backend, `CUBESTORE_AZURE_ACCOUNT` is required alongside it
- No credential refresh loop needed — `object_store`'s credential providers handle token lifecycle internally
- Telemetry instrumented with `driver:azure_blob` tags
- `Cargo.lock` only gains the `azure`-feature dependencies (`quick-xml 0.36`, `rustls-native-certs` and extra edges); nothing else is bumped

### Hardening from running this in production

- `list()` / `list_with_metadata()` fall back to `head()` when the prefix is an exact blob name: `object_store`'s `list(Some(&prefix))` treats a prefix containing a path separator as a directory and returns an empty stream, which broke callers resolving `temp://foo` to `temp-uploads/foo` (e.g. `import/mod.rs`).
- `upload_file()` compares the blob size against the number of bytes actually uploaded instead of re-reading the local temp file afterwards. Some callers (RocksDB WAL and `*-current` pointer files under `cachestore` / `metastore`) keep appending to the temp file while the upload is in flight, which tripped `check_upload_file` with a spurious "File sizes for X don't match after upload".

### Environment Variables

| Variable | Required | Description |
|---|---|---|
| `CUBESTORE_AZURE_CONTAINER` | Yes | Blob container name (activates Azure backend) |
| `CUBESTORE_AZURE_ACCOUNT` | Yes | Storage account name |
| `CUBESTORE_AZURE_ACCESS_KEY` | No | Account access key |
| `CUBESTORE_AZURE_SAS_TOKEN` | No | SAS token |
| `CUBESTORE_AZURE_SUB_PATH` | No | Path prefix within the container |
| `CUBESTORE_AZURE_ENDPOINT` | No | Custom endpoint (Azurite, sovereign clouds) |

### Docs

Documented in `docs-mintlify` only (`/docs` is deprecated): the six `CUBESTORE_AZURE_*` entries in the environment variables reference, plus an Azure docker-compose example and an "Azure" authentication section in the Cube Store production guide.

### Rebase notes

Rebased onto current `master` (previous revision conflicted on `rust/cubestore/Cargo.lock` after the `async-std` / crossbeam cleanup). `cargo fmt --check` and `cargo check -p cubestore` pass on `nightly-2025-08-01`.

**Check List**

### DIFF
--- a/docs-mintlify/cube-core/running-in-production.mdx
+++ b/docs-mintlify/cube-core/running-in-production.mdx
@@ -310,6 +310,17 @@ default.
 
 </Warning>
 
+### Azure
+
+Cube Store supports Azure Blob Storage for persistent storage using account
+keys, SAS tokens, or federated token credentials (workload identity, managed
+identity).
+
+When `CUBESTORE_AZURE_ACCESS_KEY` and `CUBESTORE_AZURE_SAS_TOKEN` are not set,
+Cube Store automatically uses federated token credentials via the standard
+`AZURE_CLIENT_ID`, `AZURE_TENANT_ID`, and `AZURE_FEDERATED_TOKEN_FILE`
+environment variables.
+
 ### Garbage collection
 
 Cleanup isn’t done in export buckets; however, it's done in the persistent
--- a/docs-mintlify/cube-core/running-in-production.mdx
+++ b/docs-mintlify/cube-core/running-in-production.mdx
@@ -281,6 +281,34 @@ services:
       - cubestore_router
 ```
 
+A simplified example using Azure Blob Storage might look like:
+
+```yaml
+services:
+  cubestore_router:
+    image: cubejs/cubestore:latest
+    environment:
+      - CUBESTORE_SERVER_NAME=cubestore_router:9999
+      - CUBESTORE_META_PORT=9999
+      - CUBESTORE_WORKERS=cubestore_worker_1:9001
+      - CUBESTORE_AZURE_CONTAINER=<CONTAINER_NAME>
+      - CUBESTORE_AZURE_ACCOUNT=<STORAGE_ACCOUNT_NAME>
+      - CUBESTORE_AZURE_ACCESS_KEY=<STORAGE_ACCOUNT_KEY>
+
+  cubestore_worker_1:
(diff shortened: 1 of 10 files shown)


gowrinath-babu:
When this will be Approved

tete17:
I wish I knew @gowrinath-babu see if you have better luck than me at contacting the maintainers

PR #10660, opened 2026-04-09
fix(bigquery): align DATE() function output to TIMESTAMP for SQL API compatibility

tlangton3:
Fixes #10643. Tableau over the SQL API generates `CAST(orderDate AS DATE) = DATE('2020-01-01')`, which fails on BigQuery with `No matching signature for operator = for argument types: TIMESTAMP, DATE`.

## Problem

`DATE()` here is CubeSQL's UDF, not a Postgres builtin: it's declared to return a day-truncated TIMESTAMP, and the planner types the comparison accordingly. But the BigQuery push-down template rendered the call literally as `DATE('2020-01-01')`, producing a DATE compared against a TIMESTAMP — which BigQuery rejects, killing the query (and in stream mode, the server — see #10875).

## Fix

Override the BigQuery `functions/DATE` template to `TIMESTAMP(DATE({{ args_concat }}))`. The inner `DATE()` keeps the UDF's day-truncation and accepts both string and timestamp arguments; the outer `TIMESTAMP()` restores the type the planner already assigned. Wire-visible types don't change.

## Testing

Integration test runs the Tableau-shaped comparison through the SQL API with an exact expected value (`qty: 4`). Verified against live BigQuery with Tesseract on and off. MSSQL is skipped (no SQL-API `DATE()` support).

### DIFF
--- a/packages/cubejs-schema-compiler/src/adapter/BigqueryQuery.ts
+++ b/packages/cubejs-schema-compiler/src/adapter/BigqueryQuery.ts
@@ -347,6 +347,10 @@ export class BigqueryQuery extends BaseQuery {
     // DATEADD is being rewritten to DATE_ADD
     templates.functions.DATE_ADD = 'DATETIME_ADD(DATETIME({{ args[0] }}), INTERVAL {{ interval }} {{ date_part }})';
     templates.functions.CURRENTDATE = 'CURRENT_DATE';
+    // DATE in CubeSQL is a day-truncating UDF that returns TIMESTAMP, so the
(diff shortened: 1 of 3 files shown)


igorlukanin:
Thanks for this, and for the clear root-cause write-up. The underlying issue (#10643) was fixed a different way in #11251 — normalizing the date-only literal in the comparison so the planner no longer wraps the left side in a redundant `CAST(... AS TIMESTAMP)` — which shipped in v1.7.3. Verified on latest: the SQL API now emits a type-matched `DATE = DATE`.

Since that covers the same case, closing this as superseded. Appreciate the contribution.

PR #10737, opened 2026-04-23
perf(postgres-driver): Fast date, timestamp/tz parsers

ovr:
`getTypeParser` routed all three timestamp OIDs through a single `moment.utc(val).format(moment.HTML5_FMT.DATETIME_LOCAL_MS)` call, which was the dominant per-row cost for result sets with any timestamp columns. Each OID has a fixed text shape, so we can parse directly:

- OID 1082 (DATE): single template literal appending `T00:00:00.000`.
- OID 1114 (TIMESTAMP): slice + pad / truncate fractional, no TZ scan.
- OID 1184 (TIMESTAMPTZ): the driver pins session TZ to UTC by default, so Postgres emits `+00` on the wire — a fast path detects all-zero offsets, strips the suffix, and delegates to the TIMESTAMP parser. Non-UTC sessions fall through to `Date.UTC` + hand-rolled `pad2`/`pad3`/`pad4` formatting, with `setUTCFullYear` handling the `Date.UTC` "years 0-99 remapped to 1900+year" legacy quirk for pre-100 AD dates.

Microbenchmark (Node 22, Apple Silicon; `moment` baseline = `moment.utc(val).format(moment.HTML5_FMT.DATETIME_LOCAL_MS)`).

## Warm (N=2M, with warmup)

Steady-state throughput after V8's JIT has fully optimized both paths — representative of large result sets.

| input | moment | new | speedup |
|-------|--------|-----|---------|
| DATE | ~2770 ns | ~6 ns | ~450x |
| TIMESTAMP (no frac) | ~4440 ns | ~28 ns | ~160x |
| TIMESTAMP (ms) | ~4970 ns | ~48 ns | ~105x |
| TIMESTAMP (us → ms trunc) | ~4920 ns | ~41 ns | ~120x |
| TIMESTAMPTZ (+00) | ~4850 ns | ~35 ns | ~135x |
| TIMESTAMPTZ (ms +00) | ~5360 ns | ~57 ns | ~95x |
| TIMESTAMPTZ (us +00) | ~5250 ns | ~62 ns | ~85x |
| TIMESTAMPTZ (+02) *(non-UTC session)* | ~4690 ns | ~390 ns | ~12x |
| TIMESTAMPTZ (ms +05:30) *(non-UTC)* | ~5350 ns | ~435 ns | ~12x |
| TIMESTAMPTZ (us -03) *(non-UTC)* | ~5200 ns | ~400 ns | ~13x |

## Cold (N=100, no warmup)

No warmup, fresh module cache — representative of the first few rows / short-lived queries. moment sits at ~15 µs/call cold (its locale tables and regex compilation happen lazily on first use), which is where the biggest absolute wins live.

| input | moment | new | speedup |
|-------|--------|-----|---------|
| DATE | ~13.8 µs | ~35 ns | ~390x |
| TIMESTAMP (no frac) | ~18.5 µs | ~139 ns | ~135x |
| TIMESTAMP (ms) | ~16.0 µs | ~135 ns | ~120x |
| TIMESTAMP (us) | ~16.2 µs | ~128 ns | ~125x |
| TIMESTAMPTZ (+00) | ~16.1 µs | ~1230 ns | ~13x |
| TIMESTAMPTZ (ms +00) | ~16.5 µs | ~290 ns | ~55x |
| TIMESTAMPTZ (us +00) | ~15.2 µs | ~390 ns | ~40x |
| TIMESTAMPTZ (+02) *(non-UTC)* | ~14.7 µs | ~880 ns | ~17x |
| TIMESTAMPTZ (ms +05:30) *(non-UTC)* | ~15.3 µs | ~1000 ns | ~15x |
| TIMESTAMPTZ (us -03) *(non-UTC)* | ~16.7 µs | ~1000 ns | ~17x |

Real showcase of the problem, parsing dates tooks ~10% of the whole request, spending 390ms.

image


waralexrom (approved):

KSDaemon (commented):
This is awesome!👍🏻

PR #10852, opened 2026-05-10
feat: Links in the data model

paveltiunov:
**Check List**

Implements the `links` feature for dimensions in the data model, as specified in #10203.

### Design

Links are implemented as **synthetic dimensions**. Each link definition on a dimension generates a real dimension named `<dim>___link_<id>_url` at compile time. This means:

- **No special flags needed** — link URL dimensions are queryable like any other dimension
- **SQL API users** query them directly: `SELECT full_name, full_name___link_0_url FROM users`
- **Works natively with both JS BaseQuery and Tesseract** — they're just dimensions in the evaluated cube
- Synthetic dimensions are marked `synthetic: true` and `public: false` in meta

The `url` field is a standard SQL expression (like `mask.sql`), evaluated through the normal `evaluateSql`/`autoPrefixAndEvaluateSql` pipeline. Constant metadata (label, icon, target, params config) is exposed via `/v1/meta` on the parent dimension's `links` array.

### Documentation Changes

- Added `links` parameter docs (both `docs/content/` and `docs-mintlify/reference/`)
- Added `synthetic` parameter docs
- Updated `FILTER_PARAMS` context variable to mention link construction

### Code Changes

**Schema Compiler (`packages/cubejs-schema-compiler`):**
- `CubeValidator.ts`: `links` validation schema — `url` is `Joi.func()` (SQL expression)
- `CubeEvaluator.ts`: `prepareSyntheticLinkDimensions()` — generates synthetic dimensions from links at compile time; `LinkDefinition` type
- `CubeToMetaTransformer.ts`: Exposes `links` metadata and `synthetic` flag on dimensions in `/v1/meta`

**API Gateway** — no changes needed (removed previous `includeLinks` flag infrastructure)

**Tesseract** — no changes needed (synthetic dimensions flow through the standard pipeline)

<div><a href="https://cursor.com/agents/bc-c27fef75-8e72-464e-a3ae-eae7e2cf638d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-c27fef75-8e72-464e-a3ae-eae7e2cf638d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/packages/cubejs-schema-compiler/test/unit/links.test.ts
+++ b/packages/cubejs-schema-compiler/test/unit/links.test.ts
@@ -138,6 +138,58 @@ cubes:
     }
   });
 
+  it('should reject duplicate link names on same dimension', async () => {
+    const schema = `
+cubes:
+  - name: users
+    sql_table: users
+    dimensions:
+      - name: full_name
+        sql: full_name
+        type: string
+        links:
+          - name: dup
+            label: First
+            url: "{full_name}"
+          - name: dup
+            label: Second
+            url: "{full_name}"
+`;
+    const compilers = prepareYamlCompiler(schema);
+    try {
+      await compilers.compiler.compile();
+      fail('Should have thrown for duplicate link name');
+    } catch (e: any) {
+      expect(e.message || e.toString()).toMatch(/[Dd]uplicate.*dup/);
+    }
+  });
+
+  it('should reject link that collides with user-defined dimension', async () => {
+    const schema = `
+cubes:
+  - name: users
+    sql_table: users
+    dimensions:
+      - name: full_name
+        sql: full_name
+        type: string
+        links:
+          - name: custom
+            label: Link
+            url: "{full_name}"
+      - name: full_name___link_custom_url
+        sql: "'manual'"
+        type: string
+`;
+    const compilers = prepareYamlCompiler(schema);
+    try {
+      await compilers.compiler.compile();
+      fail('Should have thrown for collision');
+    } catch (e: any) {
+      expect(e.message || e.toString()).toMatch(/conflict|collision|already/i);
+    }
+  });
+
   describe('dashboard links', () => {
     const schemaWithDashboardLink = `
 cubes:
--- a/packages/cubejs-schema-compiler/test/unit/links.test.ts
+++ b/packages/cubejs-schema-compiler/test/unit/links.test.ts
@@ -1,0 +1,426 @@
+import { PostgresQuery } from '../../src';
+import { prepareYamlCompiler } from './PrepareCompiler';
+
+describe('Links', () => {
+  const schemaWithLinks = `
+cubes:
+  - name: users
+    sql_table: users
+
+    dimensions:
+      - name: id
+        sql: id
+        type: number
+        primary_key: true
+
+      - name: full_name
+        sql: full_name
+        type: string
+        links:
+          - name: google_search
+            label: Search on Google
+            url: "{full_name}"
+            icon: brand-google
+            target: blank
+          - name: send_email
+            label: Write an email
+            url: "{email}"
+            icon: send
+
+      - name: email
+        sql: email
+        type: string
+`;
+
+  it('should create synthetic link URL dimensions', async () => {
+    const compilers = prepareYamlCompiler(schemaWithLinks);
+    await compilers.compiler.compile();
+
+    const googleDef = compilers.cubeEvaluator.dimensionByPath('users.full_name___link_google_search_url');
+    expect(googleDef).toBeDefined();
+    expect(googleDef.type).toBe('string');
+    expect((googleDef as any).synthetic).toBe(true);
+
+    const emailDef = compilers.cubeEvaluator.dimensionByPath('users.full_name___link_send_email_url');
+    expect(emailDef).toBeDefined();
+    expect(emailDef.type).toBe('string');
+    expect((emailDef as any).synthetic).toBe(true);
+  });
+
+  it('synthetic link dimension exists and can be referenced', async () => {
+    const compilers = prepareYamlCompiler(schemaWithLinks);
+    await compilers.compiler.compile();
+
+    const dimDef = compilers.cubeEvaluator.dimensionByPath('users.full_name___link_google_search_url');
+    expect(dimDef).toBeDefined();
+    expect(dimDef.type).toBe('string');
+    expect(typeof dimDef.sql).toBe('function');
+  });
+
+  it('should NOT include link URL columns unless explicitly queried', async () => {
+    const compilers = prepareYamlCompiler(schemaWithLinks);
+    await compilers.compiler.compile();
+
+    const query = new PostgresQuery(compilers, {
+      measures: [],
+      dimensions: ['users.full_name'],
+    });
+
+    const queryAndParams = query.buildSqlAndParams();
+    const sql = queryAndParams[0];
+
+    expect(sql).not.toContain('___link_');
+  });
+
+  it('should expose links metadata and synthetic flag in meta config', async () => {
+    const compilers = prepareYamlCompiler(schemaWithLinks);
+    await compilers.compiler.compile();
+
+    const { metaTransformer } = compilers;
+    const { cubes } = metaTransformer;
+    const usersCube = cubes.find((c: any) => c.config.name === 'users');
+    expect(usersCube).toBeDefined();
+
+    const fullNameDim = usersCube!.config.dimensions.find(
+      (d: any) => d.name === 'users.full_name'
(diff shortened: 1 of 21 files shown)


keydunov:
FYI @gleb

keydunov (approved):

packages/cubejs-schema-compiler/test/unit/links.test.ts
```diff
@@ -0,0 +1,146 @@
...
+    const { cubes } = metaTransformer;
+    const usersCube = cubes.find((c: any) => c.config.name === 'users');
+    expect(usersCube).toBeDefined();
+
+    const syntheticDim = usersCube!.config.dimensions.find(
+      (d: any) => d.name === 'users.full_name___link_google_search_url'
+    );
+    expect(syntheticDim).toBeDefined();
+    expect(syntheticDim!.public).toBe(false);
+  });
```
igorlukanin: @paveltiunov Are you sure? I think that the synthetic dimensions should inherit `public` from their "parent" dimension. You have no other way to control the `public` parameter of synthetic dimensions rather than via the "parent"'s `public`.

mngr:
Hey @paveltiunov, I'm starting [CUB-557](https://linear.app/cube-d3/issue/CUB-557/links-between-dashboards) ("Links between dashboards"), which sits on top of this PR. There are four things I'd like to change before this lands. Putting them here with the reasoning so we can chat 🙂

The big one: this primitive is great for **external links** on dimensions (Google, Salesforce, mailto, internal tools…), but the headline examples and a couple of schema fields imply **dashboard-to-dashboard navigation** is also a thing it does. We think that's the wrong layer for that use case — it should live in dashboard config, not in the data model. Most of the requests below fall out of that.

---

### 1. Add `primary: boolean` to a link

A dimension can have several external links (Google + Salesforce + mailto, say). For our UX, exactly one of them gets the "default click on the cell value" slot. Right now the only way to express that is array order, which is implicit and silently flips if anyone reorders the YAML — no validation warning, just different behavior in prod.

Concretely, in `packages/cubejs-schema-compiler/src/compiler/CubeValidator.ts`:

```ts
const LinkItemSchema = Joi.object().keys({
  // ...existing fields...
  primary: Joi.boolean().strict(),
});

const LinksSchema = Joi.array()
  .items(LinkItemSchema)
  .custom((value, helpers) => {
    const primaryCount = value.filter((l: any) => l.primary === true).length;
    if (primaryCount > 1) return helpers.error('links.multiplePrimary');
    return value;
  });
```

Thread it through `LinkDefinition` (`CubeEvaluator.ts`), `LinkConfig` (`CubeToMetaTransformer.ts`), and the `/v1/meta` output. A test asserting "two `primary: true` → validation error" would round it out.

---

### 2. Drop the dashboard-URL examples from the docs

The current `links` section in `dimensions.mdx` (both `docs-mintlify/` and `docs/content/`) leans heavily on dashboard URLs:

```yaml
- name: performance
  label: Check performance dashboard
  url: "'https://my-account.cubecloud.dev/new/d/123/dashboards/KSqDYdUz6Ble'"
  params:
    filter_user_id: "{id}"
    filter_country: "{FILTER_PARAMS.users.country}"
```

We're not going to support that pattern in Cube Cloud dashboards. Four reasons:

1. **Permissions** — data-model edits need schema-write access; dashboard authors usually don't have it. A drill-in link shouldn't require a YAML round-trip with a data engineer.
2. **Context** — the same `customer_id` should drill into different dashboards from different parents (Sales Overview → Pipeline Detail vs. Support Overview → Support Detail). A single data-model link can't capture that.
3. **URL fragility** — hardcoded host + deployment ID breaks across environments and assumes the modeler knows the dashboard ID by hand.
4. **Authoring** — picking a destination dashboard belongs in a UI that can browse what exists, not a YAML string field.

If we ship the example as-is, customers will write dashboard links in YAML, find that Cube Cloud's dashboard UI doesn't recognize them as in-app navigation, and file support tickets. CUB-557 is going to deliver dashboard-to-dashboard linking at the dashboard config layer where the UX/permissions/discovery stories are sound.

Could we swap the example for something generic (CRM lookup, internal tool, ticketing system) and add a one-liner: "For dashboard-to-dashboard navigation in Cube Cloud, configure internal links in the dashboard editor — not in the data model"?

---

### 3. Drop `propagate_filters_to_params` and `param_name_for_filters` (or at least scope them)

These two only exist to serialize the current dashboard's filter state into the outgoing URL. Once #2 lands and dashboard-to-dashboard moves out of this layer, they have no consumer. They also signal "this primitive supports cross-dashboard filter propagation," which directly contradicts the doc change we're asking for in #2.

If they stay, the docs and the schema disagree about what this thing does, and Cube Cloud dashboards will explicitly *not* honor them — that's a ticket farm.

Preference: remove both from `LinkItemSchema`, `LinkDefinition`, `LinkConfig`, the meta transformer, and the docs. If removing them is awkward at this point, the fallback is to rename + scope-restrict in docs ("for external services that accept a JSON filter blob; ignored by Cube Cloud"), but removal is the cleaner call IMO.

---

### 4. Figure out per-row `params` resolution

The docs say `params` values can reference per-row dimension values (`filter_user_id: "{id}"`), but the implementation only generates a synthetic dim for the `url` field. The `params` map gets exposed in `/v1/meta` as a `Record<string, string>` of templated strings, with no per-row resolution.

Two readings of the current state:

- The docs are aspirational and the implementation is incomplete.
- The frontend is expected to substitute `{id}` by separately querying that dimension and matching it client-side. That's fragile (the frontend has to auto-add referenced dims to every query the user didn't ask for) and breaks `{FILTER_PARAMS.users.country}` entirely (totally separate resolution path).

Two ways to fix:

- **(a) — preferred.** One synthetic dim per templated `params` entry: `<dim>___link_<name>_param_<key>`. SQL expression = the templated value. Frontend reads each per-row and assembles the URL.
- **(b)** Bake all `params` (after substitution + URL-encoding) into the `url` synthetic dim itself, server-side. Frontend reads one fully-formed URL per row.

Either works for us. Current state (template strings exposed verbatim with no resolution path) doesn't.

---

### Nits / non-blockers

- `dashboard: <id-or-slug>` shortcut would've been nice for ergonomic same-workspace links, but moot once #2 pulls dashboard-to-dashboard out of this layer.

---

Happy to chat here, Slack, or sync. Once these settle, the primitive is exactly what we want for external links from a dashboard cell 👍

mngr:
Adding a fifth ask after working through the migration story with @keydunov:

### 5. Canonical dashboards on the cube

We need a named indirection layer for "the canonical detail dashboard for this entity." Concrete use case: a workspace has 100 widgets across many dashboards linking to a "Customer Overview" dashboard. They build a "Customer Overview v2" and want to redirect all references in one place. Touching 100 widget configs is not a real migration path.

This isn't a contradiction with ask #2 (data-model should not own arbitrary dashboard-to-dashboard URLs). Canonical aliases are an **opt-in named pointer**, not auto-applied routing. Dashboard widgets that want to track the canonical reference the alias by name; widgets with context-specific destinations keep their concrete `dashboardId`. The alias just gives modelers a single place to redirect entity-level drill-ins.

**Proposed shape** — new top-level field on the cube:

```yaml
cubes:
  - name: Customers
    canonical_dashboards:
      - name: overview                       # alias name; cube-unique
        label: Customer overview             # default UI label
        dashboard: <dashboard-public-id>     # Cube Cloud workspace dashboard
        filter_mappings:                     # default filter mapping when consumers click
          - target_dimension: Customers.id
            mode: row_value
            source_dimension: Customers.id
        propagate_all_filters: false
```

Validation: unique `name` per cube; `dashboard` is a string (public_id); `filter_mappings` and `propagate_all_filters` optional. Surface as `canonicalDashboards: CanonicalDashboardConfig[]` on the cube config in `/v1/meta`.

The dashboard side reads the canonical from meta at render time. Modeler updates `Customers.canonical_dashboards.overview.dashboard` → every widget referencing `Customers.overview` follows on next render. No widget edits required.

Happy to discuss whether the right home for this is the cube schema or somewhere else (e.g. a separate workspace-level registry) — but the dashboards-team take is that it belongs with the cube because the canonical drill-in is metadata about the entity, similar to a default detail view.

PR #10892, opened 2026-05-15
feat(tesseract): default value filters for views (CORE-357)

waralexrom:
## Summary

Adds support for view-level default value filters (CORE-357). A view can now declare:

```yaml
filters:
  - member: currency
    operator: equals
    values: [USD]
    unless: [currency]
```

Whenever the view is touched by a query, the default filter is added; an explicit user filter on the same member can release it via `unless`. The motivating customer case is a `switch` dimension where the planner otherwise rolls up across all switch values.

### DIFF
--- a/packages/cubejs-schema-compiler/test/integration/postgres/view-default-value-filters.test.ts
+++ b/packages/cubejs-schema-compiler/test/integration/postgres/view-default-value-filters.test.ts
@@ -6,18 +6,6 @@ import { dbRunner } from './PostgresDBRunner';
 describe('View default value filters', () => {
   jest.setTimeout(200000);
 
-  // Two flavours of default filter live side-by-side:
-  //
-  //   * `orders_view_*_real` — `country` is a real string dimension, the
-  //     default filter rewrites the WHERE clause.
-  //   * `orders_view_*_switch` — `currency` is a virtual `type: switch`
-  //     dimension; the default filter pins the switch union to one branch.
-  //
-  // Each cube exposes both an `_unconditional` view (no `unless`) and a
-  // `_with_unless` view (`unless: [<member>]`). The seed has five rows with
-  // mixed `country` so we can spot bugs by row count alone.
-  //
-  // language=YAML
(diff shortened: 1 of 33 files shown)


ovr (approved):

packages/cubejs-schema-compiler/test/integration/postgres/view-default-value-filters.test.ts
```diff
@@ -0,0 +1,209 @@
...
+          },
+        ],
+      },
+      [
+        {
+          orders_view_real_with_unless__count: '1',
+        },
+      ]
+    ));
+  });
```
igorlukanin, line 157: This one is correct but the test would be more robust if we had more data in the seed cube, i.e., if the `orders_view_real_with_unless__count` would be 3 or 4, not 1, in this test.

packages/cubejs-schema-compiler/test/integration/postgres/view-default-value-filters.test.ts
```diff
@@ -0,0 +1,209 @@
...
+      - name: country
+        sql: country
+        type: string
+
+      - name: currency
+        type: switch
+        values:
+          - USD
+          - EUR
+          - GBP
```
igorlukanin, line 37: Ideally, in a good test, this switch dimension should be used (e.g., in a case measure). Right now it just dangles in the air.

packages/cubejs-schema-compiler/test/integration/postgres/view-default-value-filters.test.ts
```diff
@@ -0,0 +1,209 @@
...
+        measures: ['orders_view_switch_with_unless.count'],
+        dimensions: ['orders_view_switch_with_unless.currency'],
+        filters: [
+          {
+            member: 'orders_view_switch_with_unless.currency',
+            operator: 'equals',
+            values: ['EUR'],
+          },
+        ],
+        order: [{ id: 'orders_view_switch_with_unless.currency' }],
```
igorlukanin, line 199: Just curious, why do we need `order` in each test?

PR #10918, opened 2026-05-20
feat(api-gateway): granularities config and /v1/granularities endpoint

igorlukanin:
## Summary

Engine-side foundation for configurable, deployment-wide granularities:

- **New `granularities` shape on time dimensions.** Accepts a dict `{ includes?, excludes?, custom? }` alongside the legacy flat array. `includes`/`excludes` (string list or `'*'`) select from the enabled set; `custom` defines per-dimension granularities. Mutually exclusive except for `includes: '*' + excludes: [...]`.
- **Deployment-wide granularities config.** Configurable via `CUBEJS_GRANULARITIES` (comma-separated list), per-custom env vars `CUBEJS_GRANULARITIES_<NAME>_INTERVAL|TITLE|OFFSET|ORIGIN`, or `config.granularities` (list or context-dependent function). File config fully replaces env vars.
- **`GET /v1/granularities`.** New read-only endpoint returning the resolved set (built-ins + customs) for the request's security context.
- **`/v1/meta` enrichment.** Each time-dimension granularity now carries `type` (`'built-in'` / `'custom'`), `title`, `format` (d3-time-format), and `interval`. Per-dimension `includes`/`excludes` are resolved against the deployment config so the response reflects the effective set.
- **`/v1/load` + `/v1/cubesql` annotations.** Same `type`/`format` enrichment flows through `prepareAnnotation` to query responses.
- **Client SDK types widened.** `Granularity` / `GranularityAnnotation` and the Rust transport struct in `@cubejs-backend/native` now carry `type` and `format`.

Out of scope (separate follow-ups):
- `allow_no_granularity` parameter end-to-end
- XMLA / DAX exposure of custom granularities
- Elasticsearch + MS Fabric driver SQL for custom granularities
- AI Engineer prompts
- Documentation, telemetry, release notes

## Test plan

- [x] Added unit tests covering every row of the new resolution table (`granularities-shape.test.ts`) and global config precedence (`granularities-config.test.ts`).
- [x] Extended `cube-validator.test.ts` for the new dict shape: includes/excludes mutual-exclusion, built-in shadowing, legacy form acceptance.
- [x] Extended `prepare-annotation.test.ts` for the new `type` / `format` fields on the granularity annotation.
- [x] Existing pre-agg matching tests still pass (no hash-affecting changes).
- [x] Manual end-to-end against a Postgres scratch model:
  - `/v1/granularities` returns 8 built-ins + customs sourced from `CUBEJS_GRANULARITIES`.
  - `/v1/meta` for a time dim with `granularities: { includes: ['year', 'quarter'], custom: [...] }` returns exactly those built-ins plus the local custom; a dim with the legacy flat array gets all enabled built-ins plus its locals.
  - `/v1/load` with `granularity: 'week'` returns `{ type: 'built-in', title: 'Week', interval: '1 week', format: '%b %-d, %Y' }`.
- [ ] CI: lint + tsc + unit + integration.


packages/cubejs-schema-compiler/src/compiler/GranularityResolver.ts
```diff
@@ -0,0 +1,112 @@
...
+        out[builtInName] = { ...def, type: 'built-in' };
+      }
+    }
+    for (const [name, def] of Object.entries(globalCustom)) {
+      const passesIncludes = includesAllowsAll || includesSet!.has(name);
+      const blockedByExcludes = excludesSet!.has(name);
+      if (passesIncludes && !blockedByExcludes) {
+        out[name] = { ...def, type: 'custom' };
+      }
+    }
```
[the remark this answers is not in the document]
igorlukanin: Good catch. Fixed in 5d91cf7227: the global-custom loop now skips names that shadow an enabled built-in, so an overridden built-in keeps `type: 'built-in'` with its title/format folded in. Added a regression test.

packages/cubejs-schema-compiler/src/compiler/utils.ts
```diff
@@ -4,6 +4,11 @@ import { camelize } from 'inflection';
 const IGNORE_CAMELIZE = {
   1: {
     granularities: true,
+  },
+  // Custom granularity names live one level deeper than the legacy flat-array form
+  // (`granularities.custom.<name>`), so they need their own guard.
+  2: {
+    custom: true,
```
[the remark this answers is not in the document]
igorlukanin: Scoped the guard to `granularities.custom` (parent-key check) in 5d91cf7227 so an unrelated `custom` property elsewhere in the tree is unaffected. On the "is it needed" part: yes — `camelizeCube` runs before `normalizeDimensionGranularities`, so at camelize time `granularities` is still the raw dict and the custom names need protecting.

igorlukanin:
Reworked where effective granularities are resolved: moved from per-request gateway enrichment into the compiler layer.

- Env/static `granularities` configs are now baked into the compiled meta once (their hash is folded into the compiler version, so a config change recompiles). The context-function form resolves per request against a bounded (16-entry) promise-valued LRU of enriched meta variants keyed by a canonical config hash, owned by the compiled model so a recompile drops it. Concurrent misses dedup to one build; failed builds self-evict.
- `compilerId` now composes base → visibility mask → granularity hash, so CubeSQL and other compilerId-keyed caches are tenant-correct for context-dependent configs.
- Steady-state granularity cost on a 6k-cube model drops from ~2 ms/request to ~0.01 ms (baked/cache hit); a new tenant config costs one ~2 ms build.
- Time dimensions without local customization share a single default set instance per config instead of allocating identical arrays per dimension.
- `/v1/meta?extended` no longer includes the internal `granularitiesBlock` (transport-only field, never documented); `effectiveGranularities` appears instead. Non-extended `/v1/meta`, `/v1/load`, `/v1/cubesql`, `/v1/granularities` shapes are unchanged, except local custom granularities now correctly carry their `format` and `title` always falls back to the granularity name.
- Fixed a pre-existing arg swap in the pre-agg job status path (`metaConfigExtended(context, ctx)` → `metaConfigExtended(ctx, …)`) that selected visibility for the wrong tenant.

packages/cubejs-schema-compiler/src/compiler/GranularityConfigHash.ts
igorlukanin, line 1: Does this need to be separate file?

igorlukanin: No — folded into GlobalGranularitiesConfig.ts in abb0405476 (the hash is a function of that config, so it lives next to the type).

packages/cubejs-api-gateway/src/gateway.ts
```diff
@@ -453,6 +463,17 @@
       })
     );
 
+    app.get(
+      `${this.basePath}/v1/granularities`,
+      userMiddlewares,
```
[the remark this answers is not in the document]
igorlukanin: Matches the existing posture: none of the gateway routes rate-limit in-process (7 open alerts for this rule across /v1/meta, /v1/load, etc.) — rate limiting is applied at deployment ingress. Keeping /v1/granularities consistent with the other meta-scope endpoints.

igorlukanin:
Follow-up after a full review + live-validation pass (Codex normal/adversarial + a 29-agent review workflow, plus 16 scenarios driven against a real server on DuckDB):

**Critical fix — global custom granularities are now queryable.** A global custom from `config.granularities` (e.g. a `fiscal_year`) was advertised in `/v1/meta` and `/v1/granularities` but threw `Granularity "X" does not exist in dimension` on `/v1/load` — the SQL layer only knew per-dimension customs. Fixed by resolving the request's global config once in `getSqlGenerator` and threading a per-dimension custom-granularity lookup (built via the same `resolveDimensionGranularities` + `granularityInputs` as meta, so includes/excludes stay identical) into the query; `resolveGranularity` consults it as a fallback via a lightweight per-request evaluator proxy. Works for env/static AND context-function configs, honors dimension exclusions, keeps the shared evaluator immutable, and folds the definitions into the query/cache key so there's no cross-tenant pollution. Verified end-to-end: fiscal_year/half_year/sprint/retail_period all resolve through /v1/load now.

**Hardening (all with regression tests):**
- Built-in interval override no longer advertised (SQL always buckets `1 <unit>`; title/format overrides still apply).
- Included view dimensions preserve their `includes`/`excludes` (were being reset to `*`).
- Validator no longer rejects a legacy custom granularity literally named `includes`/`excludes`/`custom` (aligned with the resolver's value-shape check).
- `config.granularities` customs without an interval are dropped instead of advertised; `CUBEJS_GRANULARITIES_<NAME>_TITLE` now applies to built-ins.
- Prototype-chain names (`__proto__`, `constructor`) no longer misclassified as built-ins.
- Config hash is now faithful to the serialized output (non-string fields sanitized at resolution), closing a cross-tenant variant-cache collision.

**Working as intended (not changing):** granularity gating is meta-only — a `granularities` restriction is a display/UX filter, not access control. A built-in granularity is just `DATE_TRUNC` over an already-queryable timestamp, so it stays executable regardless of the offered set; real restriction is row/member-level access policies.

packages/cubejs-api-gateway/src/gateway.ts
```diff
@@ -714,6 +730,8 @@ class ApiGateway {
       const cubesConfig = onlyViews
         ? metaConfig.cubes.filter((c: any) => c.config?.type === 'view')
         : metaConfig.cubes;
+      // Time dimensions arrive from CompilerApi with `effectiveGranularities` already attached —
+      // baked into the compiled model at compile time (resolved once per appId, all config forms).
```
igorlukanin: Unnecessary comment.

packages/cubejs-api-gateway/src/gateway.ts
```diff
@@ -738,6 +756,51 @@ class ApiGateway {
...
+            name,
+            title: def.title || name,
+          };
+          if (def.interval !== undefined) entry.interval = def.interval;
+          if (def.origin !== undefined) entry.origin = def.origin;
+          if (def.offset !== undefined) entry.offset = def.offset;
+          if (def.format !== undefined) entry.format = def.format;
+          granularities.push(entry);
+        }
+      }
```
igorlukanin: Does this have to happen here? Can't it just go from the compiler API?

packages/cubejs-schema-compiler/src/compiler/PrepareCompiler.ts
```diff
@@ -40,6 +41,10 @@ export type PrepareCompilerOptions = {
   compiledScriptCache?: LRUCache<string, vm.Script>;
   compiledYamlCache?: LRUCache<string, string>;
   compiledJinjaCache?: LRUCache<string, string>;
+  // Resolved-once global granularities config for this appId, baked into the compiled model by
+  // CubeToMetaTransformer. CompilerApi resolves all config forms (env / static / function) before
+  // compile and passes the result here.
```
igorlukanin: This is one of many instances where really lengthy comments are left. Make concise comments in the source code here and everywhere in this PR.

packages/cubejs-api-gateway/src/gateway.ts
igorlukanin: Removed.

packages/cubejs-api-gateway/src/gateway.ts
igorlukanin: Moved — the catalog assembly now lives in `CompilerApi.getGranularities()`; the endpoint just calls it and serializes.

packages/cubejs-schema-compiler/src/compiler/PrepareCompiler.ts
igorlukanin: Trimmed the long comment blocks across the PR down to the why.

packages/cubejs-schema-compiler/src/compiler/CubeSymbols.ts
```diff
@@ -603,6 +642,75 @@ export class CubeSymbols implements TranspilerSymbolResolver, CompilerInterface
...
+        if (!dim.granularitiesBlock && 'granularities' in dim) {
+          // Keep the raw user value for the validator (it runs after this and would otherwise only
+          // see the extracted customs, never the includes/excludes/custom dict).
+          dim.rawGranularities = dim.granularities;
+          const block: NormalizedGranularitiesBlock = normalizeGranularitiesBlock(dim.granularities);
+          dim.granularitiesBlock = block;
+          dim.granularities = block.custom;
+        }
+
+        this.mergeGlobalCustomsIntoDimension(dim);
```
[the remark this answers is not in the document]
igorlukanin: Confirmed — reproduced it: a global `{ name: 'retail_period', interval: '5 days' }` fails compilation on a cube whose only time dim declares no granularities ("Arbitrary intervals cannot be used without origin point specified"). Not fixing it here: option 2 moves the merge back past validation, which is exactly the placement the CTO asked us to change in this revision, so the choice between config-time validation and reverting the phase is recorded in the planning doc for a decision rather than picked autonomously.

packages/cubejs-schema-compiler/src/compiler/CubeToMetaTransformer.ts
```diff
@@ -301,11 +340,28 @@ export class CubeToMetaTransformer implements CompilerInterface {
           const dimensionVisibility = isCubeVisible
             ? this.isVisible(extendedDimDef, !extendedDimDef.primaryKey)
             : false;
-          const granularitiesObj = extendedDimDef.granularities;
+          const { granularitiesBlock } = extendedDimDef as any;
+          // The deprecated `granularities` meta field lists only the model's own custom
```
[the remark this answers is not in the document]
igorlukanin: Confirmed — with a global `fiscal_year`, `Orders.created_at` (declares nothing) now reports it in the deprecated `granularities` field where it was previously undefined. That contradicts the wire-shape-unchanged claim in the planning doc, so it is recorded there alongside the phase-1 placement question it comes from.

packages/cubejs-schema-compiler/src/compiler/CubeSymbols.ts
```diff
@@ -1545,7 +1660,11 @@ export class CubeSymbols implements TranspilerSymbolResolver, CompilerInterface
       return { interval: `1 ${granName}` };
     }
 
-    return cube?.[dimName]?.[gr]?.[granName];
+    // A local custom wins; otherwise fall back to a global custom from the resolved config.
```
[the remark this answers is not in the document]
igorlukanin: Right that the two halves disagree — the fallback is unconditional, so an excluded global still resolves in SQL. Whether exclusion is display-only or enforced at the SQL layer is a product call, so it is recorded for a decision rather than pinned one way here; the test goes in once that is settled.

PR #10942, opened 2026-05-23
fix(query-orchestrator): replace pre-aggregation usage aliases before base table names (fixes #10929)

### LINKED ISSUE #10929: Tesseract multi-fact view with paired source-db pre-aggregations fails reading missing __usage_N relation
Summary

We are seeing a failure in Cube v1.6.46 when querying a Tesseract multi-fact view that combines two facts through shared conformed dimensions and uses two matched source-database pre-aggregations.

The raw multi-fact view works correctly without rollups. Each single-fact rollup also works when queried alone. The failure only happens when the two matched pre-aggregations are used together in the final multi-fact /load query.

Cube successfully builds the base pre-aggregation tables and /sql confirms the query matches them, but /load fails because the final query tries to read internal __usage_0 / __usage_1 relations that were not created.

Environment
Cube version: v1.6.46
Planner: Tesseract enabled
Database: PostgreSQL
Pre-aggregations: source-database pre-aggregations, external: false
Pre-aggregation schema: prod_pre_aggregations

We also reproduced this after enabling the Tesseract pre-aggregation flags:

CUBEJS_TESSERACT_SQL_PLANNER=true
CUBEJS_TESSERACT_PRE_AGGREGATIONS=true
CUBEJS_CUBESTORE_ROLLING_WINDOW_JOIN=true
Model shape

We have a multi-fact view combining revenue actuals and revenue budgets through shared dimensions:

revenue_actuals  -> shared business/client/calendar dimensions
revenue_budgets  -> shared business/client/calendar dimensions

Conceptually:

revenue_actuals.actual_amount
revenue_budgets.budget_amount
shared_client_dimension.client
shared_calendar_dimension.month_date

The view query uses both fact measures plus shared dimensions.

The relevant pre-aggregations are source-database rollups:

revenue_actuals.client_m
revenue_budgets.client_m

Both are built with external: false.

What works

The following all work:

1. Raw multi-fact view query without rollups.
2. Single-fact actuals rollup query.
3. Single-fact budgets rollup query.
4. Jobs API build for both pre-aggregations.
5. /sql matching the final query to the pre-aggregations.
What fails

The failure happens only on the final two-fact /load query after the pre-aggregations are built and matched.

Jobs API successfully builds both base pre-aggregation tables:

prod_pre_aggregations.ra_client_m_kxpdzefn_jd5um1xd_1l0u355
prod_pre_aggregations.rb_client_m_at5252us_h0f23hsf_1l0u355

The physical tables exist in prod_pre_aggregations, and they contain data.

/sql confirms the query matches the pre-aggregations.

But /load tries to read:

prod_pre_aggregations.ra_client_m_kxpdzefn_jd5um1xd_1l0u355__usage_0

That relation does not exist.

Error

Example error:

HTTP 400

Error: Internal: Error during planning:
Table prod_pre_aggregations.ra_client_m_rk21zgn_ht21vlul_1l0u5gj__usage_0 was not found

TablePath {
  table: IdRow {
    id: 18446744073709551615,
    row: Table {
      table_name: "ra_client_m_rk21zgn_ht21vlul_1l0u5gj__usage_0",
      schema_id: 18446744073709551615,
      columns: [],
      locations: None,
      import_format: None,
      has_data: false,
      is_ready: false,
      created_at: Some(...),
      ...
    }
  },
  schema: IdRow {
    id: 0,
    row: Schema {
      name: "prod_pre_aggregations"
    }
  },
  schema_lower_name: "prod_pre_aggregations",
  table_lower_name: "ra_client_m_rk21zgn_ht21vlul_1l0u5gj__usage_0"
}
Reproduction observations

We ran a focused verification:

1. Triggered exact Jobs API build for the two pre-aggregations.
2. Both jobs completed successfully.
3. Jobs API returned physical table names.
4. Verified the physical tables exist.
5. Verified the tables contain data.
6. Ran /sql for the exact view query.
7. /sql referenced prod_pre_aggregations and the expected client_m rollups.
8. Ran exactly one /load after waiting, with no rebuild and no restart.
9. /load failed immediately with missing __usage_0 relation.

The final /load failure was fast, around 250 ms, so this does not appear to be a refresh/build timing issue.

Additional detail: PostgreSQL identifier truncation masked this once

We found that a similar business-line rollup appeared to work only because the generated physical name was already at PostgreSQL’s 63-character identifier limit.

When Cube appended __usage_0, PostgreSQL truncated the identifier back to the existing base table name. That made the query appear to work, but it seems accidental and not something we should rely on.

With shorter rollup names, the generated __usage_0 name is a distinct relation, and /load fails because that relation was never created.

Expected behavior

One of the following should happen:

1. Cube should read the built base pre-aggregation tables directly, or
2. Cube should create/publish the expected __usage_0 / __usage_1 relations, or
3. Cube should document that paired pre-aggregations for Tesseract multi-fact views are unsupported in this mode.
Actual behavior

Cube builds and matches the base pre-aggregation tables, but the final /load plan tries to read internal __usage_N relations that were not created.

Question

For Cube v1.6.x, is this pattern supported?

Tesseract multi-fact view
+ two matched source-database pre-aggregations
+ external: false
+ final /load query

If yes, what component is expected to create or resolve the internal __usage_N relations?

If no, should this be modeled differently, for example by using one scenarioized source, a warehouse materialized table, or avoiding paired pre-aggregations for multi-fact views?

Current workarounds under consideration
1. Disable the paired source-db rollups and use the raw multi-fact query.
2. Materialize a monthly actual/budget table in the warehouse and wrap it with Cube.
3. Remodel as a single scenarioized Cube source:
   scenario = actual | budget
   amount
   month_date
   client / business_line / other dimensions

Would appreciate guidance on whether this is a known issue, an unsupported pattern, or a configuration/modeling gap.

mystichronicle:
**Check List**

#10929

Fixed the `__usage_N` pre-aggregation alias replacement path so usage-specific table names are rewritten before their base table name. This prevents the final `/load` query from resolving to a missing `__usage_0` relation in Tesseract multi-fact views with paired source-db pre-aggregations.

Also added a regression test covering the alias replacement order, and resolved the orchestrator package typecheck issues that surfaced while validating the change.

### DIFF
--- a/packages/cubejs-query-orchestrator/src/orchestrator/Logger.ts
+++ b/packages/cubejs-query-orchestrator/src/orchestrator/Logger.ts
@@ -1,0 +1,5 @@
+export type LoggerFnParams = {
+  [key: string]: any;
+};
+
+export type LoggerFn = (msg: string, params: LoggerFnParams) => void;
--- a/packages/cubejs-query-orchestrator/src/orchestrator/PreAggregationLoader.ts
+++ b/packages/cubejs-query-orchestrator/src/orchestrator/PreAggregationLoader.ts
@@ -1,6 +1,6 @@
 import R from 'ramda';
 import crypto from 'crypto';
-import { getEnv, MaybeCancelablePromise, LoggerFn } from '@cubejs-backend/shared';
+import { getEnv, MaybeCancelablePromise } from '@cubejs-backend/shared';
 import {
   cancelCombinator,
   DownloadQueryResultsResult,
@@ -13,6 +13,7 @@ import {
   UnloadOptions
 } from '@cubejs-backend/base-driver';
 import { DriverFactory } from './DriverFactory';
+import { LoggerFn } from './Logger';
 import { PreAggTableToTempTableNames, QueryCache, QueryWithParams } from './QueryCache';
 import { ContinueWaitError } from './ContinueWaitError';
 import { LargeStreamWarning } from './StreamObjectsCounter';
--- a/packages/cubejs-query-orchestrator/src/orchestrator/PreAggregationPartitionRangeLoader.ts
+++ b/packages/cubejs-query-orchestrator/src/orchestrator/PreAggregationPartitionRangeLoader.ts
@@ -10,7 +10,6 @@ import {
   timeSeries,
   localTimestampToUtc,
   parseUtcIntoLocalDate,
-  LoggerFn,
 } from '@cubejs-backend/shared';
 import { InlineTable, TableStructure } from '@cubejs-backend/base-driver';
 import { DriverFactory } from './DriverFactory';
@@ -28,6 +27,7 @@ import {
 } from './PreAggregations';
 import { PreAggregationLoader } from './PreAggregationLoader';
 import { PreAggregationLoadCache } from './PreAggregationLoadCache';
+import { LoggerFn } from './Logger';
 
 const DEFAULT_TS_FORMAT = 'YYYY-MM-DDTHH:mm:ss.SSS';
 
--- a/packages/cubejs-query-orchestrator/src/orchestrator/PreAggregations.ts
+++ b/packages/cubejs-query-orchestrator/src/orchestrator/PreAggregations.ts
@@ -1,6 +1,6 @@
 import R from 'ramda';
 import crypto from 'crypto';
-import { getEnv, LoggerFn } from '@cubejs-backend/shared';
+import { getEnv } from '@cubejs-backend/shared';
 
 import { BaseDriver, InlineTable, } from '@cubejs-backend/base-driver';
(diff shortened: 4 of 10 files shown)


mystichronicle:
@KSDaemon Could you please review this?

KSDaemon:
@mystichronicle Hey! Well, I potentially could review :) But I'm not a part of the Core Team anymore, sorry. So I delegate it to the Cube team.

igorlukanin:
Thanks for the fix @mystichronicle! This was independently resolved in v1.6.49 via #10925 (commit 81b9de0), which populates `usageTargetTableNames` from the pre-aggregation's `usageMapping` in the source-DB loader branch. Closing this as no longer needed — appreciate the contribution.

PR #10985, opened 2026-06-01
feat(api-gateway): accept a list of API secrets for verification

bsod90:
Adds `CUBEJS_API_SECRETS` (comma-separated) as a sibling of `CUBEJS_API_SECRET`. When set, `CUBEJS_API_SECRETS` takes precedence: the gateway tries each secret in the list when verifying a JWT and accepts the token if any of them matches the signature. Signing behavior is unchanged — callers still sign with whatever single secret they hold.

This enables zero-downtime rotation of the API secret. During a rotation window operators can publish the outgoing, current, and incoming secrets together in `CUBEJS_API_SECRETS`; in-flight tokens signed by the outgoing secret keep verifying until they expire or are reissued.

- New optional `apiSecrets: string[]` on `ApiGatewayOptions` and `CreateOptions` (server-core).
- `createDefaultCheckAuth` iterates `apiSecrets` when present; otherwise falls back to the singular `apiSecret`. An explicit `options.key` (used by the playground/system auth path) still wins, so playground behavior is unchanged.
- JWK auth is unaffected — when `jwkUrl` is configured the secret is fetched dynamically from JWKS, so iterating a static list would be meaningless.
- `Cube.apiSecrets()` static accessor mirrors `Cube.apiSecret()`.

Tests cover: accept any listed secret, reject unlisted, list takes precedence over singular, empty list falls back, playground path isolated.

**Check List**


packages/cubejs-server-core/src/core/OptsHandler.ts
```diff
@@ -468,6 +469,7 @@ export class OptsHandler {
       externalDriverFactory,
       externalDialectFactory,
       apiSecret: process.env.CUBEJS_API_SECRET,
+      apiSecrets: parseApiSecretsEnv(process.env.CUBEJS_API_SECRETS),
```
paveltiunov: getEnv

paveltiunov (approved):

PR #10996, opened 2026-06-02
fix(cubesql): Don't mix filter params between pushdown subqueries

waralexrom:
## Summary

Fixes wrong query results in SQL API push down: when a pushed-down query contains several independently generated subqueries (CTE joins, scalar subqueries), filter params of the second and subsequent subqueries were substituted with values from the preceding ones. E.g. in a CTE join where one side filters by date range and the other by date range + `service IN (...)`, the IN list silently received the date values instead of service names, dropping the tail IN values and returning understated numbers without any error. Reproduces with both the JS planner and Tesseract.

## Root cause

Each subquery numbers its `$N$` value placeholders from zero. On merge into the parent `SqlQuery`, values were concatenated as-is (`extend_values`) without renumbering placeholders in the subquery SQL, so `finalize_query` resolved colliding indexes against the combined values array.

## Changes

- Replace `SqlQuery::extend_values` with `add_values` + `remap_placeholders`: subquery values are added through `add_value` (deduplicated by value, returns local→combined index mapping), and subquery SQL placeholders are remapped accordingly
- Apply at all three merge sites: scalar subqueries (`prepare_subqueries_sql`), join subqueries in the push-to-cube path, and join subqueries (including the join condition SQL) in the regular `WrappedSelectNode::generate_sql` path

## Testing

- New regression test `test_wrapper_subqueries_filter_params_not_mixed` (two scalar IN-subqueries with filter params; fails without the fix)
- Full `cargo test -p cubesql --lib`: 712 passed
- Verified end-to-end on a local Cube server (psql → cubesql → Cube → Postgres) with the original CTE-join query shape: correct results and correct bind params in the database, with both `CUBEJS_TESSERACT_SQL_PLANNER=true` and `false`

### DIFF
--- a/rust/cubesql/cubesql/src/compile/engine/df/wrapper.rs
+++ b/rust/cubesql/cubesql/src/compile/engine/df/wrapper.rs
@@ -176,8 +176,43 @@ impl SqlQuery {
         index
     }
 
-    pub fn extend_values(&mut self, values: impl IntoIterator<Item = Option<String>>) {
-        self.values.extend(values);
+    /// Adds values of an independently generated subquery, returning the mapping from
+    /// subquery-local value indexes to indexes in `self.values`. Placeholders in the
+    /// subquery SQL must be rewritten with [`SqlQuery::remap_placeholders`] using this mapping.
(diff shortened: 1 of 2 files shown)


MazterQyou (approved):
LGTM

PR #11000, opened 2026-06-02
fix(tesseract): pre-aggregation table name corruption when one name is a prefix of another

waralexrom:
## Summary

`QueryCache.replacePreAggregationTableNames` replaced pre-aggregation placeholder table names sequentially with a global string replace per entry, which corrupts the SQL when one placeholder name is a prefix of another. With per-usage partition pruning (Tesseract), entries are suffixed `name0`..`name10`..., so 11+ usages of the same rollup in one query deterministically produce broken SQL like `(SELECT ...)0 AS "alias"` and fail on CubeStore with `ParserError("Expected: ), found: 0")`.

Sequential replacement is broken in any order: real target names are built as `tableName + '_' + versions` (see `PreAggregations.targetTableName`), so a source name also matches inside already inserted target names of other entries.

## Changes

- Replace all placeholder names in a single pass using one regex alternation (longest names first, since JS alternation matches left-to-right) with a `Map`-based replacement callback — inserted text is never rescanned
- Add unit tests: prefix collision with 11+ suffixed entries, source name inside an inserted target name, `QueryWithParams` passthrough, empty replacements, literal `$` in target names, input array immutability

## Testing

- `yarn jest test/unit/ReplacePreAggregationTableNames.test.ts` — 7 passed; the two collision tests fail against the previous implementation
- Reproduced live with Tesseract (`CUBEJS_TESSERACT_SQL_PLANNER=true`, `CUBEJS_TESSERACT_PRE_AGGREGATIONS=true`) on a query with 10+ multi-stage measures from one view (11 usages of the same rollup); fix verified against that setup

### DIFF
--- a/packages/cubejs-query-orchestrator/src/orchestrator/QueryCache.ts
+++ b/packages/cubejs-query-orchestrator/src/orchestrator/QueryCache.ts
@@ -426,10 +426,24 @@ export class QueryCache {
     const [keyQuery, params, queryOptions] = Array.isArray(queryAndParams)
       ? queryAndParams
       : [queryAndParams, []];
-    const replacedKeyQuery: string = preAggregationsTablesToTempTables.reduce(
-      (query, [tableName, { targetTableName }]) => QueryCache.replaceAll(tableName, targetTableName, query),
-      keyQuery
+    // Single-pass replacement with longest-first alternation: sequential
+    // per-name replacement would corrupt names that are prefixes of other
+    // names (e.g. `name1` vs `name10`) and rescan already inserted target
+    // names, which contain the source name as a prefix
+    const sorted = [...preAggregationsTablesToTempTables]
+      .sort(([a], [b]) => b.length - a.length);
+    const replacements = new Map(
+      sorted.map(([tableName, { targetTableName }]) => [tableName, targetTableName])
     );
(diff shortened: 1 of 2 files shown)


packages/cubejs-query-orchestrator/src/orchestrator/QueryCache.ts
```diff
@@ -426,10 +426,24 @@ export class QueryCache {
...
-    const replacedKeyQuery: string = preAggregationsTablesToTempTables.reduce(
-      (query, [tableName, { targetTableName }]) => QueryCache.replaceAll(tableName, targetTableName, query),
-      keyQuery
+    // Single-pass replacement with longest-first alternation: sequential
+    // per-name replacement would corrupt names that are prefixes of other
+    // names (e.g. `name1` vs `name10`) and rescan already inserted target
+    // names, which contain the source name as a prefix
+    const sorted = [...preAggregationsTablesToTempTables]
+      .sort(([a], [b]) => b.length - a.length);
+    const replacements = new Map(
```
ovr, line 435: @claude is it better to use `new Set` instead of `Map`?

ovr (approved):

paveltiunov (approved):

PR #11001, opened 2026-06-02
fix(pinot): support tesseract select template

### LINKED ISSUE #10988: Tesseract planner generates empty FROM clause for Pinot sql_table cubes
**Describe the bug**
When `CUBEJS_TESSERACT_SQL_PLANNER=true`, Pinot-backed cubes can generate invalid SQL with an empty FROM subquery:

```
FROM (

) AS
```

This appears to be caused by PinotQuery.sqlTemplates() overriding statements.select with a template that reads `{{ from }} / {{ from_alias }}`, while the Tesseract planner passes the source through `from_prepared`. The base query template handles `from_prepared`, but the Pinot override does not.

**To Reproduce**

1. Configure a Cube project with the Pinot driver.
2. Define a cube using sql_table.
3. Enable Tesseract with CUBEJS_TESSERACT_SQL_PLANNER=true.
4. Generate SQL for a query against the Pinot cube.
5. Observe that the generated SQL contains an empty FROM ( ) AS clause.

Example generated SQL shape:

```
SELECT
  "events".id "events__id",
  sum("events".amount) "events__total_amount"
FROM (

) AS
GROUP BY 1
```

**Expected behavior**
The generated SQL should include the Pinot table reference from sql_table, for example:

```
SELECT
  "events".id "events__id",
  sum("events".amount) "events__total_amount"
FROM events AS "events"
GROUP BY 1
```
More generally, Pinot’s statements.select template should support the same Tesseract-facing variables as the base select template, including from_prepared.

**Screenshots**
NA

In case your bug report is data modelling related please put your minimally reproducible Cube Schema here.
You can use selects without tables in order to achieve that as follows.

```javascript
cube(`Events`, {
  sql_table: `events`,

  measures: {
    totalAmount: {
      sql: `amount`,
      type: `sum`,
    },
  },

  dimensions: {
    id: {
      sql: `id`,
      type: `number`,
      primary_key: true,
    },
  },
});
```

**Version:**
Observed on Cube 1.6.52.

**Additional context**
The likely source is packages/cubejs-pinot-driver/src/PinotQuery.ts.

PinotQuery.sqlTemplates() currently overrides `templates.statements.select` and references `{{ from }}`:

```
templates.statements.select = 'SELECT ... \n' +
  'FROM (\n  {{ from }}\n) AS {{ from_alias }} \n' +
  ...
```

However, Tesseract renders select statements with `from_prepared`:

```
context! {
  from_prepared => from,
  ...
}
```

The base BaseQuery select template handles this via:

```
{% if from %}
FROM (
{{ from | indent(2, true) }}
) AS {{ from_alias }}{% elif from_prepared %}
FROM {{ from_prepared }}
{% endif %}
```

Pinot’s override likely needs to preserve that branch. It may also need to preserve other base-template variables used by Tesseract, such as ctes, distinct, joins, filter, and having.

seanm-stripe:
**Check List**

Fixes #10988

This updates the Pinot `statements.select` SQL template to support Tesseract planner inputs such as `from_prepared`.

Previously, Pinot overrode the base select template with an older template that only rendered `{{ from }}` and `{{ from_alias }}`. When `CUBEJS_TESSERACT_SQL_PLANNER=true`, Tesseract can pass the source table through `from_prepared`, which caused Pinot SQL for `sql_table` cubes to render an empty subquery:

```sql
FROM (

) AS
```

The Pinot template now preserves the base select template behavior for Tesseract-facing variables, including from_prepared, while *keeping Pinot’s dialect-specific OFFSET before LIMIT ordering* (which is why we can't just use BaseQuery)

This also adds regression coverage for a Pinot sql_table cube using the native SQL planner to verify that generated SQL renders the table source and does not emit an empty FROM () AS clause.

### DIFF
--- a/packages/cubejs-pinot-driver/src/PinotQuery.ts
+++ b/packages/cubejs-pinot-driver/src/PinotQuery.ts
@@ -145,12 +145,23 @@ export class PinotQuery extends BaseQuery {
     // epoch-millis representation produced by the timestamp_literal template
     templates.functions.UTCTIMESTAMP = 'NOW()';
     templates.functions.STRING_AGG = 'LISTAGG({% if distinct %}DISTINCT {% endif %}{{ args_concat }})';
-    templates.statements.select = 'SELECT {{ select_concat | map(attribute=\'aliased\') | join(\', \') }} \n' +
-      'FROM (\n  {{ from }}\n) AS {{ from_alias }} \n' +
-      '{% if group_by %} GROUP BY {{ group_by }}{% endif %}' +
-      '{% if order_by %} ORDER BY {{ order_by | map(attribute=\'expr\') | join(\', \') }}{% endif %}' +
-      '{% if offset %}\nOFFSET {{ offset }}{% endif %}' +
-      '{% if limit %}\nLIMIT {{ limit }}{% endif %}';
+    templates.statements.select = '{% if ctes %} WITH \n' +
+      '{{ ctes | join(\',\n\') }}\n' +
+      '{% endif %}' +
+      'SELECT {% if distinct %}DISTINCT {% endif %}' +
+      '{{ select_concat | map(attribute=\'aliased\') | join(\', \') }} {% if from %}\n' +
+      'FROM (\n' +
+      '{{ from | indent(2, true) }}\n' +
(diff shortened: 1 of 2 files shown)


ovr:
Thank you @seanm-stripe for your contribution 🍰 👍

Issue #10316, opened 2026-01-20
Invalid SQL syntax generated for clickhouse database on join keys when using tesseract


BlueMagma2:
**Describe the bug**
Invalid SQL syntax generated for clickhouse database when grouping by joined cubes
Cubejs generate join key of the form: (`table1`.`col1` = `table2`.`col2` OR ((`table1`.`col1` IS NULL) AND (`table2`.`col2` IS NULL))), trying to match keys having null values, 

Clickhouse doesn't support using "is null" in join keys

It also try to cast fields to `STRING` instead of `String` which also cause a syntax error on clickhouse (**EDITED 02/04/26**)

Clickhouse doesn't support using "is null" in join keys

**To Reproduce**
make a query involving a group by over a joined cube with clickhouse as the database backend

**Expected behavior**
Generate another equivalent syntax compatible with clickhouse, for exemple COALESCE(`table1`.`col1`, -1) = COALESCE(`table2`.`col2`, -1) works

**Version:**
cubejs:1.3, cubejs:1.6.7

***EDIT 02/04/26**:
**Relevant Configuration**
CUBEJS_TESSERACT_SQL_PLANNER: "true"  <= This gets fixed when tesseract is deactivated

BlueMagma2:
We also use tesseract, we've had a second syntax problem that was fixed when deactivating tesseract:

* With tesseract: CAST(field as STRING)
* Without tesseract: CAST(field as String) <= the correct syntax for clickhouse

It seems to me that tesseract doesn't properly take into account the specificity of the configured database

igorlukanin:
Reproduced the null-safe join-key issue on v1.6.7 with Tesseract: the multi-fact join condition is emitted as `(a = b OR (a IS NULL AND b IS NULL))`, which ClickHouse rejects in `JOIN ON`.

Fixed as of v1.6.24 — the ClickHouse dialect now emits `isNotDistinctFrom(a, b)` instead (commit 43429eb, #10494). Confirmed gone on latest v1.6.57.

Note: the separate `CAST(field as STRING)` vs `String` symptom @BlueMagma2 raised is not addressed by that fix and looks like a distinct casing bug — probably worth tracking separately.

igorlukanin:
The separate `CAST(field as STRING)` vs `String` casing @BlueMagma2 raised is a distinct, smaller bug (Tesseract doesn't override the ClickHouse `String` type template). Fix is up in #11053 and will resolve it once merged.

[...]

Issue #10415, opened 2026-02-19
ClickHouse: CAST(NULL, 'STRING') fails — ClickHouseQuery missing string type override


Fergus-Boydell:
## Bug

Queries using the new LAG/LEAD SQL pushdown (added in v1.6.14 via #10407) fail against ClickHouse with:

```
Error: Unknown data type family: STRING. Maybe you meant: ['String','Ring']
```

## Root Cause

`BaseQuery.js` defines the default type template as:

```js
types: {
  string: 'STRING',
  ...
}
```

`ClickHouseQuery.ts` overrides several types (`boolean`, `timestamp`, etc.) in `sqlTemplates()` but **does not override `string`**. ClickHouse type names are case-sensitive and require `String`, not `STRING`.

When LAG/LEAD pushdown generates the full SQL sent to ClickHouse, it includes internal columns:

```sql
CAST(NULL, 'STRING') AS __user, CAST(NULL, 'STRING') AS __cubejoinfield,
```

ClickHouse rejects `STRING` as an unknown type.

## Expected Behavior

The query should execute successfully against ClickHouse.

## Suggested Fix

Add the `string` type override in `ClickHouseQuery.ts` `sqlTemplates()`:

```typescript
templates.types.string = 'String';
```

## Environment

- Cube version: v1.6.14
- Database: ClickHouse
- API: SQL API (PostgreSQL wire protocol)

## Reproduction

Run any query using `LAG()` or `LEAD()` window functions via the Cube SQL API against a ClickHouse data source.

Fergus-Boydell:
**Update:** The fix needs to be `Nullable(String)`, not just `String`.

`CAST(NULL, 'String')` still fails with: `Cannot convert NULL to a non-nullable type`.

ClickHouse requires `Nullable(String)` to hold NULL values. The correct override should be:

```typescript
templates.types.string = 'Nullable(String)';
```

PR #11053, opened 2026-06-10
fix(schema-compiler): use ClickHouse String type in Tesseract casts (#10316)

igorlukanin:
## Issue
Fixes the ClickHouse string-cast casing bug reported in #10316.

When the Tesseract SQL planner (`CUBEJS_TESSERACT_SQL_PLANNER=true`) renders a string cast on ClickHouse — e.g. the composite-primary-key concatenation behind a `count` measure queried through a join — it emits `CAST(x AS STRING)`. `STRING` is not a valid ClickHouse type name (ClickHouse uses `String`, case-sensitive), so the query is rejected.

```sql
-- Tesseract (buggy):
count(CAST(order_id AS STRING) || CAST(product_id AS STRING))
-- Legacy planner (correct):
count(toString(CAST(`order_id` as String)) || toString(CAST(`product_id` as String)))
```

## Root cause
`BaseQuery` defines `types.string = 'STRING'`. The legacy ClickHouse path never hits it because `ClickHouseQuery.castToString()` is overridden to emit `String`. Tesseract, however, renders the `types.string` template directly, and `ClickHouseQuery.sqlTemplates()` overrides `boolean`/`timestamp` but not `string` — so the base `STRING` leaks through.

## Fix
Override `templates.types.string = 'String'` in `ClickHouseQuery.sqlTemplates()`, mirroring the existing `boolean`/`timestamp` overrides.

Note: #10316 also reported a null-safe `IS NULL` join-key bug, which was already fixed in v1.6.24 (#10494). This PR addresses the remaining string-cast-casing symptom.

### DIFF
--- a/packages/cubejs-schema-compiler/src/adapter/ClickHouseQuery.ts
+++ b/packages/cubejs-schema-compiler/src/adapter/ClickHouseQuery.ts
@@ -274,6 +274,10 @@ export class ClickHouseQuery extends BaseQuery {
     templates.quotes.escape = '\\`';
     templates.types.boolean = 'BOOL';
     templates.types.timestamp = 'DATETIME';
+    // ClickHouse's string type is `String` (case-sensitive); the base `STRING` is invalid.
+    // The legacy planner avoids this via castToString(), but Tesseract renders the types.string
+    // template directly, so it must be overridden here. See cube-js/cube#10316.
+    templates.types.string = 'String';
     delete templates.types.time;
     // ClickHouse intervals have a distinct type for each granularity
     delete templates.types.interval;


hughtopping:
Hi @igorlukanin - we've found that we get a different issue when `String` is added here. We're getting the following error from Clickhouse: `Cannot convert NULL to a non-nullable type`.

Using `Nullable(String)` rather than `String` fixes the error.

Here's a repro test case:

**Clickhouse**

```sql
CREATE TABLE buyer (id String, first_name String, last_name String, email_addresses Array(String)) ENGINE = MergeTree() ORDER BY id;
CREATE TABLE purchase (id String, buyer_id String) ENGINE = MergeTree() ORDER BY id;

INSERT INTO buyer VALUES ('1', 'Alice', 'Smith', ['alice@example.com', 'alice.smith@example.com']);
INSERT INTO purchase VALUES ('1', '1');
```

**Cube**

```yaml
# purchases.yml
cubes:
  - name: purchases
    sql: SELECT id, buyer_id FROM purchase
    joins:
      - name: buyers
        sql: "{purchases}.buyer_id = {buyers}.id"
        relationship: many_to_one
    dimensions:
      - name: id
        sql: id
        type: string
        primary_key: true
```

```yaml
# buyers.yml
cubes:
  - name: buyers
    sql: SELECT id, first_name, last_name FROM buyer
    joins:
      - name: purchases
        sql: "{buyers}.id = {purchases}.buyer_id"
        relationship: one_to_many
      - name: buyer_email_addresses
        sql: "{buyers}.id = {buyer_email_addresses}.buyer_id"
        relationship: one_to_many
    dimensions:
      - name: id
        sql: id
        type: string
        primary_key: true
      - name: first_name
        sql: first_name
        type: string
```

```yaml
# buyer_email_addresses.yml
cubes:
  - name: buyer_email_addresses
    sql: |
      SELECT id AS buyer_id, arrayJoin(email_addresses) AS email_address
      FROM buyer
    joins:
      - name: buyers
        sql: "{buyer_email_addresses}.buyer_id = {buyers}.id"
        relationship: many_to_one
    dimensions:
      - name: id
        sql: buyer_id
        type: string
        primary_key: true
        public: false
      - name: email_address
        sql: email_address
        type: string
    measures: []
```

**Cube SQL API query**

```sql
SELECT
    purchases.id,
    buyers.first_name,
    array_agg(buyer_email_addresses.email_address) AS email_addresses
FROM purchases
CROSS JOIN buyers
CROSS JOIN buyer_email_addresses
WHERE purchases.id = '1'
GROUP BY purchases.id, buyers.first_name
```

Edit: this was also reported by another user in https://github.com/cube-js/cube/issues/10415#issuecomment-3930474756

igorlukanin:
Confirmed, and thanks for the repro — it's a distinct bug from the one this PR fixes, not a regression from it.

`templates.types.string` is the shared cast-target type: `count.rs` renders it for composite-primary-key counts, and cubesql's push-down renders it for every untyped NULL literal (`DataType::Null` maps to `string` in `SqlTemplates::sql_type`). So `Nullable(String)` would fix your query but also change the type of any cast that lands in a pre-aggregation's projection — and ClickHouse rejects a `Nullable` column in a MergeTree `ORDER BY` unless `allow_nullable_key=1`, so pre-agg builds would start failing instead. The nullability belongs on the NULL-literal cast, not on the dialect's string type.

The NULL cast is covered separately in #10415.

PR #11097, opened 2026-06-16
fix(cubejs-client-core): surface cubeSql error chunks instead of dropping rows

keydunov:
## Problem

The cubesql endpoint streams JSONL — a schema line, then data lines — and on a post-processing failure appends a trailing `{ "error": "...", "requestId": "..." }` chunk. The non-streaming `cubeSql()` parser mapped every non-schema line through `JSON.parse(d).data`:

```js
const [schema, ...data] = response.error.split('\n');
data.filter(d => d.trim().length).map(d => JSON.parse(d).data).reduce((a,b) => a.concat(b), [])
```

For an `error` chunk, `JSON.parse(d).data` is `undefined`, so it gets concatenated as a **phantom `undefined` row** and the failure is silently swallowed — callers receive a "successful" result with a null row instead of an error. Downstream this surfaced as e.g. a workbook table crashing on the null row, with the actual cast-error message lost.

`cubeSqlStream()` already classifies `error` chunks correctly (it yields `{ type: 'error' }`); only the non-streaming `cubeSql()` was affected.

## Fix

Classify each JSONL line and **throw** on an `error` chunk, mirroring `cubeSqlStream()`. Genuine parse failures (malformed schema/data line) still fall back to throwing the raw response body, preserving existing behavior.

## Testing

Added a regression test to `CubeApi.test.ts` using the exact real-world payload (schema line followed by a post-processing cast-error chunk) asserting `cubeSql()` rejects with the clean message rather than resolving with a phantom row. All `cubeSql` tests pass.

### DIFF
--- a/packages/cubejs-client-core/src/index.ts
+++ b/packages/cubejs-client-core/src/index.ts
@@ -786,19 +786,49 @@ class CubeApi {
 
         const [schema, ...data] = response.error.split('\n');
 
+        let parsedSchema: any;
         try {
-          const parsedSchema = JSON.parse(schema);
-          return {
-            schema: parsedSchema.schema,
-            data: data
-              .filter((d: string) => d.trim().length)
-              .map((d: string) => JSON.parse(d).data)
-              .reduce((a: any, b: any) => a.concat(b), []),
-            ...(parsedSchema.lastRefreshTime ? { lastRefreshTime: parsedSchema.lastRefreshTime } : {}),
-          };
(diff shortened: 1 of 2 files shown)


keydunov:
Consumer-side fixes (ai-engineer inline parsers + console-ui table hardening) live in cubedevinc/cubejs-enterprise#12668.

vasilev-alex (approved):

PR #11137, opened 2026-06-22
fix(tesseract): Push ad-hoc filters into view measure aggregation

ovr:
A measure exposed through a view is a reference wrapper whose declared `type` is collapsed to `number` (a Calculated kind), hiding the real aggregation. PatchMeasure then rejected ad-hoc CASE-WHEN filters with "Unsupported additional filters for measure ... type number", because Calculated measures don't support additional filters.

Resolve the reference chain to the owning cube measure before patching, so the filter is pushed inside the real aggregation (`SUM(CASE WHEN ... END)`), matching the legacy planner's `patchMeasurePushDownFilterSql`. `resolve_reference_chain` is a no-op for plain cube measures, so non-view PatchMeasure behaviour is unchanged.

### DIFF
--- a/rust/cube/cubesqlplanner/cubesqlplanner/src/planner/query_properties_compiler.rs
+++ b/rust/cube/cubesqlplanner/cubesqlplanner/src/planner/query_properties_compiler.rs
@@ -272,7 +272,13 @@ impl QueryPropertiesCompiler {
                 }
                 let source_measure_compiled =
                     evaluator_compiler.add_measure_evaluator(source_measure.clone())?;
-                let symbol = if let Ok(source_measure) = source_measure_compiled.as_measure() {
+                // A view measure is a reference wrapper whose type collapses to
+                // `number` (Calculated), which rejects additional filters. Resolve
+                // the reference chain to the owning cube measure so the filter is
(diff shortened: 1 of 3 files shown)


rust/cube/cubesqlplanner/cubesqlplanner/src/planner/query_properties_compiler.rs
```diff
@@ -272,7 +272,13 @@ impl QueryPropertiesCompiler {
                 }
                 let source_measure_compiled =
                     evaluator_compiler.add_measure_evaluator(source_measure.clone())?;
-                let symbol = if let Ok(source_measure) = source_measure_compiled.as_measure() {
+                // A view measure is a reference wrapper whose type collapses to
+                // `number` (Calculated), which rejects additional filters. Resolve
+                // the reference chain to the owning cube measure so the filter is
+                // pushed inside the aggregation (`SUM(CASE WHEN ... END)`); a no-op
+                // for plain cube measures.
+                let resolved_source = source_measure_compiled.clone().resolve_reference_chain();
```
ovr, line 280: @claude please double verify that without this change, test test_many_to_one_view_patched_measure_filter fails. Please verify it carefully

waralexrom (approved):

PR #11144, opened 2026-06-23
perf(cubestore): worker-side group-by-limit hash-aggregate trim

waralexrom:
## Summary

Adds a worker-side **group-by-limit** optimization to CubeStore: for `GROUP BY <non-index-prefix> [ORDER BY <subset of group cols>] LIMIT k` queries, each worker bounds its partial hash aggregate to the top-k groups (`GroupByLimitAggregateExec`) instead of materializing every group. This bounds worker hash-table memory (OOM avoidance on high-cardinality group-bys) and cuts the rows crossing the network. All behavior is runtime-env-gated and **off by default**.

## Changes

- **`GroupByLimitAggregateExec`** (`queryplanner/group_by_limit_aggregate/`): a custom partial hash-aggregate exec that trims its output to the top-k groups by the total order `T = ORDER BY cols ++ remaining group cols`, built over DataFusion's `GroupValues` (mirrors the existing `InlineAggregateExec` pattern). Correct because top-k membership depends only on the group key, and `T` is applied at **both** the worker cut and the router's final sort/merge.
- **Rewriter + descriptor** (`group_by_limit_rewriter.rs`, `planning.rs`): plant the exec for the matching plan shape; extend limit pushdown to **bare `LIMIT`** (no `ORDER BY`) and **through `UNION`**. Correctness guards: bail on `HAVING`/filter-above-aggregate, skip global (no-GROUP-BY) aggregates, keep the custom exec's output partitioning following its input.
- **Worker delivery** (`distributed_partial_aggregate.rs`): inline/sorted aggregates stay bounded by a per-partition `Sort(fetch)`; hash aggregates use the trimming exec. `CUBESTORE_GROUP_BY_LIMIT_PER_PARTITION` toggles where the worker hash table lives — **over-merge** (default: one table per worker over coalesced partitions, lower memory) vs **under-merge** (per-partition, more parallelism).
- **Flags** (all default off): `CUBESTORE_GROUP_BY_LIMIT_FACTOR` (`0` = disabled), `CUBESTORE_GROUP_BY_LIMIT_PER_PARTITION`, `CUBESTORE_COALESCE_UNDER_HASH_AGGREGATE` (drop the useless sort-merge under a Linear hash aggregate).

> **Note — datafusion fork dependency.** `Cargo.toml` points the 4 datafusion crates at the fork branch `cubestore-hash-aggregate-limit` (= `cube-46.0.1` + a single `pub new_group_values` line needed by the exec). That one-line fork change must land before this merges; everything else the exec needs was already public.

## Testing

- Unit tests for `GroupByLimitAggregateExec` (partial/final, multi-partition, IPC roundtrip, global-aggregate skip, output-partitioning-follows-rechilded-input).
- In-process SQL: `planning_topk_hash_aggregate`, `topk_hash_aggregate_trim`, `limit_pushdown_group*` (incl. a new `limit_pushdown_group_null_appended` covering a NULL in an appended total-order column spanning chunks), `group_by_prefix_limit_high_cardinality`; cluster test `worker_sort_and_limit_cluster`.
- Full in-process SQL suite (181) and `cubestore` lib suite (290) green with all optimization envs enabled.
- Benchmarked vs a single-process ClickHouse on 144M rows: on the UNION hash-trim shape CubeStore reaches ClickHouse parity (~1.46s vs ~1.45s) at ~3–4× lower memory, where the unbounded baseline OOMs.


ovr (commented):
LGTM

PR #11151, opened 2026-06-23
test(testing-drivers): isolate pre-aggregations schema per run to fix shared-cloud races

paveltiunov:
Parallel CI legs (use_tesseract_sql_planner false/true) for the same
shared-cloud database pointed at the same hardcoded pre-aggregations
schema, so they raced building, refreshing, and dropping the same
pre-agg tables non-deterministically (confirmed on BigQuery; athena and
databricks-jdbc have the identical latent race).

Make the schema overridable via docker-compose interpolation, defaulting
to each fixture's existing value so local/other runs are unaffected. The
companion cube-runtime workflow sets
DRIVERS_TESTS_CUBEJS_PRE_AGGREGATIONS_SCHEMA per matrix leg so each
concurrent leg gets its own pre-aggregations schema.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_015MhpNg5eCmC34941RWYvnw


ovr:
@claude Question, does anyone setup `DRIVERS_TESTS_CUBEJS_PRE_AGGREGATIONS_SCHEMA` on CI? Could you point me to exact place where it's done

PR #11165, opened 2026-06-26
feat(duckdb-driver): migrate to @duckdb/node-api to enable DuckDB 1.5+

samjewell:
The legacy `duckdb` npm package tops out at 1.4.x, which prevents loading extensions that require DuckDB >= 1.5.2 (such as quack). Migrate the driver to @duckdb/node-api (the Neo bindings), which wraps released DuckDB 1.5.x binaries and is the actively maintained client.

- Replace Database/Connection (callback API) with DuckDBInstance/ DuckDBConnection (promise API); preserve MotherDuck, S3, extension install/load, initSql and schema introspection behavior.
- query() uses runAndReadAll + getRowObjectsJS; stream() consumes the streaming result via yieldRowObjectJs. The JS value conversion keeps HydrationStream output identical (numbers/bigints -> string, Date -> ISO).
- Drop the stale `rm -rf node_modules/duckdb/src` step from the Docker images: node-api ships prebuilt platform binaries with no bundled C++ sources to strip.

**Check List**


npwalker:
Reviewed this migration for our own adoption (we run a thin subclass of `@cubejs-backend/duckdb-driver`). The API-contract usage looks correct — sync `getRowObjectsJS()` on the reader, lazy `yieldRowObjectJs()` chunk-flatten for `stream()`, Date binding via `timestampMillisValue`. Structurally solid.

I did find three value-conversion behavior changes vs the legacy `duckdb` npm driver, all rooted in the fact that `HydrationStream.transformRow` only normalizes **top-level** `number`/`bigint`/`Date`. Two of them I verified end-to-end against real `@duckdb/node-api@1.5.4-r.1` vs `duckdb@1.4.1`.

### 1. INTERVAL / TIME_TZ → unserializable (hard crash)

`objectFromIntervalValue` returns `{ months, days, micros }` where `micros` is now a **bigint** (legacy returned a plain `number`). `transformRow` doesn't recurse into the object, so the nested bigint survives to `JSON.stringify`, which throws.

```
SELECT INTERVAL '1 month 2 days 3 hours' AS v
```
- legacy `duckdb@1.4.1`: `{"months":1,"days":2,"micros":10800000000}` (serializes fine)
- `@duckdb/node-api@1.5.4-r.1`: `TypeError: Do not know how to serialize a BigInt`

TIME_TZ has the same shape (`{ micros: <bigint>, offset }`) and the same failure.

### 2. TIME → raw microseconds

`bigintFromTimeValue` returns micros-since-midnight as a bigint; `transformRow`'s bigint branch stringifies the raw count.

```
SELECT TIME '04:05:06' AS v
```
- legacy: `"04:05:06"`
- node-api: `"14706000000"`

### 3. `stream()` no longer tolerates `null` values

`normalizeValues = (values = []) => values.map(...)` — a default parameter only substitutes for `undefined`, so `stream(query, null)` hits `null.map(...)` → `TypeError`. The old code was `connection.stream(query, ...(values || []))`, which tolerated `null`. `query()` is unaffected (its `values` has a `= []` default). This is a defensive regression — I haven't confirmed a caller that passes `null` in practice, but the tolerance changed.

A recursive `transformRow` (descend into plain objects/arrays, convert nested bigint, and format INTERVAL/TIME) plus `normalizeValues(values = [])` also guarding `null` would cover all three.

For what it's worth, two things that look scary from the type decls turned out to be **non-issues** at runtime, in case they come up in review:
- **DECIMAL** precision is unchanged — both legacy and node-api coerce DECIMAL to a JS double identically, so no new precision loss.
- **BLOB** — despite the `Uint8Array` type in `JS.d.ts`, node-api returns a Node `Buffer` at runtime, so `JSON.stringify` output is identical to legacy (`{"type":"Buffer","data":[...]}`).

Repro (both bindings side by side):
```js
import { DuckDBInstance } from '@duckdb/node-api'; // 1.5.4-r.1
const inst = await DuckDBInstance.create(':memory:');
const conn = await inst.connect();
const reader = await conn.runAndReadAll("SELECT INTERVAL '1 month 2 days 3 hours' AS v");
const row = reader.getRowObjectsJS()[0];
console.log(typeof row.v.micros);        // "bigint"
JSON.stringify(row);                      // throws: Do not know how to serialize a BigInt
```

ovr:
Thank you @samjewell for your contribution! 👍 🍰 I've picked up your changes and finished work in https://github.com/cube-js/cube/pull/11910.

I hope to make a new release later today that includes this change.

PR #11166, opened 2026-06-26
fix(cubestore): expose CUBESTORE_CACHESTORE_WAL_TTL_SECONDS to bound the cachestore WAL

### LINKED ISSUE #11168: cubestore: cachestore WAL archive grows unbounded (no ttl config), filling the data volume
**Describe the bug**

When Cube Store is the cache + queue backend (`CUBEJS_CACHE_AND_QUEUE_DRIVER=cubestore`), the cachestore RocksDB keeps archived write-ahead-log (WAL) files for a retention time that is hardcoded (`meta_store_snapshot_interval + meta_store_log_upload_interval`, ~330s). Under sustained high-cardinality query churn the archived WAL under `cachestore/archive/` accumulates and is not reclaimed sooner, growing into the multi-GB range. On Kubernetes with an `emptyDir` `sizeLimit` the volume fills and the kubelet evicts the pod, dropping in-flight requests.

There is no env var to tune the cachestore WAL retention, and `CUBESTORE_MAX_DISK_SPACE_GB` does not bound it (it caps pre-aggregation data, not the cachestore). Cachestore log upload -- the only consumer of the retained WAL (`get_updates_since`) -- is off by default (`CUBESTORE_CACHESTORE_LOG_ENABLED=false`), so for single-instance deployments the retained WAL is pure overhead.

**To Reproduce**

1. Run a single-instance cubestore as the cache+queue backend (`CUBEJS_CACHE_AND_QUEUE_DRIVER=cubestore`), Filesystem store provider, on a size-bounded volume (e.g. a Kubernetes `emptyDir` with `sizeLimit`).
2. Drive sustained, high-cardinality `/load` queries so each produces a distinct large cache entry.
3. Watch `du -sh <data-dir>/cachestore/archive` -- it climbs to ~the retention window's worth of WAL (multi-GB) and stays high; eventually the volume fills and the pod is evicted.

**Expected behavior**

The cachestore WAL retention should be configurable (or otherwise bounded) so single-instance / no-log-upload deployments can keep the on-disk WAL small and avoid filling the data volume.

**Version:**

cubestore on RocksDB 7.10.2; reproduced on cube v1.6.x.

**Additional context**

The WAL ttl exists for `get_updates_since` (log upload). RocksDB purges archived WAL older than the ttl on a `wal_ttl/2` cycle, so a lower ttl yields a lower on-disk plateau. Note `WAL_size_limit_MB` is not a fit on this RocksDB version: the fast `wal_ttl/2` purge cadence only runs when the size limit is unset, so enabling it alongside the ttl drops the cadence to the 10-minute default and keeps more WAL.

Proposed fix: #11166 (exposes `CUBESTORE_CACHESTORE_WAL_TTL_SECONDS`, default unchanged).

killzoner:
**Check List**
- [x] Tests have been run in packages where changes have been made if available (`cargo test -p cubestore cachestore` + the new config tests pass)
- [x] Linter has been run for changed code (`cargo fmt --check` clean; `cargo clippy` shows no new issues in the changed files)
- [x] Tests for the changes have been added if not covered yet (unit tests for the WAL ttl min-clamp validation)
- [x] Docs have been added / updated if required (added `CUBESTORE_CACHESTORE_WAL_TTL_SECONDS` to the environment variables reference)

Fixes #11168

The cachestore's RocksDB keeps old write-ahead-log (WAL) files for a retention time that is currently hardcoded (~330s).

- Under sustained write load these archived WAL files pile up and can fill the data volume (e.g. a Kubernetes `emptyDir`), which gets the pod evicted.
- This adds `CUBESTORE_CACHESTORE_WAL_TTL_SECONDS` to make that retention configurable:
  - **Unset** -> same value as today, so behavior is unchanged.
  - **Lower** -> RocksDB drops archived WAL sooner, keeping it smaller on disk. Useful when cachestore log upload is off (the default) -- that is the only thing that needs the WAL kept.
  - Minimum **2s** (`0` would turn WAL cleanup off entirely).
- `WAL_size_limit_MB` was deliberately not used: enabling it alongside the ttl makes RocksDB switch from the fast cleanup cycle to a 10-minute one, so it keeps *more* WAL, not less.

### DIFF
--- a/docs/content/product/configuration/reference/environment-variables.mdx
+++ b/docs/content/product/configuration/reference/environment-variables.mdx
@@ -1553,6 +1553,19 @@ The address/port pair for Cube Store's MySQL-compatible interface.
 | ------------------------- | ---------------------- | --------------------- |
 | A valid address/port pair | `0.0.0.0:3306`         | `0.0.0.0:3306`        |
 
+## `CUBESTORE_CACHESTORE_WAL_TTL_SECONDS`
+
+How long, in seconds, Cube Store keeps an archived cachestore write-ahead log
+(WAL) file before deleting it; RocksDB checks on a `ttl / 2` cycle. Lower values
+keep less WAL on disk under sustained write load. When unset, it defaults to the
+cachestore snapshot interval plus the log-upload interval. Lower it only when
+cachestore log upload is disabled (the default), since that is the only consumer
+of the retained WAL. Minimum `2`.
+
+| Possible Values | Default in Development | Default in Production |
+| --------------- | ---------------------- | --------------------- |
+| A valid number  | N/A                    | N/A                   |
+
 ## `CUBESTORE_DATA_DIR`
 
 A path on the local filesystem to store a local replica of the data. Must be
--- a/rust/cubestore/cubestore/src/cachestore/cache_rocksstore.rs
+++ b/rust/cubestore/cubestore/src/cachestore/cache_rocksstore.rs
@@ -96,10 +96,13 @@ impl RocksStoreDetails for RocksCacheStoreDetails {
         opts.set_compaction_filter_factory(compaction::MetaStoreCacheCompactionFactory::new(
             compaction_state,
         ));
-        // TODO(ovr): Decrease after additional fix for get_updates_since
-        opts.set_wal_ttl_seconds(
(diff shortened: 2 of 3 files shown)


killzoner:
From what I've seen in the commits, probably @ovr would be best to have a look when you have time. Thanks in advance

PR #11167, opened 2026-06-26
feat(api-gateway): support relative date ranges in OR/AND filter groups

elijah-evans-deel:
## Summary

- Date operators (`inDateRange`, `notInDateRange`, `beforeDate`, `beforeOrOnDate`, `afterDate`, `afterOrOnDate`, `onTheDate`) inside `filters.or` / `filters.and` groups now accept relative-date strings (e.g. `"last 2 weeks"`, `"today"`, `"from 7 days ago to now"`) in `values`, using the same parser as `timeDimensions.dateRange`.
- Date filter leaves now emit SQL inside their parent logical group, so users can express constraints like `(created_at IN [last 2 weeks]) OR (status = 'pending')` which was previously impossible because relative strings reached `BaseFilter` as literals and the only relative-date-aware path was the global-AND `timeDimensions.dateRange`.
- Top-level `timeDimensions.dateRange` behavior is unchanged. Both paths share the same resolver (`resolveDateRange`) so they cannot drift.
- Because both paths still emit SQL conditions, specifying a date constraint in both timeDimensions and filters produces two predicates that are AND-ed together. This would be redundant if identical, but surprising if different. The new doc warning covers this; see the "Date filters in logical operators" section.
- I did this on purpose, otherwise we'd create a breaking change to existing queries using timeDimensiosn.dateRange

## Motivation

A user filtering for *"orders in the last 2 weeks OR orders with status = pending"* could not express that query before this PR. The only date-range mechanism that supported relative strings was `timeDimensions.dateRange`, which is always emitted as a global `AND` and cannot be scoped to one branch of an `or` group. Putting `{ "operator": "inDateRange", "values": ["last 2 weeks"] }` inside `filters.or` reached `BaseFilter` as a literal string and produced a runtime error.

## Approach

Resolution lives in `normalizeQuery` at the gateway, co-located with the existing `timeDimensions.dateRange` resolution. A small `normalizeDateFilterValues` helper dispatches by operator family (range vs single-date) and delegates to the existing `resolveDateRange` resolver. The recursive filter walker (`normalizeQueryFilters`) calls the helper at every leaf, so OR/AND-nested date filters get the same treatment as top-level ones.

## Out of scope (intentionally)

- **`timeDimensions.dateRange` and a date filter on the same member.** Both predicates are emitted and `AND`-ed (existing `BaseQuery.allFilters` behavior); this is documented as a footgun in the new doc warning but not enforced. Deciding whether to reject the combination is a separate API design conversation.
- **Tesseract (Rust planner).** No changes. My understanding is that because we're resolving at gateway, Tesseract will hanlde the OR-with-resolved-date-range case correctly without any change.

## Files changed

| File | Change |
|---|---|
| `packages/cubejs-api-gateway/src/query.js` | Added `normalizeDateFilterValues` helper; threaded `timezone` into `normalizeQueryFilters`; called helper at every leaf. |
| `packages/cubejs-api-gateway/test/normalize-query-filters-dates.test.js` | New tests for the helper, the wiring through `normalizeQuery`, OR/AND/nested groups, timezone handling, error surface. |
| `packages/cubejs-schema-compiler/test/unit/or-date-filter.test.ts` | New end-to-end SQL tests via `PostgresQuery.buildSqlAndParams()` covering OR groups, AND groups, deeply nested groups, two independent date filters in one OR, `notInDateRange`, regression guard for top-level `timeDimensions`. |
| `docs-mintlify/reference/core-data-apis/rest-api/query-format.mdx` | Documented relative-date support in date operators, added a "Date filters in logical operators" section with a `<Warning>` explaining the difference vs. `timeDimensions.dateRange`. |

## Test plan

- [x] `cd packages/cubejs-schema-compiler && yarn tsc && TZ=UTC yarn jest --coverage=false dist/test/unit/or-date-filter.test.js` — 8/8 passing.
- [x] `cd packages/cubejs-api-gateway && yarn jest --coverage=false --forceExit dist/test/normalize-query-filters-dates.test.js dist/test/normalize-query.test.js dist/test/date-parser.test.js` — 43/43 passing.
- [x] Full schema-compiler unit suite — 687/689 passing (2 unrelated failures in `error-reporter.test.ts` from a `@babel/code-frame` snapshot drift).
- [x] Full gateway test suite — 197/216 passing (19 unrelated failures in `index.test.js`/`auth.test.js` from a pre-existing `@cubejs-backend/native` ABI mismatch that also blocks the gateway TS build).
- [ ] Manual verification against a running Cube instance with a real Postgres source.
- [ ] Verify docs render correctly in Mintlify preview (`cd docs-mintlify && yarn dev`).


packages/cubejs-api-gateway/src/query.js
```diff
@@ -244,15 +244,62 @@ export const preAggsJobsRequestSchema = Joi.object({
...
+  }
+
+  if (DATE_RANGE_OPERATORS.includes(filter.operator)) {
+    return { ...filter, values: resolveDateRange(filter.values[0], timezone) };
+  }
+
+  if (SINGLE_DATE_OPERATORS.includes(filter.operator)) {
+    const [start] = resolveDateRange(filter.values[0], timezone);
+    return { ...filter, values: [start] };
+  }
```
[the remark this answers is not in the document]
ovr: @elijah-evans-deel I find the reviewer bot's reasoning correct. As initial step, you should move your postgresql tests from unit to integration to test this new functionality with both planner: legacy and tesseract.

In tesseract there are some difference with date comparing + `onTheDate` requires data range instead of single value. That's why it's important to use integration test (you can move it from`unit` to `integration` folder in `schema-compiler`)

ovr: I agree with the bot's finding 1, but I would raise its severity: taking `[start]` for all five operators bakes in the legacy semantics that Tesseract (upcoming default planner) has already intentionally fixed. In `physical_plan/filter/operators/date_single.rs` Tesseract uses:

- `beforeDate` → `< startOfDay`
- `beforeOrOnDate` → `<= endOfDay`
- `afterDate` → `> endOfDay`
- `afterOrOnDate` → `>= startOfDay`

The gateway should dispatch the same way — then the results converge in both planners, because both pass a full timestamp through `format_to_date`/`formatFromDate` as-is:

```js
const START_DATE_OPERATORS = ['beforeDate', 'afterOrOnDate'];
const END_DATE_OPERATORS = ['beforeOrOnDate', 'afterDate'];

if (START_DATE_OPERATORS.includes(filter.operator)) {
  const [start] = resolveDateRange(value, timezone);
  return { ...filter, values: [start] };
}

if (END_DATE_OPERATORS.includes(filter.operator)) {
  const [, end] = resolveDateRange(value, timezone);
  return { ...filter, values: [end] };
}
```

What this changes for `["yesterday"]` (today is 2026-07-02, tz UTC):

| Operator | Current PR (`[start]`) | With dispatch |
|---|---|---|
| `beforeDate` | `< 2026-07-01T00:00:00.000` ✓ | unchanged |
| `beforeOrOnDate` | `<= 2026-07-01T00:00:00.000` — all of yesterday excluded | `<= 2026-07-01T23:59:59.999` |
| `afterDate` | `> 2026-07-01T00:00:00.000` — "after yesterday" includes yesterday | `> 2026-07-01T23:59:59.999` |
| `afterOrOnDate` | `>= 2026-07-01T00:00:00.000` ✓ | unchanged |

packages/cubejs-api-gateway/src/query.js
```diff
@@ -244,15 +244,62 @@ export const preAggsJobsRequestSchema = Joi.object({
...
+  if (!filter || !filter.operator || !Array.isArray(filter.values) || filter.values.length !== 1) {
+    return filter;
+  }
+
+  if (DATE_RANGE_OPERATORS.includes(filter.operator)) {
+    return { ...filter, values: resolveDateRange(filter.values[0], timezone) };
+  }
+
+  if (SINGLE_DATE_OPERATORS.includes(filter.operator)) {
+    const [start] = resolveDateRange(filter.values[0], timezone);
```
ovr: There are two regressions here for **existing** queries with absolute values — only relative strings should be resolved.

**(a) Bare dates under Tesseract — silent change of query results.**
Before this PR the gateway didn't touch filter values; now it rewrites them. For `beforeOrOnDate: ["2024-01-15"]` under Tesseract:

```sql
-- before the PR: the value arrives as a bare date, format_to_date pads it to end of day
WHERE created_at <= '2024-01-15T23:59:59.999'   -- Jan 15 included

-- after the PR: the gateway resolves it to "2024-01-15T00:00:00.000", format_to_date passes it as-is
WHERE created_at <= '2024-01-15T00:00:00.000'   -- Jan 15 excluded
```

Mirrored for `afterDate`. This is a data-correctness regression under the default planner — more severe than the "low" in the bot's finding 2 (which evaluated legacy only).

**(b) Precise timestamps get truncated to day granularity (both planners).**
Every string now goes through `dateParser` → chrono with `exactGranularity = 'day'` (`date-parser.js:99`) → `startOf('day')`:

```sql
-- afterDate: ["2024-01-15T10:30:00.000"]
-- before the PR: the value passes through verbatim
WHERE created_at > '2024-01-15T10:30:00.000'

-- after the PR: chrono stripped the time component
WHERE created_at > '2024-01-15T00:00:00.000'
```

(or a `UserError` if chrono fails to parse the string). The `DateRegex` short-circuit suggested by the bot doesn't cover this — that regex only matches `YYYY-MM-DD`.

The fix is a guard on already-absolute values before resolving:

```js
// Covers bare dates and ISO datetimes (with optional time / ms)
const AbsoluteDateTimeRegex = /^\d{4}-\d{2}-\d{2}([T ]\d{2}:\d{2}(:\d{2}(\.\d{1,6})?)?)?$/;

if (AbsoluteDateTimeRegex.test(value)) {
  return filter; // pass through byte-exact
}
```

The guard is only needed for the comparison operators (`beforeDate`/`beforeOrOnDate`/`afterDate`/`afterOrOnDate`). For `inDateRange`/`notInDateRange`/`onTheDate` with a single value, resolving an absolute date into `[start, end]` is a fix, not a regression: before this PR a single value was broken for them in both planners (see the `onTheDate` comment).

packages/cubejs-api-gateway/src/query.js
```diff
@@ -244,15 +244,62 @@ export const preAggsJobsRequestSchema = Joi.object({
 
 const DateRegex = /^\d\d\d\d-\d\d-\d\d$/;
 
-const normalizeQueryFilters = (filter) => (
+const DATE_RANGE_OPERATORS = ['inDateRange', 'notInDateRange'];
+const SINGLE_DATE_OPERATORS = ['onTheDate', 'beforeDate', 'beforeOrOnDate', 'afterDate', 'afterOrOnDate'];
```
ovr: A clarification to the bot's finding 1: the "collapses to a single instant" mechanics are described inaccurately, but the conclusion is right — and there is an opportunity to fix the operator while we're here.

Actual behavior of `onTheDate` with a single value:
- **Legacy**: `onTheDateWhere` reads `values[1]`, which doesn't exist → `formatToDate(undefined)` returns the end of the *current* day → the range is "from the start of the given date to the end of today". Broken before this PR too.
- **Tesseract**: `onTheDate` maps to `InDateRange` (`filter_operator.rs:47`), which strictly requires 2 values (`typed_filter.rs:253`) → error `2 arguments expected for date range`. Also broken before this PR.

If the gateway resolves `onTheDate` into a two-element `values`, the operator will work correctly in **both** planners for the first time — legacy `onTheDateWhere` uses exactly `values[0]`/`values[1]`, and Tesseract needs exactly two arguments:

```js
if (filter.operator === 'onTheDate') {
  // "yesterday" → ["2026-07-01T00:00:00.000", "2026-07-01T23:59:59.999"]
  return { ...filter, values: resolveDateRange(value, timezone) };
}
```

```sql
-- onTheDate: ["yesterday"] with this fix (both planners):
WHERE created_at >= '2026-07-01T00:00:00.000'
  AND created_at <= '2026-07-01T23:59:59.999'

-- current PR: values = ["2026-07-01T00:00:00.000"] →
--   legacy:    BETWEEN start of yesterday AND end of TODAY (matches both yesterday and today)
--   Tesseract: error "2 arguments expected for date range"
```

ovr (commented):
The bot review is directionally correct, but it only analyzes the legacy JS planner. On the v1.7.0 branch the default planner is Tesseract, and that changes the priorities: finding 1 becomes a must-fix, finding 2 becomes a blocker, and there is one more missed regression with precise timestamps. Details are in the inline comments; here is the summary and a cohesive patch.

Principles of the fix:

1. **Don't resolve already-absolute values** (a bare `YYYY-MM-DD` date or an ISO timestamp) for the comparison operators — pass them through byte-exact. `dateParser` should only run for genuinely relative strings. This keeps the PR purely additive: only values that previously errored out get new behavior.
2. For relative strings — **per-operator start/end dispatch**: `beforeDate`/`afterOrOnDate` → start, `beforeOrOnDate`/`afterDate` → end. This is exactly Tesseract's semantics (`rust/cube/cubesqlplanner/.../physical_plan/filter/operators/date_single.rs`), already accepted as intentional for 1.7.
3. **Resolve `onTheDate` to two values `[start, end]`** — this makes the operator work correctly in both planners for the first time (see the inline comment on `onTheDate`).
4. Optional: a clean `UserError` for `inDateRange: ["last week", "this week"]` (two relative values) instead of failing deep inside query execution.

Proposed patch for `normalizeDateFilterValues` (query.js):

```diff
 const DATE_RANGE_OPERATORS = ['inDateRange', 'notInDateRange'];
-const SINGLE_DATE_OPERATORS = ['onTheDate', 'beforeDate', 'beforeOrOnDate', 'afterDate', 'afterOrOnDate'];
+// Mirrors Tesseract's date_single.rs boundary semantics:
+// Before → < start, AfterOrOn → >= start
+const START_DATE_OPERATORS = ['beforeDate', 'afterOrOnDate'];
+// BeforeOrOn → <= end, After → > end
+const END_DATE_OPERATORS = ['beforeOrOnDate', 'afterDate'];
+
+// Absolute values must pass through byte-exact: bare dates keep each planner's
+// own day-boundary handling, timestamps keep their time component (dateParser
+// would truncate them to day granularity).
+const AbsoluteDateTimeRegex = /^\d{4}-\d{2}-\d{2}([T ]\d{2}:\d{2}(:\d{2}(\.\d{1,6})?)?)?$/;

 export const normalizeDateFilterValues = (filter, timezone) => {
   if (!filter || !filter.operator || !Array.isArray(filter.values) || filter.values.length !== 1) {
     return filter;
   }

-  if (DATE_RANGE_OPERATORS.includes(filter.operator)) {
-    return { ...filter, values: resolveDateRange(filter.values[0], timezone) };
+  const value = filter.values[0];
+  if (typeof value !== 'string') {
+    return filter;
+  }
+
+  if (DATE_RANGE_OPERATORS.includes(filter.operator) || filter.operator === 'onTheDate') {
+    // onTheDate resolves to a two-sided range: legacy onTheDateWhere reads
+    // values[0]/values[1], and Tesseract maps onTheDate to InDateRange which
+    // requires exactly 2 values.
+    return { ...filter, values: resolveDateRange(value, timezone) };
   }

-  if (SINGLE_DATE_OPERATORS.includes(filter.operator)) {
-    const [start] = resolveDateRange(filter.values[0], timezone);
+  if (AbsoluteDateTimeRegex.test(value)) {
+    return filter;
+  }
+
+  if (START_DATE_OPERATORS.includes(filter.operator)) {
+    const [start] = resolveDateRange(value, timezone);
     return { ...filter, values: [start] };
   }

+  if (END_DATE_OPERATORS.includes(filter.operator)) {
+    const [, end] = resolveDateRange(value, timezone);
+    return { ...filter, values: [end] };
+  }
+
   return filter;
 };
```

The tests in `normalize-query-filters-dates.test.js` and `or-date-filter.test.ts` will need to be updated for the new boundaries (end-of-range for `beforeOrOnDate`/`afterDate`, two values for `onTheDate`, pass-through for absolute values).

Separately, regarding "Tesseract will handle the OR-with-resolved-date-range case correctly without any change" from the PR description: this is true for `inDateRange` (two resolved values), but not for the single-date operators — see the inline comments.

waralexrom (approved):

ovr:
Merged, I will make a release later today.

Thank you, @elijah-evans-deel, for your contribution! 🍰 🥮

PR #11169, opened 2026-06-26
fix(tesseract): honor rolling window offset without granularity

### LINKED ISSUE #11162: `rollingWindow` with `trailing: 'unbounded'` + `offset: 'start'` incorrectly filters by `end_date` instead of `start_date` under Tesseract
When Tesseract SQL Planner is enabled (`CUBEJS_TESSERACT_SQL_PLANNER=true`), measures using `rollingWindow: { trailing: 'unbounded', offset: 'start' }` produce the **same** `WHERE` clause as `offset: 'end'` measures — both use `<= end_date` instead of the expected `< start_date`.

- `offset: 'end'` **(correct)**: `billdate <= end_date`
- `offset: 'start'` **(wrong)**: also generates `billdate <= end_date`, should be `billdate < start_date`

This means beginning-of-period calculations return incorrect results when Tesseract is active.

### 1. Create a self-contained cube (copy-paste ready, no external tables needed):

```javascript
cube(`test_cube`, {
  sql: `
    SELECT '2025-12-01'::timestamptz AS billdate, 10 AS amount
    UNION ALL
    SELECT '2026-02-01'::timestamptz AS billdate, 20 AS amount
  `,

  dimensions: {
    billdate: { sql: `billdate`, type: `time` },
  },

  measures: {
    beg_balance: {
      sql: `amount`,
      type: `sum`,
      rollingWindow: { trailing: `unbounded`, offset: `start` },
    },
    end_balance: {
      sql: `amount`,
      type: `sum`,
      rollingWindow: { trailing: `unbounded`, offset: `end` },
    },
  },
});
```

### 2. Enable Tesseract:

```bash
export CUBEJS_TESSERACT_SQL_PLANNER=true
```

### 3. Query both measures via `/cubejs-api/v1/sql`:

```json
{
  "measures": ["test_cube.beg_balance", "test_cube.end_balance"],
  "timeDimensions": [{
    "dimension": "test_cube.billdate",
    "dateRange": ["2026-01-01 00:00:00", "2026-06-17 23:59:59"]
  }]
}
```

## Actual Behavior

Both measures produce **identical** `WHERE` clauses:

```sql
-- end_balance (offset: 'end') — correct
WHERE ("test_cube"."billdate" <= $1::timestamptz)
-- $1 = 2026-06-17T23:59:59.000Z  (end of date range)

-- beg_balance (offset: 'start') — WRONG, same as above!
WHERE ("test_cube"."billdate" <= $1::timestamptz)
-- $1 = 2026-06-17T23:59:59.000Z  (should be start of date range!)
```

Both return `SUM(amount) = 30` (all rows included).

## Expected Behavior

```sql
-- end_balance (offset: 'end') — billdate <= end_date ✓
WHERE ("test_cube"."billdate" <= $1::timestamptz)
-- $1 = 2026-06-17T23:59:59.000Z → SUM(amount) = 30

-- beg_balance (offset: 'start') — billdate < start_date
WHERE ("test_cube"."billdate" < $1::timestamptz)
-- $1 = 2026-01-01T00:00:00.000Z → SUM(amount) = 10
```

(Only the `2025-12-01` row is before the range start, so `SUM(amount) = 10`.)

## Additional Notes

- When the time dimension **has** a granularity (e.g., `"granularity": "day"`), `offset: 'start'` works correctly — the issue only appears when **no granularity** is specified.
- Disabling Tesseract (`CUBEJS_TESSERACT_SQL_PLANNER=false`) makes the bug disappear: the JS-based SQL planner generates the correct `WHERE` clause for both offsets.

## Environment

- Cube.js: **1.6.58**
- `@cubejs-backend/native`: **1.6.58**
- `@cubejs-backend/schema-compiler`: **1.6.58**
- OS: macOS
- Database: PostgreSQL

waralexrom:
## Summary

Fixes #11162. Under the Tesseract SQL planner, rolling window measures queried with only a `dateRange` (no granularity) ignored the `offset` and dropped finite trailing/leading intervals, producing wrong results — e.g. a `trailing: unbounded, offset: start` "beginning balance" measure returned the same value as `offset: end`.

## Why

Without granularity there is no `time_series` CTE, so the window is emulated by rewriting the `InDateRange` filter on the base measure. That rewrite only handled `unbounded` sides and discarded `offset` entirely:

- `trailing: unbounded` + `offset: start` filtered by `<= to` instead of `< from`
- `leading: unbounded` + `offset: end` filtered by `>= from` instead of `> to`
- finite `trailing`/`leading` without granularity fell back to the raw date range (interval ignored)

The correct semantics are taken from the authoritative with-granularity path (`physical_plan/join.rs`): treat the whole range as a single window anchored by `offset` (`start` → range start `from`, `end` → range end `to`); the trailing side is the lower bound, the leading side the upper bound; an `unbounded` side drops its bound, a finite interval shifts it.

## Changes

- New synthetic `RollingWindowOffsetDateRange` filter operator + `RollingWindowOffsetOp` renderer that emits the offset-anchored bounds over the literal date range (reuses `extend_date_range_bound`).
- `replace_date_range_for_rolling_window_without_granularity` now threads `offset` and rewrites via the shared `change_date_range_filter_impl` (recurses into filter groups, single pass).
- `FILTER_PARAMS` callbacks on the time dimension now receive the `from`/`to` dates, never the window config (`unbounded`/`offset`).
- User-facing `beforeDate`/`afterDate` operators left untouched (no value overloading).
- Behavior change: bare `leading: unbounded` without an explicit `offset` (no granularity) now anchors at the range end (`> to`) per the default `offset: end`, instead of `>= from`.

## Testing

- New Rust integration tests `offset_no_granularity.rs` (SQL assertions + real-Postgres snapshots) covering trailing/leading × unbounded/finite × offset start/end; un-ignored the two finite no-granularity FIXME tests.
- New JS schema-compiler integration test verifying correct results under both planners and that `FILTER_PARAMS` does not leak window config into params.
- `cubesqlplanner`: 1052 passed; rolling_window with Postgres: 125 passed; JS tests green under both BaseQuery and Tesseract; `cargo fmt` clean.

### DIFF
--- a/rust/cube/cubesqlplanner/cubesqlplanner/src/tests/integration/rolling_window/offset_no_granularity.rs
+++ b/rust/cube/cubesqlplanner/cubesqlplanner/src/tests/integration/rolling_window/offset_no_granularity.rs
@@ -1,0 +1,228 @@
+use crate::test_fixtures::cube_bridge::MockSchema;
+use crate::test_fixtures::test_utils::TestContext;
+use indoc::indoc;
+
+fn create_context() -> TestContext {
+    let schema = MockSchema::from_yaml_file("common/integration_rolling_window.yaml");
+    TestContext::new(schema).unwrap()
+}
+
+const SEED: &str = "integration_rolling_window_tables.sql";
+
+// Rolling window measures queried without a granularity (dateRange only). The
+// whole range is a single window anchored by `offset`: 'start' anchors at the
+// range start (`from`), 'end' at the range end (`to`). The seed has orders before
+// (January) and after (March) the queried February range, which is what makes the
+// start/end anchors produce different aggregates.
+
+#[tokio::test(flavor = "multi_thread")]
+async fn test_trailing_unbounded_offset_start_no_granularity() {
+    let ctx = create_context();
+
+    let query = indoc! {r#"
+        measures:
+          - orders.rolling_sum_trailing_offset_start
+        time_dimensions:
+          - dimension: orders.created_at
+            dateRange:
+              - "2024-02-01"
+              - "2024-02-29"
+    "#};
+
+    let sql = ctx.build_sql(query).expect("Should generate SQL");
+
+    // offset start + trailing unbounded => everything strictly before the range start.
+    assert!(
+        sql.contains(r#""orders".created_at < $_0_$::timestamptz"#),
+        "Should have strict upper bound at range start (< from), got: {sql}"
+    );
+    assert!(
+        !sql.contains("created_at <= "),
+        "Should not use inclusive upper bound, got: {sql}"
+    );
+    assert!(
+        !sql.contains("created_at >"),
+        "Should not have a lower bound, got: {sql}"
+    );
+    assert!(
+        !sql.contains("time_series"),
+        "Without granularity should not reference time_series CTE, got: {sql}"
+    );
+
+    if let Some(result) = ctx.try_execute_pg(query, SEED).await {
+        insta::assert_snapshot!(result);
+    }
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn test_trailing_unbounded_offset_end_no_granularity() {
+    let ctx = create_context();
+
+    let query = indoc! {r#"
+        measures:
+          - orders.rolling_sum_trailing_offset_end
+        time_dimensions:
+          - dimension: orders.created_at
+            dateRange:
+              - "2024-02-01"
+              - "2024-02-29"
+    "#};
+
+    let sql = ctx.build_sql(query).expect("Should generate SQL");
+
+    // offset end + trailing unbounded => everything up to and including the range end.
+    assert!(
+        sql.contains(r#""orders".created_at <= $_0_$::timestamptz"#),
+        "Should have inclusive upper bound at range end (<= to), got: {sql}"
+    );
+    assert!(
+        !sql.contains("created_at >"),
+        "Should not have a lower bound, got: {sql}"
+    );
+
+    if let Some(result) = ctx.try_execute_pg(query, SEED).await {
+        insta::assert_snapshot!(result);
+    }
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn test_leading_unbounded_offset_start_no_granularity() {
+    let ctx = create_context();
+
+    let query = indoc! {r#"
+        measures:
+          - orders.rolling_sum_leading_offset_start
+        time_dimensions:
+          - dimension: orders.created_at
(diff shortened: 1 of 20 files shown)


rust/cube/cubesqlplanner/cubesqlplanner/src/tests/integration/rolling_window/offset_no_granularity.rs
```diff
@@ -0,0 +1,228 @@
...
+const SEED: &str = "integration_rolling_window_tables.sql";
+
+// Rolling window measures queried without a granularity (dateRange only). The
+// whole range is a single window anchored by `offset`: 'start' anchors at the
+// range start (`from`), 'end' at the range end (`to`). The seed has orders before
+// (January) and after (March) the queried February range, which is what makes the
+// start/end anchors produce different aggregates.
+
+#[tokio::test(flavor = "multi_thread")]
+async fn test_trailing_unbounded_offset_start_no_granularity() {
```
ovr, line 19: @claude could you share with me generated SQL from this test?

rust/cube/cubesqlplanner/cubesqlplanner/src/tests/integration/rolling_window/offset_no_granularity.rs
```diff
@@ -0,0 +1,228 @@
...
+        "Without granularity should not reference time_series CTE, got: {sql}"
+    );
+
+    if let Some(result) = ctx.try_execute_pg(query, SEED).await {
+        insta::assert_snapshot!(result);
+    }
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn test_trailing_unbounded_offset_end_no_granularity() {
```
ovr, line 58: @claude could you share with me generated SQL from this test?

rust/cube/cubesqlplanner/cubesqlplanner/src/tests/integration/rolling_window/offset_no_granularity.rs
```diff
@@ -0,0 +1,228 @@
...
+        "Should not have an upper bound, got: {sql}"
+    );
+
+    if let Some(result) = ctx.try_execute_pg(query, SEED).await {
+        insta::assert_snapshot!(result);
+    }
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn test_leading_unbounded_offset_end_no_granularity() {
```
ovr, line 120: @claude could you share with me generated SQL from this test?

ovr (approved):

PR #11180, opened 2026-06-30
docs: per-column table chart width controls

igorlukanin:
## What

Documents per-column **width controls** for the table chart: each column can be **Fixed** (a locked pixel width) or **Flexible** (a weight that shares the remaining table width), with bulk convert, drag-to-resize, and fit-to-content. Updates the "Column width" section and the column field-options table.

Companion to the Cube Cloud feature work (CUB-3033).


keydunov (approved):

keydunov (approved):
can we have screenshots / videos in docs for tables?

igorlukanin:
I'll add screenshots / video in a bit, when the UI stops changing that much. Also, we'll have plenty of videos to reuse: https://linear.app/cube-d3/issue/CUB-3103/add-screenshots-reuse-videos-in-docs

PR #11183, opened 2026-06-30
feat(cubesql): add disable_post_processing to /v1/cubesql endpoint

### LINKED ISSUE #11170: SQL API: option to reject queries that don't fully push down (instead of silently falling back to in-memory)
On the SQL API (/v1/cubesql), a query that can't be fully pushed down to the data source silently falls back to Cube/DataFusion in-memory execution. Cube fetches up to the row cap (CUBEJS_DB_QUERY_LIMIT, default 50k) as an intermediate result, post-processes it in memory, then applies the user's LIMIT. The returned rows are computed over a truncated, arbitrarily-ordered subset & silently incorrect, with no error signal. For analytics/agent use cases where correctness matters, this is a footgun: users (and LLM agents) can't tell a wrong answer from a right one.

There's currently no way to (a) make Cube reject such queries, or (b) reliably detect on the execution path whether a query fully pushed down. CUBESQL_SQL_PUSH_DOWN only toggles whether pushdown is attempted, and the JS config hooks (checkSqlAuth, queryRewrite) don't see the pushdown decision (it's made in the Rust cubesql layer).

Either or both:

1. Enforcement mode: a config flag (e.g. CUBESQL_REQUIRE_PUSHDOWN / per-request option) that makes the SQL API return an error instead of executing in memory when a query doesn't fully push down.

2. A first-class pushdown signal on the execution path e.g. a response field/header, or a documented, stable way to obtain "did this fully push down?" for a given SQL query (analogous to the sql4sql status in #10591, but on /v1/cubesql).

We built a transparent reverse proxy in front of /v1/cubesql that, before executing, runs EXPLAIN <query> and inspects the physical-plan root node: a fully pushed-down query roots at CubeScanExecutionPlan (the scan reports a single output partition, so no plumbing wrapper is added); any other root means an operator runs above the scan → block. This works but: it adds an EXPLAIN round-trip per query, relies on parsing plan text (version-fragile), and every consumer must front Cube with it rather than Cube enforcing centrally. A built-in option would be far more robust.

**Additional context**
- Affects correctness governance when multiple services share one Cube SQL API. 
- The fallback is invisible today; even a clear error would be a major improvement over silent truncation

We would be happy to work on the contribution if you are open to PRs

rynmccrmck:
Closes [here](https://github.com/cube-js/cube/issues/11170).

Queries that require DataFusion post-processing fetch intermediate rows from the source DB capped at CUBEJS_DB_QUERY_LIMIT (default 50k). On large datasets this silently truncates the intermediate result, producing wrong aggregates.

The sql4sql (/v1/sql?format=sql) path already classifies plans as pushdown/regular/post-processing and reports it to the caller, but never executes so it cant truncate. The /v1/cubesql path executes and streams results, so to guarantee accuracy it should block before execution.

This adds the same control to /v1/cubesql with a hard guarantee: if the final plan still requires post-processing after the cost model is biased against it, the request fails (the error is delivered in the streamed response body) rather than returning truncated data.

Changes:
- env.ts: register CUBESQL_DISABLE_POST_PROCESSING env var (server-wide default, off by default) so all services are covered without per-request changes
- gateway.ts: read disable_post_processing from request body, fall back to env var, pass through to execSql
- sql-server.ts, index.ts: thread disablePostProcessing param to native
- node_export.rs: set CUBESQL_PENALIZE_POST_PROCESSING_VAR before planning (biases the cost model toward pushdown) and, after planning, fail if the plan root is not a single CubeScan(Wrapped) node

Tests:
- smoke-cubesql.test.ts: error path (asserts the post-processing error is surfaced in the streamed body), push-down success, default-off baseline
- api-gateway/test/index.test.ts: gateway unit test asserting the param is forwarded to execSql; existing /v1/cubesql call-signature assertions updated for the new trailing argument
- backend-shared/test/env.test.ts: unit test for CUBESQL_DISABLE_POST_PROCESSING

**Check List**


rynmccrmck:
@keydunov is this something you would consider merging?

PR #11265, opened 2026-07-15
fix(snowflake-driver): surface an actionable error when the connection fails

igorlukanin:
## Summary

Improves how the Snowflake driver surfaces connection failures. Previously it let the raw `snowflake-sdk` error bubble up unattributed (e.g. `RequestFailedError: Request to Snowflake failed.`), which isn't clearly tied to Snowflake once it reaches the API/logs and, on the startup connection check, can crashloop the pod.

Adds a `ConnectionError` (same pattern as the Postgres/Cubestore drivers) wrapping SDK connection failures into `Unable to connect to Snowflake: …`. It preserves the original message, keeps the original on `cause`, and defensively maps the opaque `TypeError: Cannot read properties of undefined (reading 'replace')` the SDK can throw from inside its own error formatting to a credentials hint (kept alongside the original message, not replacing it). Wired into `createConnection()` + a new `connect()` helper used by `testConnection()` and `initConnection()`. Non-breaking: error surface only.

Note on scope: the opaque `reading 'replace'` TypeError originally reported lives in the SDK's interactive browser/OAuth-authorization-code flow, which this server-side driver does not use — on the driver's actual paths, invalid creds already surface as a structured `OperationFailedError`/`RequestFailedError`. This change is defensive attribution of any connection failure, not a repro-driven fix of that specific SDK path.

## Test plan

- [x] 17 unit tests (`test/errors.test.ts`), no live Snowflake
- [x] `tsc` / `eslint` / `yarn build` green
- [x] Verified before→after against the real driver with a bogus account (`RequestFailedError` → `Unable to connect to Snowflake: …`)
- [ ] CI

### DIFF
--- a/packages/cubejs-snowflake-driver/src/SnowflakeDriver.ts
+++ b/packages/cubejs-snowflake-driver/src/SnowflakeDriver.ts
@@ -25,6 +25,7 @@ import crypto from 'crypto';
 import { S3ClientConfig } from '@aws-sdk/client-s3';
 import { HydrationMap, HydrationStream } from './HydrationStream';
 import { hydrators } from './type-parsers';
+import { ConnectionError } from './errors';
 
 const SUPPORTED_BUCKET_TYPES = ['s3', 'gcs', 'azure'];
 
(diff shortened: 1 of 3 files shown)


igorlukanin:
Closing — investigation showed the reported TypeError isn't from the Snowflake driver. Root cause is Cloud-side multi-cluster Cube Store routing (CUBE_CLOUD_MULTICLUSTER_CUBE_STORE_PER_ORCHESTRATOR left set after a downgrade dropped the tenant's Cube Store cluster, so the URL-template env was absent and the runtime did undefined.replace(...)). Tracking the real fix internally.

PR #11285, opened 2026-07-17
docs: document sending email notifications to user groups

mngr:
## Description

Documents that dashboard scheduled-refresh **email** notifications can now be sent to **user groups** alongside individual users.

Changes to `docs-mintlify/docs/explore-analyze/notifications.mdx`:

- **Delivery channel** — the Email bullet now notes that the recipient picker is grouped into **Users** and **User groups** sections, so you can add individual workspace users, entire user groups, or a mix of both.
- **Email notifications** — new **Recipients** subsection describing the two picker sections and the behaviour that matters at send time:
  - A group is expanded to its **current** members each time the notification is sent, so membership changes take effect on the next run without editing the schedule.
  - The **User groups** section only appears when the workspace has at least one user group.
  - A recipient selected individually who is also a member of a selected group is emailed only once.
- Added a `[ref-user-groups]` cross-link to the [user groups](/admin/users-and-permissions/user-groups) page.

## Notes

Docs-only change. I did not run the full Mintlify build locally; the link targets were verified against the existing pages (`user-groups.mdx`, matching the `[ref-roles]` link style already used on this page).

### DIFF
--- a/docs-mintlify/docs/explore-analyze/notifications.mdx
+++ b/docs-mintlify/docs/explore-analyze/notifications.mdx
@@ -33,8 +33,10 @@ scratch — the copy carries over the source schedule's notification settings.
 
 Toggle between two delivery channels:
 
-- **Email** — select multiple recipients from workspace users using a searchable
-  picker with checkboxes.
+- **Email** — select recipients using a searchable picker with checkboxes. The
+  picker is grouped into **Users** and **User groups** sections, so you can add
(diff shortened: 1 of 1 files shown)


keydunov (approved):

mngr:
Superseded by #11301 — recreated from a branch on the upstream repo so the required `review-with-tracking` check can run (it can't fetch an OIDC token on fork PRs). Same content, rebased onto current master.

PR #11288, opened 2026-07-17
fix(schema-compiler): Throw UserError for unknown granularity so API returns 400 instead of 500

BrianSigafoos-SQ:
**Check List**

Querying a time dimension with an unknown granularity (for example `granularity: "all"`, which some older clients send) throws a plain `Error` from `Granularity`'s constructor. The API gateway only classifies `UserError` as a client error, so the REST API responds with `500 Internal Server Error` for what is really invalid client input:

```
Error: Granularity "all" does not exist in dimension Sales.local_reporting_timestamp
```

This change throws `UserError` instead, so `/v1/load` returns `400` for unknown granularities, consistent with other invalid-query errors (unknown members, malformed filters, etc.).

Added a unit test in `base-query.test.ts` covering the unknown-granularity case; ran the full `Custom granularities` suite (35 tests) and the schema-compiler linter (0 errors).

### DIFF
--- a/packages/cubejs-schema-compiler/src/adapter/Granularity.ts
+++ b/packages/cubejs-schema-compiler/src/adapter/Granularity.ts
@@ -10,6 +10,7 @@ import {
   timeSeriesFromCustomInterval,
   TimeSeriesOptions
 } from '@cubejs-backend/shared';
+import { UserError } from '../compiler/UserError';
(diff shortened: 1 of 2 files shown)


ovr:
Thank you @BrianSigafoos-SQ for your contribution!

PR #11315, opened 2026-07-21
feat(cubesql): Support joining ungrouped queries with SQL push down

MazterQyou:
**Check List**

This PR supports SQL push down for joins where either side is an ungrouped (raw) query, e.g. a self-join of a CTE selecting dimensions, by planning them as plain SQL joins of standalone subqueries instead of failing with `Use __cubeJoinField to join Cubes`. Related tests are included.

### DIFF
--- a/rust/cubesql/cubesql/src/compile/engine/df/optimizers/snapshots/cubesql__compile__engine__df__optimizers__sort_push_down__tests__sort_stays_above_filter_over_join.snap
+++ b/rust/cubesql/cubesql/src/compile/engine/df/optimizers/snapshots/cubesql__compile__engine__df__optimizers__sort_push_down__tests__sort_stays_above_filter_over_join.snap
@@ -1,0 +1,11 @@
+---
+source: cubesql/src/compile/engine/df/optimizers/sort_push_down.rs
+expression: optimize(&plan)
+---
+Sort: #j1.c1 ASC NULLS LAST
+  Filter: #j1.c1 >= #j2.c2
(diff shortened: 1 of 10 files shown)


MazterQyou:
> `egraph.add` for `out_grouped_subqueries` runs before any check can reject

Acknowledged, keeping as is. The node is hash-consed: `WrapperReplacerContextGroupedSubqueries(vec![])` is interned once for the lifetime of the e-graph, so every attempted match after the first is a hashmap lookup, not an allocation. Moving it back behind the checks would either re-inline it into the key loops (the exact readability problem the hoist fixed) or require lazy-init machinery that costs more in code than the lookup does at runtime.

MazterQyou:
> `sort_can_reach_scan`'s `_ => true` catch-all treats every unlisted node as scan-reachable

Deliberate, and empirically load-bearing in that direction. A stricter draft of this guard (default `false`, explicit allow-list) regressed `powerbi_year_month_split` and `test_metabase_dow`/`test_metabase_doy` to raw-scan post-processing plans: grouped CubeScans legitimately absorb an ORDER BY *through* an Aggregate into the load query's `order` field, and the same optimism has to extend to nodes we haven't enumerated. As you note, the failure mode of over-permitting is a sort that gets pushed and simply isn't absorbed - order-preserving either way - while over-restricting demonstrably loses pushdown. `Join`/`CrossJoin` stay the only barriers because they're the only nodes where a stuck sort changes which level the ORDER BY lands at in emitted SQL.

PR #11325, opened 2026-07-22
fix(native): Jinja - forward kwargs on sync Python function/method calls

igorlukanin:
## Summary

- Keyword arguments passed to Python functions/methods from Jinja templates on the **synchronous** call path were flattened into the positional tuple instead of being forwarded as kwargs. The callee received the kwargs dict as a positional value, so e.g. `model.as_dimensions(skip=['id'])` reached Python as `as_dimensions({'skip': ['id']})` — the `skip` list never applied, producing surprising results (e.g. a column that should have been skipped still rendered).
- The async function-call path (`process_task` in `runtime.rs`) already split kwargs correctly (#9276); the three sync helpers in `python/utils.rs` (`python_fn_call_sync`, `python_obj_call_sync`, `python_obj_method_call_sync`) did not. This brings them to parity: a shared `split_positional_and_kwargs` partitions args by `is_kwarg()` into a positional tuple + optional `PyDict`, and the calls switch from `call1`/`call_method1` (positional-only) to `call_bound`/`call_method_bound` (kwargs-aware).
- No behavior change for positional-only calls (empty kwargs → `None`, identical to before).

## Test plan

- [x] Added method-call-with-kwargs coverage to `arguments-test.yml.jinja` (a new `ExampleClassModelA.method_kwargs` / `method_named_arguments` invoked with keyword args, including out-of-order named args) — snapshot confirms kwargs now resolve by name.
- [x] `yarn native:build-debug-python`, `yarn tsc`, `cargo fmt --check`, `cargo clippy --features python`, `yarn lint` all clean.
- [x] Jinja test suite green (17/17).
- [ ] CI must pass.

### DIFF
--- a/packages/cubejs-backend-native/src/python/utils.rs
+++ b/packages/cubejs-backend-native/src/python/utils.rs
@@ -5,10 +5,7 @@ use pyo3::types::{PyDict, PyFunction, PyString, PyTuple};
 use pyo3::{ffi, AsPyPointer};
 use std::ffi::c_int;
 
-/// Splits minijinja call arguments into a positional tuple and an optional
-/// kwargs dict, so `fn(key=value)` reaches Python as a keyword argument rather
-/// than a positional kwargs dict. Mirrors the async path in runtime.rs.
-fn split_positional_and_kwargs(
+pub(crate) fn split_positional_and_kwargs(
     py: Python<'_>,
     arguments: Vec<CLRepr>,
 ) -> PyResult<(Vec<PyObject>, Option<Bound<'_, PyDict>>)> {
--- a/packages/cubejs-backend-native/src/python/utils.rs
+++ b/packages/cubejs-backend-native/src/python/utils.rs
@@ -5,11 +5,9 @@ use pyo3::types::{PyDict, PyFunction, PyString, PyTuple};
 use pyo3::{ffi, AsPyPointer};
 use std::ffi::c_int;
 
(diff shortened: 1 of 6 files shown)


packages/cubejs-backend-native/src/python/utils.rs
```diff
@@ -1,21 +1,38 @@
...
 use pyo3::{ffi, AsPyPointer};
 use std::ffi::c_int;
 
-pub fn python_fn_call_sync(py_fun: &Py<PyFunction>, arguments: Vec<CLRepr>) -> PyResult<CLRepr> {
-    Python::with_gil(|py| {
-        let mut args_tuple = Vec::with_capacity(arguments.len());
+/// Splits minijinja call arguments into a positional tuple and an optional
+/// kwargs dict, so `fn(key=value)` reaches Python as a keyword argument rather
+/// than a positional kwargs dict. Mirrors the async path in runtime.rs.
+fn split_positional_and_kwargs(
```
ovr: **Mirrors the async path in runtime.rs.**

Let's re-use this function in `async` path in `runtime.rs` too and drop this whole comment, because it's slop.

igorlukanin: Done — pulled it out into a shared fn used by the async path too, comment gone.

ovr (approved):
🚀

igorlukanin:
Thank you @ovr!

PR #11328, opened 2026-07-22
fix(tesseract): Emit GROUP BY for non-windowed Aggregate stages in ungrouped queries

### LINKED ISSUE #11327: SQL API: ungrouped query selecting a group_by-locked multi-stage measure generates aggregates without GROUP BY (rejected by Postgres)
## Describe the bug

When the SQL API receives an **ungrouped** query (a bare `SELECT` without `GROUP BY` — exactly what BI tools like Metabase send when a user browses a table) that selects a **multi-stage measure locked with `group_by` / `reduce_by`** together with at least one dimension, the generated data-source SQL contains plain aggregate functions (`sum(...)`) next to bare dimension columns **but no `GROUP BY` clause anywhere**. Any database that enforces SQL grouping rules rejects it:

```
ERROR: Database Execution Error: column "fk_aggregate_keys.orders__city"
must appear in the GROUP BY clause or be used in an aggregate function
```

Grouped queries (`GROUP BY` + `MEASURE(...)`) over the same members work correctly. Ungrouped queries selecting only plain measures, or only dimensions, or the multi-stage measure without any dimension, also work.

Real-world impact: Metabase's table-browse view emits exactly this bare-select shape, so any table synced through the SQL API whose view exposes such a measure is un-browsable.

## To Reproduce

Data model (any Postgres source table `public.orders(id, customer_id, status, city)`):

```yaml
cubes:
  - name: orders
    sql_table: public.orders
    dimensions:
      - name: status
        sql: status
        type: string
      - name: city
        sql: city
        type: string
    measures:
      - name: unique_customers
        type: count_distinct
        sql: customer_id
      # multi-stage measure: total locked to `status` grain
      - name: unique_customers_by_status
        multi_stage: true
        type: sum
        sql: "{unique_customers}"
        group_by:
          - status
```

Connect any Postgres client to the SQL API and run:

```sql
SELECT unique_customers_by_status, city FROM orders LIMIT 1;
```

The query fails with the `must appear in the GROUP BY clause` error from the upstream database. Selecting the measure alone (no dimension) works; adding any dimension fails. `SELECT status, MEASURE(unique_customers_by_status) FROM orders GROUP BY 1` works.

Note the inner measure being non-sum-rollable (`count_distinct`, `avg`, `max`, …) matters: with a plain `sum`-over-`sum`/`count` the planner takes the window-function path (`sum(...) OVER (...)`), which is valid without `GROUP BY`, so the bug is masked.

## Generated SQL (simplified)

The multi-stage Aggregate stage CTE is rendered as:

```sql
cte_2 AS (
  SELECT
    "fk_aggregate_keys"."orders__status" "orders__status",
    "fk_aggregate_keys"."orders__city"   "orders__city",
    sum("q_0"."orders__unique_customers") "orders__unique_customers_by_status"
  FROM (SELECT DISTINCT "orders__status", "orders__city" FROM cte_1) AS "fk_aggregate_keys"
  LEFT JOIN (SELECT * FROM cte_0) AS "q_0"
    ON ("fk_aggregate_keys"."orders__status" = "q_0"."orders__status")
  -- missing: GROUP BY "fk_aggregate_keys"."orders__status", "fk_aggregate_keys"."orders__city"
)
```

`sum()` next to bare dimension columns, and no `GROUP BY` in the whole statement.

## Root cause (traced in v1.7.7 sources)

The statement is produced by the Tesseract planner (`rust/cube/cubesqlplanner`):

1. The SQL API pushdown marks the whole query `ungrouped: true`.
2. [`planner/planners/multi_stage/multi_stage_query_planner.rs#L168-L171`](https://github.com/cube-js/cube/blob/v1.7.7/rust/cube/cubesqlplanner/cubesqlplanner/src/planner/planners/multi_stage/multi_stage_query_planner.rs#L168-L171) — an **Aggregate** inode inherits `is_ungrupped` from the top-level query's `ungrouped` flag. (Rank/Calculate inodes are always `true`, which is fine — they render window functions / plain expressions, not bare aggregates.)
3. [`planner/planners/multi_stage/member_query_planner.rs#L298`](https://github.com/cube-js/cube/blob/v1.7.7/rust/cube/cubesqlplanner/cubesqlplanner/src/planner/planners/multi_stage/member_query_planner.rs#L298) — the flag is stored on `MultiStageMeasureCalculation`.
4. [`physical_plan_builder/processors/multi_stage_measure_calculation.rs#L61`](https://github.com/cube-js/cube/blob/v1.7.7/rust/cube/cubesqlplanner/cubesqlplanner/src/physical_plan_builder/processors/multi_stage_measure_calculation.rs#L61) — `GROUP BY` is emitted only `if !measure_calculation.is_ungrouped()`, so it is skipped.
5. But the measure projection in the same select renders the outer aggregate function (`sum(...)`) unconditionally: the ungrouped-measure rendering path (`physical_plan/sql_nodes/factory.rs#L220`) applies only to leaf CTEs via `measure_for_ungrouped`, never to the Aggregate calculation stage.

So for `calculation_type == Aggregate` with `window_function_to_use == None`, the ungrouped flag suppresses `GROUP BY` while the projection still contains plain aggregates → invalid SQL. Reproduced on v1.7.1 through v1.7.7; the relevant code is unchanged on current master.

## Expected behavior

A non-windowed Aggregate stage must always be grouped by its projected dimensions — it renders plain aggregate functions by construction. The query-level ungrouped semantics are preserved because the enclosing `FullKeyAggregate` join broadcasts the stage output back to detail rows.

## Version

- Cube: v1.7.1 … v1.7.7 (Docker `cubejs/cube`), code unchanged on master
- Data source: PostgreSQL (any version enforcing standard grouping rules)
- Client: any Postgres client via SQL API (reproduced with psql and Metabase v0.63.x)

esrever10:
**Check List**

Fixes #11327

An ungrouped SQL API query (bare `SELECT` without `GROUP BY`, the shape BI tools emit when browsing a table) selecting a `group_by`/`reduce_by`-locked multi-stage measure plus any dimension produced invalid SQL: the measure's Aggregate stage projected plain `sum(...)` next to bare dimension columns with no `GROUP BY`, so databases enforcing SQL grouping rules (e.g. Postgres) rejected the statement. Full trace in #11327.

The fix follows from what the stage renders: a **non-windowed Aggregate stage projects plain aggregate functions by construction**, so its select must be grouped by its projected dimensions regardless of the query-level `ungrouped` flag. Ungrouped semantics are unaffected — the enclosing `FullKeyAggregate` join broadcasts the stage output back to detail rows. Rank / Calculate / windowed stages are untouched: they render window functions or plain expressions and are valid without `GROUP BY`. `ORDER BY` behavior is unchanged.

Changes:
- `physical_plan_builder/processors/multi_stage_measure_calculation.rs`: emit `GROUP BY` when the stage is grouped **or** renders plain aggregates (`calculation_type == Aggregate && window_function_to_use == None`).
- `test_fixtures/.../integration_multi_stage.yaml`: add `unique_customers_group_by_status` — a `group_by`-locked `sum` over `count_distinct`, i.e. a non-sum-rollable input that is not window-path eligible and therefore hits the plain-aggregate JOIN model (existing ungrouped tests only covered Calculate-type / window-path measures, which are valid without `GROUP BY`).
- `tests/integration/multi_stage/ungrouped.rs`: regression test — fails on master (`assert sql.contains("GROUP BY")`; the whole statement previously had none), passes with the fix; also executes against Postgres under the `integration-postgres` feature.

Verification: `cargo test -p cubesqlplanner` — 1076 passed, 0 failed (new test red on master, green with the fix).

### DIFF
--- a/rust/cube/cubesqlplanner/cubesqlplanner/src/physical_plan_builder/processors/multi_stage_measure_calculation.rs
+++ b/rust/cube/cubesqlplanner/cubesqlplanner/src/physical_plan_builder/processors/multi_stage_measure_calculation.rs
@@ -1,6 +1,8 @@
 use super::super::context::PushDownBuilderContext;
 use super::super::{LogicalNodeProcessor, ProcessableNode};
-use crate::logical_plan::{MultiStageCalculationWindowFunction, MultiStageMeasureCalculation};
+use crate::logical_plan::{
+    MultiStageCalculationType, MultiStageCalculationWindowFunction, MultiStageMeasureCalculation,
+};
 use crate::physical_plan::ReferencesBuilder;
 use crate::physical_plan::{Expr, MemberExpression, QueryPlan, SelectBuilder};
 use crate::physical_plan_builder::PhysicalPlanBuilder;
@@ -58,7 +60,18 @@ impl<'a> LogicalNodeProcessor<'a, MultiStageMeasureCalculation>
             select_builder.add_projection_member(measure, alias);
         }
 
-        if !measure_calculation.is_ungrouped() {
+        // A non-windowed Aggregate stage projects plain aggregate functions
+        // (e.g. `sum(...)`) next to its dimensions, so the select must be
+        // grouped by those dimensions regardless of the query-level ungrouped
+        // flag — the enclosing FullKeyAggregate join broadcasts the stage
+        // output back to detail rows, keeping ungrouped semantics intact.
+        // Rank / Calculate / windowed stages render window functions or plain
+        // expressions and stay valid without GROUP BY.
+        let renders_plain_aggregates = measure_calculation.calculation_type()
+            == &MultiStageCalculationType::Aggregate
+            && measure_calculation.window_function_to_use()
+                == &MultiStageCalculationWindowFunction::None;
+        if !measure_calculation.is_ungrouped() || renders_plain_aggregates {
             let group_by = all_dimensions
                 .iter()
                 .map(|dim| -> Result<_, CubeError> {
@@ -66,6 +79,8 @@ impl<'a> LogicalNodeProcessor<'a, MultiStageMeasureCalculation>
                 })
                 .collect::<Result<Vec<_>, _>>()?;
             select_builder.set_group_by(group_by);
+        }
+        if !measure_calculation.is_ungrouped() {
             select_builder.set_order_by(
                 self.builder
                     .make_order_by(measure_calculation.schema(), measure_calculation.order_by())?,
--- a/rust/cube/cubesqlplanner/cubesqlplanner/src/test_fixtures/schemas/yaml_files/common/integration_multi_stage.yaml
+++ b/rust/cube/cubesqlplanner/cubesqlplanner/src/test_fixtures/schemas/yaml_files/common/integration_multi_stage.yaml
@@ -249,6 +249,16 @@ cubes:
                 - orders.status
                 - orders.created_at.month
 
+          # group_by-locked sum over a non-sum-rollable input
+          # (count_distinct): not window-path eligible, so the Aggregate
+          # stage renders a plain sum() over the FullKeyAggregate join.
+          - name: unique_customers_group_by_status
+            type: sum
+            sql: "{CUBE.unique_customers}"
+            multi_stage: true
+            group_by:
+                - orders.status
+
           - name: amount_grain_keep_only_status
             type: sum
             sql: "{CUBE.total_amount}"
--- a/rust/cube/cubesqlplanner/cubesqlplanner/src/tests/integration/multi_stage/snapshots/cubesqlplanner__tests__integration__multi_stage__ungrouped__ungrouped_group_by_locked_aggregate_with_dimension.snap
+++ b/rust/cube/cubesqlplanner/cubesqlplanner/src/tests/integration/multi_stage/snapshots/cubesqlplanner__tests__integration__multi_stage__ungrouped__ungrouped_group_by_locked_aggregate_with_dimension.snap
@@ -1,0 +1,9 @@
+---
+source: cubesqlplanner/cubesqlplanner/src/tests/integration/multi_stage/ungrouped.rs
+expression: result
+---
+orders__category | orders__unique_customers_group_by_status
+-----------------+-----------------------------------------
+books            | 30                                      
+clothing         | 30                                      
+electronics      | 30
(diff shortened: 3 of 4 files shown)


esrever10:
Pushed a rebase onto latest `master` plus the missing insta snapshot for
`test_ungrouped_group_by_locked_aggregate_with_dimension` — that was the only
CI failure attributable to this PR (`unit-core`: the test file was committed
without its `.snap`, so insta reported the result table as a new snapshot).

The snapshot was generated by running the test against the testcontainers
Postgres, not hand-written. Verification on the rebased branch:

- `cargo test -p cubesqlplanner --lib --features cubesqlplanner/integration-postgres`
  — 1078 passed, 0 failed, 8 ignored.
- `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` — clean.
- Short-circuiting the new gating condition makes the test fail on the
  `GROUP BY` assertion, printing exactly the shape from the report —
  `sum("q_0"."orders__unique_customers")` beside a bare
  `"fk_aggregate_keys"."orders__category"`, with no `GROUP BY` in the whole
  statement — so it is a genuine regression test rather than a snapshot-only one.

The other three red checks on the previous run were unrelated to this diff:

- **`Unit (Rewrite Engine)`** — `error: aborting because of unreferenced snapshots`,
  listing only pre-existing `cubesql/e2e/tests/snapshots/e2e__tests__postgres__*.snap`
  files. Nothing in this PR touches `rust/cubesql`.
- **`integration-smoke (24.x, 3.13, false)`** — `(HTTP code 500) Get
  "https://mcr.microsoft.com/v2/": context deadline exceeded` while pulling the
  mssql image. Infrastructure flake.
- **`review-with-tracking`** — `Unable to get ACTIONS_ID_TOKEN_REQUEST_URL` /
  missing `id-token: write`; OIDC is not available to fork PRs.

@paveltiunov thanks for the confirmation and for building the live regression
test — the gating condition in this PR
(`calculation_type() == Aggregate && window_function_to_use() == None`) is the
same one you validated. Your branch `claude/serene-mendel-m0eoyz` adds
`test_ungrouped_group_by_locked_measure_with_dimension` over a
`unique_customers_group_by_status` fixture, which is essentially the same
scenario as the test here. Happy to drop mine in favour of yours, or to keep
both — whichever you prefer for the merge.

Also worth noting for whoever reviews this area: #11366 touches the adjacent
multi-stage `add_group_by` path (`member_query_planner.rs`, rolling-window
rollup CTE grain). Different root cause and no overlapping hunks with this PR,
so the two rebase cleanly against each other.

PR #11329, opened 2026-07-22
fix(cubesql): Escape PatchMeasure filters for member expression evaluation

MazterQyou:
**Check List**

This PR escapes backslashes and backticks in PatchMeasure filter SQL the same way as other member expressions, fixing database syntax errors for `SUM(CASE WHEN ... LIKE/ILIKE ...)` queries on dialects with backslash escape clauses. Related test is included.

### DIFF
--- a/rust/cubesql/cubesql/src/compile/engine/df/wrapper.rs
+++ b/rust/cubesql/cubesql/src/compile/engine/df/wrapper.rs
@@ -1401,6 +1401,7 @@ impl WrappedSelectNode {
                             Some(push_to_cube_context),
                             subqueries,
                         )?;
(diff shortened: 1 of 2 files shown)


ovr (approved):

rust/cubesql/cubesql/src/compile/engine/df/wrapper.rs
```diff
@@ -1401,6 +1401,7 @@ impl WrappedSelectNode {
                             Some(push_to_cube_context),
                             subqueries,
                         )?;
+                        let filter = Self::escape_interpolation_quotes(filter, true);
```
ovr, line 1404: @claude could you verify that we didn't miss other places where escape interplotation is required

PR #11346, opened 2026-07-24
feat(tesseract): pre-agg matching without stored calc-group dims; collapse trivial pass-through subqueries; e2e specs for cross-cube rolling measures over rollups

paveltiunov:
**Check List**

**Feature: pre-aggregation matching no longer requires calc-group (switch) dimensions to be stored in the rollup (Tesseract).** Calc-group dimensions are virtual enumerations — they carry no stored data, and materializing one in a rollup just CROSS JOINs the value list into the build SQL, multiplying rollup size by the number of values while adding zero information. They previously had to be listed in every rollup definition purely because the matcher treated them like ordinary dimensions. Now:

- `dimension_matcher.rs`: a calc-group dimension always matches (`MatchState::Full`) regardless of the rollup's stored dimensions.
- `query.rs` (physical plan builder): calc-group resolution — a filter-pinned value renders as a literal, otherwise the enumeration is cross-joined as a virtual values table — now also applies when the query source is a pre-aggregation, covering exactly the calc groups **not** stored in the rollup. Stored calc groups keep resolving to the rollup column, so existing rollup definitions behave exactly as before (backward compatible). Literal render references are re-added after the pre-aggregation render-reference reset so they survive it.

**Fix: physical-plan optimizer collapsing trivial pass-through subqueries (Tesseract).** Multi-stage planning wraps every CTE reference into a derived table of the form `(SELECT * FROM cte_n AS cte_n) AS alias`. Each wrapper is a semantic no-op, but engines that inline CTE bodies at every reference (Cube Store / DataFusion) pay two extra plan nodes (projection + subquery alias) per usage — deep multi-stage queries over pre-aggregations could overflow Cube Store's serialized-plan protobuf decode recursion limit and fail with `Error during planning: Error decoding expr as protobuf: ... recursion limit reached`. The new pass in `physical_plan/optimizers/trivial_subquery.rs` rewrites such wrappers into direct `<table> AS alias` references across the whole plan (top-level select, CTE bodies, join sides, unions, nested subqueries), collapsing only pure pass-throughs (`SELECT *`, no filter/group/having/order/distinct/limit/offset/CTEs, single table-reference source). On the repro query below this removes all 23 wrappers, and the previously failing query now executes in Cube Store.

Verification: 1076 cubesqlplanner tests (including `--features integration-postgres`, which executes generated SQL and compares results) and the `integration-cubestore` test pass; a test asserts the collapsed SQL shape and unchanged execution results.

**Test coverage added** for rolling-window metrics exposed through multi-stage `case` entrypoint measures dispatched by a `type: switch` dimension (calc group) across two joined fact cubes combined in a view:

1. **Schema-compiler integration spec** (`pre-aggregations-shared-calc-group.test.ts`, Postgres rollups): a shared single-row calc-group cube joined `1 = 1` into both fact cubes lets one filter resolve every case measure — cross-cube query matches both rollups; per-cube switch dimensions also match now that calc groups no longer block matching (asserted); a slim-rollup model verifies rollups that do **not** store the calc-group dimension still match, render the pinned value as a literal, and execute; deterministic value assertions pin the rolling/time-shift arithmetic.
2. **Birdbox smoke test** (`smoke-shared-calc-group.test.ts` + fixture): the same model on a real Cube server with rollups in **Cube Store** (`CUBEJS_ROLLUP_ONLY`, Tesseract), with the switch dimension **omitted from both rollup definitions**. Single-cube, cross-cube, three-measure, and the previously failing **four-measure deep query** all execute from Cube Store rollups.

Remaining gaps documented as skipped `FIXME` tests:

1. **Granular time dimension:** fails in the physical plan builder with `Alias not found for partition_by dimension …rolling_window` — the calc-group dimension is not projected into the multi-stage window input CTE.
2. **Date-range-only time dimension** (no granularity) stops matching the rollups even though `allow_non_strict_date_range_match: true` is set and the range is month-aligned.

https://claude.ai/code/session_01QxbxLWpBwunw3FypC8zfnY


paveltiunov:
CI status note: several red checks on `9927f50` are runner-infrastructure failures, not related to the diff:

- `tests (pinot, true)` and `tests (databricks-jdbc-export-bucket-s3, true)` — Docker Hub registry timeouts (`Get "https://registry-1.docker.io/v2/": Client.Timeout exceeded`), all tests fail fast at container pull.
- `Build macos-14 aarch64-apple-darwin (Python fallback)` — rustup toolchain download failed with a DNS error resolving `static.rust-lang.org`.

The failures this PR actually caused (`Check fmt/clippy`, `lint`, `unit` — rustfmt on the new optimizer file and three OracleQuery alias-shape assertions) are fixed in `9927f50`. The infra-failed jobs need a re-run once the registry recovers — this integration doesn't have permission to re-run them.

---

paveltiunov:
CI note: the red `tests (redshift, true)` check is an infrastructure failure on the manual re-run, not a test failure. The job died in setup — `actions/download-artifact` resolved the `docker-image` artifact (ID 8613461437) but `GetSignedArtifactURL` returned `404 Not Found: workflow run not found`, a known transient GitHub Actions backend error when re-running a single job. Zero tests executed.

This matrix leg is non-required and was originally cancelled by concurrency dedup; every other check on `ddf3e24` is green (103 passing). I don't have API permission to re-run failed jobs from this session (403), so if a green mark is wanted here, one more "Re-run failed jobs" from the Actions UI should do it — the artifact itself is present and only hours old.

---

PR #11349, opened 2026-07-24
docs: Private API Connectivity pages for Azure and GCP + Chrome local network access caveat

bsod90:
## Why

Chromium-based browsers block requests from a page served over the public internet to a host that resolves to a private address, unless the site holds the **Local network access** permission. That is exactly the shape of a private Cube setup: the Cube UI is served from the public `*.cubecloud.dev` origin, while the deployment's data API hostname resolves to an RFC 1918 address over the customer's VPN. The UI's own fetches are therefore *local network requests*, and Chrome drops them.

It surfaces as a red **"Network Error"** banner on charts, the Semantic Model IDE, and the playground — while `curl` from the same laptop succeeds, because only browsers enforce this. We have explained it one-off to several private-connectivity and BYOC customers and it has never been written down, so every occurrence starts as a suspected Cube bug. The tell is that it works for one colleague and fails for another on the same VPN, which is almost always an unmanaged Chrome profile that enterprise policy does not reach.

Chrome enforces this as [Local Network Access](https://developer.chrome.com/blog/local-network-access) from Chrome 142 onwards, and as its predecessor [Private Network Access](https://developer.chrome.com/blog/private-network-access-update) before that.

## What changed

Private API Connectivity existed for AWS only — the Azure and GCP pages pointed readers at the AWS page and said the equivalent was "available on request". This adds a page per provider so each platform has a real home for frontend connectivity, and gives all three the Chrome caveat.

**New pages**

- **Private API Connectivity on GCP** — Private Service Connect service attachments fronting the ingress and SQL API internal load balancers, consumer PSC endpoints (`gcloud` + console), Cloud DNS private zones, both TLS options, and the accept-list/`pscConnectionStatus` failure modes.
- **Private API Connectivity on Azure** — the equivalent over Private Link Services and consumer private endpoints (`az` + portal), with Azure Private DNS zones and Application Gateway hostname-override guidance.

**On all three pages** — a `Chrome local network access permission` section: the cause, why it applies to both TLS options (the browser decides from the resolved IP, not the hostname), the `ERR_BLOCKED_BY_PRIVATE_NETWORK_ACCESS_CHECKS` signature, the per-user prompt / site setting, and the fleet fix as a version-split table:

| Chrome version | Policy |
| --- | --- |
| 142 and later | [`LocalNetworkAccessAllowedForUrls`](https://chromeenterprise.google/policies/local-network-access-allowed-for-urls/) |
| Before 142 | [`InsecurePrivateNetworkRequestsAllowedForUrls`](https://chromeenterprise.google/policies/insecure-private-network-requests-allowed-for-urls/) |

Plus the `chrome://policy` verification step and the unmanaged-profile caveat. On the AWS page it is also wired into the existing "Network Error" troubleshooting steps.

**Wiring** — Azure and GCP cross-references now land on their own provider's page instead of the AWS one; the new pages are listed in `docs.json`; the Azure and GCP BYOC pages get the same pointer the AWS BYOC page already carried.

## Notes for review

- **Azure is documented as an on-request pattern.** AWS and GCP both have producer-side infrastructure in place today (VPC Endpoint Services on AWS; PSC service attachments in front of the Traefik and cubesql-router internal LBs on GCP). Azure has the consumer side and the cube-store Private Link Service producer, but no customer-facing API-ingress Private Link Service yet. The page's availability callout says so explicitly — "provisioned per tenant on request … before you design around this setup" — which matches the claim the Azure pages already made. Flagging it so we can decide whether that framing is strong enough.
- Diagrams for the Azure and GCP pages are left as `{/* TODO: diagram */}` placeholders rather than reusing the AWS topology image.
- `yarn broken-links --check-anchors --check-redirects --check-snippets` reports only the 16 pre-existing failures in `embedding/iframe/events.mdx` and `embedding/iframe/localization.mdx`; nothing here is new. All external Chrome, GCP, and Azure doc links verified 200.

### DIFF
--- a/docs-mintlify/admin/deployment/dedicated/aws/private-api-connectivity.mdx
+++ b/docs-mintlify/admin/deployment/dedicated/aws/private-api-connectivity.mdx
@@ -382,6 +382,54 @@ below for steps to isolate the failure.
 
 </Warning>
 
+## Chrome local network access permission
+
+<Warning>
+
+Chromium-based browsers block requests from a page served over the public
+internet to a host that resolves to a **private** address, unless the site has
+the **Local network access** permission. The Cube UI is served from the public
+`https://<customer>.cubecloud.dev` origin while your private API hostname
+resolves to an RFC 1918 address over the VPN — so every browser that loads the
+Cube UI needs this permission before charts, the Semantic Model IDE, the
+playground, and any other data-driven view will render.
+
+</Warning>
+
+This is a browser policy that Cube cannot grant on your behalf, and it applies
+to both TLS options above: the browser decides from the **resolved IP address**,
+not the hostname, so reusing `<cube-region>.cubecloudapp.dev` does not avoid it.
+Non-browser clients — `curl`, `psql`, your application servers, BI gateways — are
+unaffected, which is why `curl` from a laptop can succeed while the Cube UI on
+the same laptop shows **"Network Error"**.
+
+Chrome enforces this as [Local Network Access][chrome-lna] from Chrome 142
+onwards; earlier versions enforced its predecessor,
+[Private Network Access][chrome-pna]. A blocked call appears in DevTools as a
+failed request with `ERR_BLOCKED_BY_PRIVATE_NETWORK_ACCESS_CHECKS`, or as a
+CORS-style console error naming a private or local target address space.
+
+**Individual users** can accept Chrome's *"wants to look for and connect to any
+device on your local network"* prompt, or set **Local network access** to
+**Allow** under the padlock icon → **Site settings** on the Cube UI origin.
+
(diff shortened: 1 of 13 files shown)


bsod90:
Superseded by #11350, which lands the Chrome local network access section for AWS on its own.

Splitting this because the Azure and GCP Private API Connectivity pages here still need diagrams, and the Azure page documents a pattern we have not built producer-side yet (no customer-facing API-ingress Private Link Service in `cube-cloud-infra` — only the cube-store PLS producer and the consumer side). Both are being tracked as follow-up work.

Closing rather than merging partially; the drafted Azure and GCP pages stay on the `maxim/docs-chrome-local-network-access` branch and will be picked up from there.

PR #11351, opened 2026-07-24
feat(cube-cli): expose the dev-mode branch name and document dev-mode-only data-model writes

paveltiunov:
**Check List**

Entering dev mode on Cube Cloud forks a personal `dev-…` branch off the branch you name, and data-model file writes only land on that branch — but the CLI never showed it, so `data-model put --branch <feature-branch>` targeted branches where writes are ignored. Pairs with the Cube Cloud API change that returns the real dev branch from `POST .../dev-mode` / `POST .../branches` and rejects file writes to non-dev-mode branches.

- `cube data-model dev-mode <branch>` and `cube data-model create-branch --dev-mode` now print the actual dev-mode branch from the response, e.g. `✓ Entered dev mode on dev-pavel-ab12cd34 (forked from feature-x)`, with a hint to pass it via `--branch` (omitting `--branch` still uses your active dev-mode session). `--json` exposes the raw response.
- `put` / `delete` / `rename` help text now states writes require a dev-mode branch (from `dev-mode`; defaults to your active dev-mode branch).
- README: documented the dev-mode-only write workflow in the `data-model` command group.

Verified with `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo test` (all green). Output falls back to the old message when the server doesn't return a branch name, so the CLI stays compatible with older APIs.

https://claude.ai/code/session_019fUkRZ53YkiVnr4iYrMgQa

---
_Generated by [Claude Code](https://claude.ai/code/session_019fUkRZ53YkiVnr4iYrMgQa)_

### DIFF
--- a/rust/cube-cli/README.md
+++ b/rust/cube-cli/README.md
@@ -132,7 +132,7 @@ Every endpoint of the Console Server public API is covered:
 | `regions` | list available deployment regions |
 | `logs` | tail deployment pod logs (`--pod`, `-c/--container`; defaults to the Cube API container) |
 | `github` (`gh`) | status, installations, repos, branches, connect (import a repo into a deployment + first build) |
-| `data-model` (`dm`) | list, get, put, delete, rename files; branches, create-branch, dev-mode, exit-dev-mode, commit, pull |
+| `data-model` (`dm`) | list, get, put, delete, rename files; branches, create-branch, dev-mode, exit-dev-mode, commit, pull. File writes only land on a **dev-mode branch**: `dev-mode <branch>` forks a personal `dev-…` branch and prints its name — pass that via `--branch` (or omit `--branch` to use your active dev-mode branch); puts to any other branch are rejected by the API |
 | `environments` | list, tokens, create-token (incl. `--meta-sync`) |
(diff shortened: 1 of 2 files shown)


paveltiunov:
Thanks — addressed the review nits:

- **Nit 2 (duplicated hint string)**: extracted into `dev_branch_hint()` in d423938, used by both `dev-mode` and `create-branch --dev-mode`.
- **Nit 1 (response shape)**: confirmed rather than assumed. The raw server response from a tenant running the paired Cube Cloud change is top-level, no `data` wrapper: `{"branchName":"dev-…","parentBranch":"…","active":true}`. The existing `is_empty()` fallback already fails open to the pre-change message if that ever changes.
- **Nit 3 (single `from`/`to` on rename)**: leaving as is — the CLI has never exposed multi-rename, and the array body is now in place if bulk rename comes up.

On the testing note: the whole `data-model` write surface was exercised end to end against a live tenant on this branch — put/get/rename/delete on a dev-mode branch, plus the rejection paths (write to a shared feature branch → 400, unknown branch → 404). That is how the two bugs in 2cff5e0 were found. An automated smoke test would be a reasonable follow-up.

---

PR #11370, opened 2026-07-27
fix(athena-driver): Correct SQL parameter escaping

ovr:


packages/cubejs-athena-driver/src/AthenaDriver.ts
```diff
@@ -588,12 +588,7 @@ export class AthenaDriver extends BaseDriver implements DriverInterface {
   }
 
   protected async startQuery(query: string, values: unknown[]): Promise<AthenaQueryId> {
-    const queryString = applyParams(
-      query,
-      (values || []).map(s => (typeof s === 'string' ? {
-        toSqlString: () => SqlString.escape(s).replace(/\\\\([_%])/g, '\\$1').replace(/\\'/g, '\'\'')
```
ovr, line 594: Similar issue to what I've fixed in trino/presto

waralexrom (approved):

PR #11376, opened 2026-07-27
feat(query-orchestrator): add trace_id SQL comment to data source queries

igorlukanin:
## Summary

- Adds an opt-in `trace_id` SQL comment to the queries Cube sends to a data source, so a warehouse's query history can be joined back to Cube's own query history. Off by default, enabled with `CUBEJS_SQL_INCLUDE_TRACE_ID=true`. Works for every driver, since it operates on the SQL text rather than on a driver-specific query id.
- The comment is **appended**, not prepended: Snowflake strips leading comments before recording the SQL text in `QUERY_HISTORY`, so a leading comment is silently lost on exactly the engine this is most useful for. Verified live against Snowflake — the same query prepended loses the comment and appended retains it. No supported engine requires a leading comment, so appending is both simpler and more correct than per-dialect placement.
- Tagging happens at execution time, **after** the cache key is computed. The SQL text *is* the cache key, so tagging any earlier would give every request a unique key and defeat both result caching and in-flight query coalescing.
- Only the query the user actually asked for is tagged (`primaryQuery`). Refresh keys, pre-aggregation helper queries and background renewals travel the same path but answer to no API request, so they stay untagged.
- The request id can be client-supplied (the REST API reads `x-request-id`/`traceparent` verbatim), so it is sanitized with a character allowlist. Both comment delimiters are stripped: engines disagree on whether block comments nest, so a stray closing delimiter would leak live SQL on some and a stray opening one would swallow the statement on others.
- The emitted `trace_id` deliberately drops the `-span-N` suffix so it matches the `trace_id` column of the Query History export, and the span is emitted as a separate field to tell apart the several data source queries one request fans out into.

## Test plan

- [x] Unit tests for comment construction, sanitization and injection, including a hostile request id that attempts to escape the comment (17 tests)
- [x] Unit tests for the orchestrator wiring: flag off/on, cache-key stability, tuple-form queries, and that refresh-key/background-renewal queries stay untagged (10 tests)
- [x] Unit test for the new environment variable
- [x] Full pre-existing `cubejs-query-orchestrator` unit suite green (78 tests), covering the queued and streaming paths
- [x] Verified live end-to-end on Snowflake (`QUERY_HISTORY`), Postgres (server log) and DuckDB (driver boundary): the main query is tagged, streaming queries are tagged with the correct SQL text, refresh-key queries are untagged, and results are unchanged
- [ ] CI must pass

### DIFF
--- a/docs-mintlify/admin/connect-to-data/query-tracing.mdx
+++ b/docs-mintlify/admin/connect-to-data/query-tracing.mdx
@@ -68,9 +68,14 @@ Otherwise Cube generates one.
 
 ## Scope
 
-Only the query serving an API request is tagged. Pre-aggregation builds, export
-bucket operations, refresh keys, background refreshes, and Cube's own internal
-queries are left untagged, since there is no API request to trace them back to.
+Queries issued to serve an API request are tagged. Pre-aggregation builds,
+export bucket operations, refresh keys, background refreshes, and Cube's own
+internal queries are left untagged, since there is no API request to trace them
+back to.
+
+A single request usually produces one tagged query, but not always — a query
+reading a lambda pre-aggregation's source table, for instance, is tagged too.
+Match on `trace_id` rather than assuming one row per request.
 
 Not every request produces a tagged query to find: one served from the cache
 never reaches your data source, and when identical queries arrive together Cube
--- a/docs-mintlify/admin/connect-to-data/query-tracing.mdx
+++ b/docs-mintlify/admin/connect-to-data/query-tracing.mdx
@@ -1,0 +1,92 @@
+---
+title: Query tracing
+description: Trace a Cube API request through to the queries it runs on your data source.
+---
+
+Cube can tag the queries it sends to your data source with a `trace_id` SQL
+comment, letting you match a query in your data source's query history back to
+the API request that caused it.
+
+This is useful when one API request fans out into several data source
+queries — every one of them carries the same `trace_id`, distinguished by a
+`-span-N` suffix.
+
+## Configuration
+
+Set [`CUBEJS_SQL_INCLUDE_TRACE_ID`](/reference/configuration/environment-variables#cubejs_sql_include_trace_id)
+to `true`:
+
+```dotenv
+CUBEJS_SQL_INCLUDE_TRACE_ID=true
+```
+
+Queries then reach your data source with the comment appended:
+
+```sql
+SELECT ...
+/* trace_id: f47ac10b-58cc-4372-a567-0e02b2c3d479 span: 1 */
+```
+
+`trace_id` identifies the API request and matches the field of the same name in
+the [Query History export][ref-query-history-export]. `span` identifies the
+attempt within that request.
+
+<Warning>
+
+A unique comment per request makes each query textually distinct, which
+prevents your data source from reusing cached results — Snowflake's persisted
+query results and BigQuery's result cache both require identical query text.
+Expect more queries to actually execute, and on Snowflake a warehouse that
+would have stayed suspended may resume.
+
+</Warning>
+
+## Finding the comment in your data source
+
+Some query history views truncate long SQL, and a few omit comments entirely.
+If you can't find the `trace_id`, check where you're reading it from:
+
+| Data source | Where to look |
+| ----------- | ------------- |
+| Snowflake | `QUERY_HISTORY`. Note that Snowflake strips *leading* comments, which is why Cube appends. |
+| Redshift | `SYS_QUERY_TEXT`, reassembled with `LISTAGG`. `SYS_QUERY_HISTORY` and `STL_QUERY` truncate at 4000 characters. |
+| BigQuery | `INFORMATION_SCHEMA.JOBS`. Cloud Audit Logs truncate at 50K characters. |
+| Databricks | `system.query.history`. Empty when customer-managed keys are enabled. |
+| Postgres | Server logs (`log_statement`). `pg_stat_activity.query` truncates at `track_activity_query_size` (1024 bytes by default). |
+| MySQL | The general or slow query log. |
+
+<Warning>
+
+`pg_stat_statements` and MySQL's digest tables normalize queries and discard
+comments, keeping only the text of the first query with a given fingerprint.
+The `trace_id` you see there belongs to some earlier request, not the one you
+are looking at. Use the query logs instead.
+
+</Warning>
+
+## Supplying your own identifier
+
+Clients calling the [REST (JSON) API](/reference/core-data-apis/rest-api) can
+supply their own identifier through the `x-request-id` or `traceparent` request
+header, letting you trace a query back to a request in your own application.
+Otherwise Cube generates one.
+
+## Scope
+
+Only the query serving an API request is tagged. Pre-aggregation builds, export
+bucket operations, refresh keys, background refreshes, and Cube's own internal
+queries are left untagged, since there is no API request to trace them back to.
+
+Not every request produces a tagged query to find: one served from the cache
+never reaches your data source, and when identical queries arrive together Cube
+runs them once, under whichever request got there first.
+
+## Snowflake query tagging
+
+Snowflake also supports a connection-level
+[query tag](/admin/connect-to-data/data-sources/snowflake#query-tagging), which
+labels every query on a connection with the same static value. The two are
+complementary: a query tag identifies the deployment, while `trace_id`
+identifies the individual request.
+
+[ref-query-history-export]: /admin/monitoring/monitoring-integrations#query-history-export
--- a/docs-mintlify/admin/connect-to-data/query-tracing.mdx
+++ b/docs-mintlify/admin/connect-to-data/query-tracing.mdx
@@ -8,8 +8,7 @@ comment, letting you match a query in your data source's query history back to
 the API request that caused it.
 
 This is useful when one API request fans out into several data source
-queries — every one of them carries the same `trace_id`, distinguished by a
-`-span-N` suffix.
+queries — every one of them carries the same `trace_id`.
 
 ## Configuration
 
@@ -24,12 +23,14 @@ Queries then reach your data source with the comment appended:
 
 ```sql
 SELECT ...
-/* trace_id: f47ac10b-58cc-4372-a567-0e02b2c3d479 span: 1 */
+/* trace_id: f47ac10b-58cc-4372-a567-0e02b2c3d479 */
 ```
 
 `trace_id` identifies the API request and matches the field of the same name in
-the [Query History export][ref-query-history-export]. `span` identifies the
-attempt within that request.
+the [Query History export][ref-query-history-export].
+
+Trace ids longer than 128 characters are truncated, and any character outside
+`A-Z a-z 0-9 . _ : -` is removed.
 
 <Warning>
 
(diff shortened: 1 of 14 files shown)


docs-mintlify/admin/connect-to-data/query-tracing.mdx
```diff
@@ -0,0 +1,92 @@
...
+
+```dotenv
+CUBEJS_SQL_INCLUDE_TRACE_ID=true
+```
+
+Queries then reach your data source with the comment appended:
+
+```sql
+SELECT ...
+/* trace_id: f47ac10b-58cc-4372-a567-0e02b2c3d479 span: 1 */
```
igorlukanin: `span: 1` is totally unnecessary, remove.

docs-mintlify/admin/connect-to-data/query-tracing.mdx
```diff
@@ -0,0 +1,92 @@
...
+If you can't find the `trace_id`, check where you're reading it from:
+
+| Data source | Where to look |
+| ----------- | ------------- |
+| Snowflake | `QUERY_HISTORY`. Note that Snowflake strips *leading* comments, which is why Cube appends. |
+| Redshift | `SYS_QUERY_TEXT`, reassembled with `LISTAGG`. `SYS_QUERY_HISTORY` and `STL_QUERY` truncate at 4000 characters. |
+| BigQuery | `INFORMATION_SCHEMA.JOBS`. Cloud Audit Logs truncate at 50K characters. |
+| Databricks | `system.query.history`. Empty when customer-managed keys are enabled. |
+| Postgres | Server logs (`log_statement`). `pg_stat_activity.query` truncates at `track_activity_query_size` (1024 bytes by default). |
+| MySQL | The general or slow query log. |
```
igorlukanin: Reformat as list.

docs-mintlify/admin/connect-to-data/query-tracing.mdx
```diff
@@ -0,0 +1,92 @@
...
+
+<Warning>
+
+`pg_stat_statements` and MySQL's digest tables normalize queries and discard
+comments, keeping only the text of the first query with a given fingerprint.
+The `trace_id` you see there belongs to some earlier request, not the one you
+are looking at. Use the query logs instead.
+
+</Warning>
```
igorlukanin: Remove.

docs-mintlify/admin/connect-to-data/query-tracing.mdx
```diff
@@ -0,0 +1,92 @@
...
+never reaches your data source, and when identical queries arrive together Cube
+runs them once, under whichever request got there first.
+
+## Snowflake query tagging
+
+Snowflake also supports a connection-level
+[query tag](/admin/connect-to-data/data-sources/snowflake#query-tagging), which
+labels every query on a connection with the same static value. The two are
+complementary: a query tag identifies the deployment, while `trace_id`
+identifies the individual request.
```
igorlukanin: Remove from here.

packages/cubejs-base-driver/src/trace-comment.ts
```diff
@@ -0,0 +1,90 @@
...
+ */
+function splitRequestId(requestId: string): { traceId: string, span?: string } {
+  const idx = requestId.lastIndexOf('-span-');
+  if (idx === -1) {
+    return { traceId: requestId };
+  }
+
+  return {
+    traceId: requestId.substring(0, idx),
+    span: requestId.substring(idx + '-span-'.length),
```
igorlukanin: We don't need spans in this PR at all.

packages/cubejs-base-driver/src/trace-comment.ts
```diff
@@ -0,0 +1,90 @@
...
+    return null;
+  }
+
+  const { traceId, span } = splitRequestId(sanitized);
+  if (!traceId) {
+    return null;
+  }
+
+  return span
+    ? `/* trace_id: ${traceId} span: ${span} */`
```
igorlukanin: No spans, just trace ids.

docs-mintlify/admin/connect-to-data/query-tracing.mdx
igorlukanin: Removed.

docs-mintlify/admin/connect-to-data/query-tracing.mdx
igorlukanin: Done.

docs-mintlify/admin/connect-to-data/query-tracing.mdx
igorlukanin: Removed.

docs-mintlify/admin/connect-to-data/query-tracing.mdx
igorlukanin: Removed — the pointer from the Snowflake query tag page to this one stays, so the two are still connected.

packages/cubejs-base-driver/src/trace-comment.ts
igorlukanin: Spans are gone from the comment — it's just `trace_id` now. Still stripping the `-span-N` suffix off the request id, since the Query History export's `trace_id` doesn't carry it and the join wouldn't match otherwise.

packages/cubejs-base-driver/src/trace-comment.ts
igorlukanin: Done.

packages/cubejs-query-orchestrator/src/orchestrator/QueryCache.ts
```diff
@@ -406,6 +409,45 @@ export class QueryCache {
...
+    renewCycle?: boolean,
+    primaryQuery?: boolean,
+  }): any {
+    if (
+      !getEnv('sqlIncludeTraceId') ||
+      !req.requestId ||
+      // Only the query the user asked for. Refresh keys, pre-aggregation
+      // helpers and background renewals ride the same path but answer to no
+      // API request.
+      !req.primaryQuery ||
```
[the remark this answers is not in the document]
igorlukanin: Good catch — real bug. `cacheMode` comes straight off the client (`req.query.cache`), so anyone sending `cache=stale-while-revalidate` got untagged SQL and the join silently didn't work. Added `primaryQuery: true` on that branch plus a test that fails without the fix.

packages/cubejs-query-orchestrator/src/orchestrator/QueryCache.ts
```diff
@@ -406,6 +409,45 @@ export class QueryCache {
...
+      // Independent guards, deliberately: `renewCycle` catches a refresh of a
+      // user's query, the prefix catches one the scheduler started itself.
+      req.renewCycle ||
```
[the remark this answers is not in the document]
igorlukanin: Agreed, and the wording was the wrong half. The lambda source-table read genuinely is a query issued for that request, so I'd rather tag it than suppress it — fixed the claim instead: the docs now say queries issued to serve a request are tagged, and that one request can produce more than one, so match on `trace_id` rather than expecting a single row.

docs-mintlify/admin/connect-to-data/query-tracing.mdx
```diff
@@ -0,0 +1,84 @@
...
+internal queries are left untagged, since there is no API request to trace them
+back to.
+
+A single request usually produces one tagged query, but not always — a query
+reading a lambda pre-aggregation's source table, for instance, is tagged too.
+Match on `trace_id` rather than assuming one row per request.
+
+Not every request produces a tagged query to find: one served from the cache
+never reaches your data source, and when identical queries arrive together Cube
+runs them once, under whichever request got there first.
```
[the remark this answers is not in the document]
igorlukanin: Right, and pre-aggregations are the case people will hit first. Added them to that sentence in `12d2a0a`.

packages/cubejs-base-driver/src/trace-comment.ts
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * The emitted id has to match the `trace_id` of the Query History export, which
+ * carries the request id without that suffix — joining on the full request id
+ * would never match.
+ */
+function toTraceId(requestId: string): string {
+  const idx = requestId.lastIndexOf('-span-');
+
+  return idx === -1 ? requestId : requestId.substring(0, idx);
+}
```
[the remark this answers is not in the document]
igorlukanin: Agreed, that one has to stay in sync. `toTraceId` is now exported from here and `utils.ts` re-exports it as `extractRequestUUID`, so there is a single implementation in `fb69fd7`.

igorlukanin:
Fair on the nit: added a direct `toTraceId` block in `2957448`, pinning the non-numeric span suffix specifically.

igorlukanin:
Took the nit: `addTraceComment` now collapses the whole trailing run in `c12ea2c`. Pinned from both sides while I was there, since the obvious wider regex is wrong in a way the suite didn't catch — matching trailing whitespace without a leading `;` appends a separator to a query that never had one.

igorlukanin:
CodeQL caught that the widened regex was a polynomial ReDoS — `;[\s;]*$` backtracks quadratically on a long run of semicolons that never reaches the anchor (~1.8s at 60k chars, and `sql` is uncontrolled input). Replaced it with a linear backward scan in `81a751a`, with a timing assertion so it can't come back as a regex.

packages/cubejs-base-driver/src/trace-comment.ts
```diff
@@ -0,0 +1,94 @@
...
+  // Take the whole trailing run, not one semicolon: a leftover `;` would leave the
+  // comment between two statements, the second empty. Scanned, not matched — `sql`
+  // is uncontrolled and `;[\s;]*$` backtracks quadratically on a long run.
+  let end = sql.length;
+  let sawSemicolon = false;
+  while (end > 0) {
+    const ch = sql[end - 1];
+    if (ch === ';') {
+      sawSemicolon = true;
+    } else if (!/\s/.test(ch)) {
```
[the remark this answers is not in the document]
igorlukanin: Hoisted to a module-level `WHITESPACE` in `6bcc239`, matching the `ALLOWED_CHARS` convention already in the file — keeping `\s` rather than the explicit set, since narrowing it would change behaviour.

On the cost though: I measured it and V8 caches the literal, so inline and hoisted both come out at 0.54ms/iter over a 200k run — no allocation difference to recover. Worth doing as intent, not as a fix. Also note the scan only reaches the matcher on non-`;` characters, so a semicolon run never touches it; the case that does is a long whitespace run, which the linearity test now covers as well.

packages/cubejs-base-driver/test/unit/trace-comment.test.ts
```diff
@@ -0,0 +1,172 @@
...
+    // Guards against going back to `;[\s;]*$`, which backtracks quadratically
+    // when the run never reaches the anchor — ~1.8s at 60k on uncontrolled input.
+    // Both shapes: semicolons, and the whitespace the scan actually tests.
+    for (const filler of [';'.repeat(60_000), ' '.repeat(60_000)]) {
+      const runaway = `SELECT 1${filler}x`;
+      const started = Date.now();
+      expect(addTraceComment(runaway, 'abc-123')).toBe(`${runaway}\n/* trace_id: abc-123 */`);
+      expect(Date.now() - started).toBeLessThan(1000);
+    }
+  });
```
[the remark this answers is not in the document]
igorlukanin: Correct, and I confirmed it by counting iterations: both `…x` shapes run the loop exactly once, only a run reaching end-of-string runs 60,001. Kept the `…x` case — it is still the ReDoS guard, since that is the input the regex backtracks on — and added the complement in `c7df3db`, both fillers, asserting the collapse and the leave-whitespace-alone branch.

It earns its place on the red-check too: the `…x` case only reddens the quadratic-regex mutant, while the new one reddens the single-`;` strip, the dropped-guard variant and the no-op as well.

packages/cubejs-base-driver/src/trace-comment.ts
```diff
@@ -0,0 +1,97 @@
...
+export function toTraceId(requestId: string): string {
+  const idx = requestId.lastIndexOf('-span-');
+
+  return idx === -1 ? requestId : requestId.substring(0, idx);
+}
+
+export function buildTraceComment(requestId: string | undefined | null): string | null {
+  // Cap only after the span is stripped: capping the raw id first can cut it
+  // mid-`-span-`, leaving a partial marker in what gets emitted.
+  const traceId = toTraceId(sanitizeTraceId(requestId)).slice(0, MAX_TRACE_ID_LENGTH);
```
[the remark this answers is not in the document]
igorlukanin: Good catch — the export strips the raw id, so sanitizing first could synthesize a marker that side never sees. Reordered to strip-then-sanitize (cap stays last) in `55565ed`, with a regression test red-checked against both wrong orderings.

PR #11382, opened 2026-07-27
docs: DAX single-perspective env var and measure names

igorlukanin:
## Summary

Follow-up to the DAX single-perspective mode docs (#10891, #10901), covering two gaps:

- `CUBEJS_DAX_SINGLE_PERSPECTIVE` was missing from the environment variables reference, even though its sibling `CUBEJS_DAX_CREATE_DATE_HIERARCHIES` is documented there. Added with its `false` default and a link to the cross-view filtering page.
- The `<view>.<measure>` measure naming used in single-perspective mode was not documented anywhere. Added a short "Measure names" section to the cross-view filtering page: field names in Power BI are unaffected, but explicit DAX must reference measures by their qualified name.

Both pages already existed, so there are no new pages and no `docs.json` changes.

## Test plan

- Behavior verified against the implementation rather than assumed: the env var name and its `false` default, the `<view>.<measure>` separator, that Power BI captions stay unprefixed, and that both the qualified and unqualified reference forms resolve.
- `mint dev` — both changed pages plus the DAX API index render (HTTP 200), no MDX errors.
- `mint broken-links` — no broken links, including the two new link references.

### DIFF
--- a/docs-mintlify/reference/configuration/environment-variables.mdx
+++ b/docs-mintlify/reference/configuration/environment-variables.mdx
@@ -1578,6 +1578,15 @@ If `true`, the DAX API will expose time dimensions as calendar hierarchies.
 | --------------- | ---------------------- | --------------------- |
 | `true`, `false` | `true` | `true` |
 
+## `CUBEJS_DAX_SINGLE_PERSPECTIVE`
+
+If `true`, the [DAX API][ref-dax-api] exposes all views as part of a single
+perspective, enabling [cross-view filtering][ref-dax-cross-view-filter].
+
+| Possible Values | Default in Development | Default in Production |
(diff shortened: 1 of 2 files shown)


igorlukanin:
Fixed all three nits: the `CALCULATE` example now takes a filter argument, the table-qualifier wording matches the example, and the stray blank line is gone. Also linked the env var mention at the top of the page to the new reference anchor, mirroring `dax-api/index.mdx`.

On the default: the implementation lives in cube-runtime, not this repo, which is why the grep came up empty. It reads `env_parse("CUBEJS_DAX_SINGLE_PERSPECTIVE", false)` — a single default with no development/production split, so `false`/`false` is right. The sibling defaulting to `true` is a separate code path that builds its value differently.

Issue #10363, opened 2026-02-04
BigQuery driver doesn't pass decimal precision/scale to CubeStore, causing schema mismatch error


lvauvillier:
# BigQuery driver doesn't pass decimal precision/scale to CubeStore, causing schema mismatch error

## Describe the bug

When building pre-aggregations with BigQuery as the source database and CubeStore as the external database, CubeStore fails with a decimal precision mismatch error. This happens because the BigQuery driver doesn't pass precision/scale information to CubeStore, even when `CUBEJS_DB_PRECISE_DECIMAL_IN_CUBESTORE=true` is set.

The root cause is that `BigQueryDriver.tableColumnTypes()` calls `toGenericType(c.type)` without passing precision and scale arguments. The `toGenericType()` method only returns `decimal(precision, scale)` when both precision AND scale are provided - otherwise it returns just `decimal`, which CubeStore then defaults to `Decimal128(18, 5)`.

**Error:**
```
2026-02-04T10:05:47.528Z ERROR [cubestore::cluster::ingestion::job_runner] <pid:23> Error while running job 6: Internal: task 12447 panicked with message "called `Result::unwrap()` on an `Err` value: InvalidArgumentError(\"column types must match schema types, expected Decimal128(18, 5) but found Decimal128(28, 5) at column index 11\")"
```

## To Reproduce

1. Set up Cube with BigQuery as the source database and CubeStore as the external database
2. Set `CUBEJS_DB_PRECISE_DECIMAL_IN_CUBESTORE=true`
3. Create a cube with `SUM` measures on `NUMERIC` columns
4. Create a pre-aggregation that includes these measures
5. Trigger the pre-aggregation build
6. See error in CubeStore logs about decimal precision mismatch

## Expected behavior

When `CUBEJS_DB_PRECISE_DECIMAL_IN_CUBESTORE=true` is set, the BigQuery driver should pass the correct precision and scale for `NUMERIC` (38, 9) and `BIGNUMERIC` (76, 38) types, so CubeStore creates tables with matching decimal precision.

## Minimally reproducible Cube Schema

```yaml
cubes:
  - name: orders
    sql: |
      SELECT 1 as id, CAST(100.123456789 AS NUMERIC) as amount
      UNION ALL
      SELECT 2 as id, CAST(200.987654321 AS NUMERIC) as amount

    measures:
      - name: totalAmount
        sql: amount
        type: sum

    dimensions:
      - name: id
        sql: id
        type: number

[...]

Issue #10289, opened 2026-01-06
ClickHouse enforces timestamptz


mytototo:
**Describe the bug**
Working with `Date` types from ClickHouse doesn't work, and `timestamptz` is always enforced even though it's not supported.

**To Reproduce**
Create a ClickHouse table with a `Date` column:
```sql
CREATE TABLE series (
  date Date,
);
```

Create Cube.js model, as pasted below.

Then, query the date column, for example:
```json
{
  "dimensions": [
    "series.date"
  ]
}
```

You will see an error:
```
Query failed: Error: Unknown data type family: timestamptz. Maybe you meant: ['TIMESTAMP']:
```

Indeed, the generated SQL is as follow:
```sql
SELECT
  timezone('UTC', "series".date :: timestamptz) "series__date"
FROM
  series AS "series"
GROUP BY
  1
ORDER BY
  1 ASC
LIMIT
  10000
```

**Expected behavior**

**Minimally reproducible Cube Schema**

```javascript
cube("Series", {
  data_source: "clickhouse",

  sql: `
    SELECT * FROM series
  `,

  dimensions: {
    date: {
      type: "time",
      sql: "date",
    },
  },
});

```

**Version:**
1.6.1

Issue #7127, opened 2023-09-12
Cube crashes when storing pre-aggregation with duckdb (type `hugeInt`)


hugoJuhel:
**Describe the bug**

Cubestore does not support the `HugeInt` type used by `duckdb`

I have first encountered the bug with the on-prem cube, but I have been able to reproduce it on the `cube cloud`.
Both the cube files and printscreens are comming from the cloud version.

When using Cube with the `duckdb` driver,  Cube fails at creating (external) pre-aggregation 

**To Reproduce**
Steps to reproduce the behavior :
1. Create parquet file on s3 containing at least one integer column
2. Connect cube to the s3 account
3. Create a schema where one measure is a `sum` over the integer column
4. Add the measure a preaggregation
5. The cube will fail when trying to create the pre-aggregation table with a `hugeInt` column to store the data.

I have also created a small cube definition showcasing the bug.

**Expected behavior**
The driver should be casting the `hugeInt` to the corresponding MySql type 

**Screenshots**
Schema :
![image](https://github.com/cube-js/cube/assets/22279443/f00821a9-377b-43c4-a991-5ca93aa44aed)

Playground : 
![image](https://github.com/cube-js/cube/assets/22279443/f86f471a-e1b7-4dd3-a605-b430a7c09327)

Pre-aggregation error : 
![image](https://github.com/cube-js/cube/assets/22279443/bcebf7ec-53a6-4146-a796-419e61619928)

**Stack trace** 
```
Error: Error during create table: CREATE TABLE prod_pre_aggregations.spam_main_w1hqvvny_l2pn4q2e_1ig1fr6 (`spam__org` varchar(255), `spam__metric` HUGEINT) LOCATION ?: Custom type 'HUGEINT' is not supported
    at WebSocket.<anonymous> (/cube/node_modules/@cubejs-backend/cubestore-driver/src/WebSocketConnection.ts:121:30)
    at WebSocket.emit (node:events:513:28)
    at Receiver.receiverOnMessage (/cube/node_modules/ws/lib/websocket.js:1008:20)
    at Receiver.emit (node:events:513:28)
    at Receiver.dataMessage (/cube/node_modules/ws/lib/receiver.js:502:14)
    at Receiver.getData (/cube/node_modules/ws/lib/receiver.js:435:17)
    at Receiver.startLoop (/cube/node_modules/ws/lib/receiver.js:143:22)
    at Receiver._write (/cube/node_modules/ws/lib/receiver.js:78:10)
[...]

Issue #7017, opened 2023-08-09
external  pre_aggregations with cubestore ER_INTERNAL_ERROR: Custom type 'timestamptz' is not supported


iiq374:
**Describe the bug**
I'm using CubeJS connecting Postgres. When integrating Cube Store I got the error with timestamptz type of Postgres

```
ERROR [cubestore::http] <pid:1> Error processing HTTP command: Custom type 'timestamptz' is not supported
```

**To Reproduce**
1. Create a Pre-Aggregation with a TimestampTZ datatype as a dimension
2. Wait for Pre-Aggregation to build
3. Observe error in logs

**Expected behavior**
Postgres Timestamptz should not cause CubeStore to be unusable

**Screenshots**

**Version:**
0.33.43

**Additional context**
Seems related to this earlier issue / related fix:  https://github.com/cube-js/cube/issues/1796

paveltiunov:
@iiq374 There's no timestamptz type in Cube Store, so you should cast it to timestamp first if you're using it as the raw value of a dimension. If you put in a time dimension, it should work just fine.

nurulhudaapon:
@paveltiunov Where should I cast the value? Should I create a new dimension with a value casted to `timestamp` or just casting the existing time dimension will be fine (assuming it wont' affect timezone)?

igorlukanin:
@nurulhudaapon Here's what I believe Pavel meant.

```yaml
    dimensions:
        # It's OK to have a TIMESTAMPZ value as a time dimension, it will work just fine
      - name: created_at
        sql: created_at
        type: time

        # You have to cast it to `TIMESTAMP` if you'd like to use it in a non-`time` dimension
      - name: created_at_raw
        sql: "created_at::TIMESTAMP"
        type: string
```

PR #11404, opened 2026-07-28
fix(snowflake-driver): map TIMESTAMP_TZ/TIMESTAMP_LTZ to generic timestamp so they can be pre-aggregated

### LINKED ISSUE #11403: Snowflake TIMESTAMP_TZ/TIMESTAMP_LTZ cannot be pre-aggregated: type map falls through, Cube Store rejects "Custom type 'timestamp_tz' is not supported"
**Describe the bug**

`TIMESTAMP_TZ` and `TIMESTAMP_LTZ` columns cannot be used as pre-aggregation time dimensions on Snowflake. Building the rollup fails in Cube Store with:

```
Custom type 'timestamp_tz' is not supported
```

The cause is a gap in the driver's type map rather than anything about the data. `SnowflakeToGenericType` in `packages/cubejs-snowflake-driver/src/SnowflakeDriver.ts` maps only `timestamp_ntz`:

```ts
const SnowflakeToGenericType: Record<string, GenericDataBaseType> = {
  object: 'HLL_SNOWFLAKE',
  number: 'decimal',
  fixed: 'decimal',
  timestamp_ntz: 'timestamp'
};
```

and `BaseDriver.toGenericType()` falls through by returning the input unchanged:

```ts
const genericType = DbTypeToGenericType[columnType.toLowerCase()] || columnType;
```

`DbTypeToGenericType` has no `timestamp_tz` / `timestamp_ltz` entry either, so the raw Snowflake type name is passed straight to Cube Store as if it were a generic type. Both driver paths that type a pre-aggregation column hit this — `queryColumnTypes()` (where `INFORMATION_SCHEMA.COLUMNS.DATA_TYPE` yields `TIMESTAMP_TZ`) and the snowflake-sdk path (`column.getType()` yields `timestamp_tz`).

Reproduced against `v1.7.8`:

```
 type                 toGenericType()
 TIMESTAMP_NTZ        timestamp
 TIMESTAMP_TZ         TIMESTAMP_TZ      <-- passed through verbatim
 TIMESTAMP_LTZ        TIMESTAMP_LTZ     <-- passed through verbatim
```

**To Reproduce**

1. Point Cube at Snowflake with a table whose time column is `TIMESTAMP_TZ` (the default for FOCUS/FinOps billing exports, and for anything derived from `CURRENT_TIMESTAMP`).
2. Declare a `rollup` pre-aggregation with `time_dimension` set to that column and any `granularity`.
3. Build it. Cube Store rejects the `CREATE TABLE`.

**Expected behavior**

All three `TIMESTAMP` variants should reach Cube Store as the generic `timestamp`. They are all UTC instants, and since #11010 the driver pins `TIMEZONE = 'UTC'` via `ALTER SESSION` on every connection, so all three render deterministically as UTC.

**The value semantics are already correct — only the type map is wrong**

This is worth emphasising because it makes the fix a one-liner rather than a behavioural change:

- `packages/cubejs-snowflake-driver/src/type-parsers.ts` **already** lists `timestamp_tz` and `timestamp_ltz` in its hydrator and normalises both through `formatUtcTimestamp`.
- The existing test `packages/cubejs-snowflake-driver/test/SnowflakeDriver.test.ts` **already asserts** that a `TIMESTAMP_TZ` column hydrates to exactly the same string as the equivalent `TIMESTAMP_NTZ` column.

So the driver already treats `TIMESTAMP_TZ` as a UTC timestamp for query results, and simply never told the type map. The asymmetry is the bug.

**Cube Store handles the offset correctly**

I verified this directly against `cubestored` v1.7.8, through both `INSERT` and the CSV-location ingest path that the export bucket uses, since the Snowflake unload writes `TIMESTAMP_FORMAT = 'YYYY-MM-DD"T"HH24:MI:SS.FF3TZH:TZM'` (offset included):

| Ingested | Stored |
|---|---|
| `2025-08-02T17:00:00.000` (no offset) | `2025-08-02T17:00:00.000` |
| `2025-08-02T17:00:00.000+00:00` | `2025-08-02T17:00:00.000` |
| `2025-08-02T17:00:00.000-07:00` | `2025-08-03T00:00:00.000` |
| `2025-08-02T17:00:00.000+05:30` | `2025-08-02T11:30:00.000` |

`date_trunc('day', ts)` grouped correctly in every case. Cube Store's parser is offset-*aware* — it converts to UTC rather than truncating the wall clock — so nothing downstream needs to change and a `+00:00` value is byte-equivalent to the `TIMESTAMP_NTZ` case that works today.

**Prior art — this is a known, recurring failure class, and the remedy has always been the same**

The "unmapped type reaches Cube Store verbatim" bug has been fixed repeatedly by adding the missing entry:

- #1796 — `Custom type 'int8' is not supported`, the original instance. Fixed by adding `int8: 'bigint'` to `DbTypeToGenericType`.
- #2592 / #2601 — Snowflake pre-aggregation export broken; fixed by adding **`timestamp_ntz: 'timestamp'` to this very map**. This is the direct precedent: the same file, the same one-line shape, for the sibling type. `timestamp_tz`/`timestamp_ltz` were simply never added alongside it.
- #6458 — the Firebolt driver adding `timestamptz` support.
- #10175 — extending base-driver type support so Cube Store data exports carry numeric precision/scale.

The same class is still being reported against other drivers, which suggests the verbatim fall-through is a persistent footgun rather than a one-off:

- #7017 — `external pre_aggregations with cubestore ER_INTERNAL_ERROR: Custom type 'timestamptz' is not supported` (Postgres). **Identical error signature to this report.**
- #5277 — `Custom type 'timestamptz' is not supported`.
- #7127 — *(still open)* Cube crashes storing a pre-aggregation with DuckDB `hugeInt`.
- #10363 — BigQuery driver doesn't pass decimal precision/scale to Cube Store, causing a schema mismatch.
- #10289 — *(open)* ClickHouse enforces `timestamptz`.
- #5849 — Athena "Casting a Timestamp with Time Zone to Timestamp is not supported".

**Workarounds, and why they're bad**

Without this fix the only options are to cast in the cube's dimension SQL:

```js
charge_period_start: {
  sql: `CAST(CONVERT_TIMEZONE('UTC', charge_period_start) AS TIMESTAMP_NTZ)`,
  type: `time`,
}
```

or to re-materialise the column as `TIMESTAMP_NTZ` upstream. The cast is costly: wrapping the column defeats precise partition pruning against a clustering key on it. Measured with `EXPLAIN USING JSON` on a 410M-row / 2061-partition table clustered by the time column:

| Filter range | Partitions (bare column) | Partitions (cast) | |
|---|---|---|---|
| 1 day | 66 | 105 | +59% |
| 1 week | 111 | 138 | +24% |
| 2 weeks | 63 | 90 | +43% |
| 1 month | 126 | 167 | +33% |
| 3 months | 591 | 617 | +4% |

So every narrow interactive range filter scans up to ~1.6x the partitions purely to work around the type map. Re-materialising upstream avoids that but means rewriting models and rebuilding history.

**Version:** 1.7.8 (also present on `master` at time of filing)

**Additional context**

While confirming the above I noticed `real` has the same gap — it appears in the `type-parsers.ts` hydrator alongside `fixed`, but is absent from both type maps, so `toGenericType('real')` returns `'real'`. I've left that out of the linked PR to keep it focused on the timestamps; happy to add it if you'd prefer one change.

npwalker:
**Check List**

Fixes #11403

`TIMESTAMP_TZ` and `TIMESTAMP_LTZ` columns can't be used as pre-aggregation time dimensions on Snowflake — the rollup build fails in Cube Store with `Custom type 'timestamp_tz' is not supported`.

`SnowflakeToGenericType` maps only `timestamp_ntz`, and `BaseDriver.toGenericType()` falls through as `DbTypeToGenericType[columnType.toLowerCase()] || columnType` — i.e. unknown types are returned **verbatim**. `DbTypeToGenericType` has no `timestamp_tz`/`timestamp_ltz` entry either, so the raw Snowflake type name is handed to Cube Store as though it were a generic type. Both paths that type a pre-aggregation column are affected: `queryColumnTypes()` (`INFORMATION_SCHEMA.COLUMNS.DATA_TYPE` → `TIMESTAMP_TZ`) and the snowflake-sdk path (`column.getType()` → `timestamp_tz`). Both normalise through `toLowerCase()`, so one map entry covers each.

This adds the two missing entries:

```diff
   fixed: 'decimal',
-  timestamp_ntz: 'timestamp'
+  timestamp_ntz: 'timestamp',
+  timestamp_ltz: 'timestamp',
+  timestamp_tz: 'timestamp'
 };
```

**Why mapping all three to the same generic type is correct**

This is a type-map omission, not a change in value semantics:

- `src/type-parsers.ts` already lists `timestamp_tz` and `timestamp_ltz` in its hydrator and normalises both via `formatUtcTimestamp`.
- `test/SnowflakeDriver.test.ts` already asserts that a `TIMESTAMP_TZ` column hydrates to the *same string* as the equivalent `TIMESTAMP_NTZ` column.
- Since #11010, `initConnection()` pins `TIMEZONE = 'UTC'` on every connection, so all three variants render deterministically as UTC.

The driver already treats `TIMESTAMP_TZ` as a UTC timestamp everywhere except the type map.

I also confirmed Cube Store handles the offset the unload emits (`TIMESTAMP_FORMAT = 'YYYY-MM-DD"T"HH24:MI:SS.FF3TZH:TZM'`). Against `cubestored` v1.7.8, via both `INSERT` and the CSV-location ingest path used by the export bucket:

| Ingested | Stored |
|---|---|
| `2025-08-02T17:00:00.000` (no offset) | `2025-08-02T17:00:00.000` |
| `2025-08-02T17:00:00.000+00:00` | `2025-08-02T17:00:00.000` |
| `2025-08-02T17:00:00.000-07:00` | `2025-08-03T00:00:00.000` |
| `2025-08-02T17:00:00.000+05:30` | `2025-08-02T11:30:00.000` |

`date_trunc('day', ts)` grouped correctly throughout — the parser converts offsets to UTC rather than truncating wall-clock time, so a `+00:00` value is byte-equivalent to the `TIMESTAMP_NTZ` case that already works.

**Precedent**

`timestamp_ntz: 'timestamp'` was itself added to this exact map to fix the same class of breakage (#2592 / #2601); `timestamp_tz` and `timestamp_ltz` were never added alongside it. Same remedy in #1796 (`int8: 'bigint'`), #6458 (Firebolt `timestamptz`) and #10175. The class is still open elsewhere: #7017 and #5277 (identical Cube Store error, Postgres), #7127 (DuckDB `hugeInt`), #10363 (BigQuery precision/scale), #10289 (ClickHouse).

**Tests**

The `stream` integration test in `test/SnowflakeDriver.test.ts` now asserts `tableData.types`, so each of `TIMESTAMP_NTZ` / `TIMESTAMP_LTZ` / `TIMESTAMP_TZ` is checked to reach Cube Store as the generic `timestamp` rather than its raw Snowflake name (per review, this replaced an earlier unit-test file).

`test/SnowflakeDriver.test.ts` gains `TIMESTAMP_TZ` and `TIMESTAMP_LTZ` columns carrying a **non-UTC** offset (`-0700`), asserting they hydrate to the UTC instant (`13:07:42.123 -07:00` → `20:07:42.123`) rather than the wall clock. That pins down the property that makes the shared mapping correct.

One caveat, in the interest of transparency: I have no Snowflake account to run the integration suite against, so `test/SnowflakeDriver.test.ts` has **not** been executed locally — the expected hydration values are derived from `formatUtcTimestamp` formatting the UTC components of the SDK's `Date`, and the expected `types` from `getTypes()` / `toGenericType()`. Please flag if CI disagrees and I'll correct them.

**Note**

`real` has the same gap — present in the `type-parsers.ts` hydrator next to `fixed`, absent from both type maps, so `toGenericType('real')` returns `'real'`. Left out to keep this focused; happy to fold it in.

### DIFF
--- a/packages/cubejs-snowflake-driver/src/SnowflakeDriver.ts
+++ b/packages/cubejs-snowflake-driver/src/SnowflakeDriver.ts
@@ -53,7 +53,16 @@ const SnowflakeToGenericType: Record<string, GenericDataBaseType> = {
   // For some reason, snowflake SDK returns `fixed` type for DWH types like NUMBER(38, 15)
   // @see https://docs.snowflake.com/en/sql-reference/data-types-numeric for more info on types
   fixed: 'decimal',
-  timestamp_ntz: 'timestamp'
+  // All three TIMESTAMP variants map to the same generic timestamp. TIMESTAMP_LTZ and
+  // TIMESTAMP_TZ both store a UTC instant, and initConnection() pins the session to
+  // TIMEZONE = 'UTC', so every variant is rendered/hydrated as UTC. Leaving TZ/LTZ
+  // unmapped made BaseDriver.toGenericType() fall through to its `|| columnType`
+  // default and hand Cube Store the literal type name, which it rejects with
+  // "Custom type 'timestamp_tz' is not supported" when building a pre-aggregation.
+  // @see https://docs.snowflake.com/en/sql-reference/data-types-datetime
+  timestamp_ntz: 'timestamp',
+  timestamp_ltz: 'timestamp',
+  timestamp_tz: 'timestamp'
 };
 
 // User can create own stage to pass permission restrictions.
--- a/packages/cubejs-snowflake-driver/test/type-mapping.test.ts
+++ b/packages/cubejs-snowflake-driver/test/type-mapping.test.ts
@@ -1,0 +1,58 @@
+import { describe, expect, test } from 'vitest';
+
+import { SnowflakeDriver } from '../src';
+
+// Unit tests: no Snowflake connection is opened, so these run without credentials.
+// `toGenericType` is protected, hence the casts.
+function genericType(columnType: string): string {
+  const driver = new SnowflakeDriver({});
+  return (driver as any).toGenericType(columnType);
+}
+
+describe('SnowflakeDriver type mapping', () => {
+  // Every TIMESTAMP variant must reach Cube Store as `timestamp`. TIMESTAMP_TZ/LTZ used to
+  // fall through BaseDriver.toGenericType()'s `|| columnType` default and arrive as the
+  // literal type name, which Cube Store rejects when creating the pre-aggregation table
+  // ("Custom type 'timestamp_tz' is not supported").
+  test.each([
+    'TIMESTAMP_NTZ',
+    'TIMESTAMP_LTZ',
+    'TIMESTAMP_TZ',
+  ])('maps %s to timestamp', (columnType) => {
+    expect(genericType(columnType)).toEqual('timestamp');
+  });
+
+  // INFORMATION_SCHEMA.COLUMNS.DATA_TYPE reports upper case (`TIMESTAMP_TZ`) while the
+  // snowflake-sdk column type is lower case (`timestamp_tz`); both must resolve.
+  test.each([
+    'timestamp_ntz',
+    'timestamp_ltz',
+    'timestamp_tz',
+  ])('maps %s to timestamp regardless of case', (columnType) => {
+    expect(genericType(columnType)).toEqual('timestamp');
+  });
+
+  // Regression guard for the actual failure mode: a TIMESTAMP variant must never be
+  // returned as its raw Snowflake name. (Types like DATE/TEXT/BOOLEAN legitimately map to
+  // the same word, so this only applies to the timestamps.)
+  test.each([
+    'TIMESTAMP_NTZ',
+    'TIMESTAMP_LTZ',
+    'TIMESTAMP_TZ',
+  ])('does not pass %s through verbatim', (columnType) => {
+    expect(genericType(columnType).toLowerCase()).not.toEqual(columnType.toLowerCase());
+  });
+
+  // Guard the pre-existing mappings so the additions above cannot shadow them.
+  test.each([
+    ['NUMBER', 'decimal'],
+    ['fixed', 'decimal'],
+    ['object', 'HLL_SNOWFLAKE'],
+    ['DATE', 'date'],
+    ['TEXT', 'text'],
+    ['VARCHAR', 'text'],
+    ['BOOLEAN', 'boolean'],
+  ])('still maps %s to %s', (columnType, expected) => {
+    expect(genericType(columnType)).toEqual(expected);
+  });
+});
--- a/packages/cubejs-snowflake-driver/test/SnowflakeDriver.test.ts
+++ b/packages/cubejs-snowflake-driver/test/SnowflakeDriver.test.ts
@@ -12,12 +12,18 @@ const QUERY_TO_TEST_HYDRATION = `
     CAST(1265.88 AS NUMBER(10,2))                  AS "n",
     CAST('2026-04-28 13:07:42.123' AS TIMESTAMP_NTZ)         AS "ts_ntz",
     CAST('2026-04-28 13:07:42.123 +0000' AS TIMESTAMP_TZ)    AS "ts_tz",
+    -- A non-UTC offset must hydrate to the UTC instant, not the wall clock:
+    -- 13:07:42.123 -07:00 == 20:07:42.123 UTC.
+    CAST('2026-04-28 13:07:42.123 -0700' AS TIMESTAMP_TZ)    AS "ts_tz_offset",
+    CAST('2026-04-28 13:07:42.123 -0700' AS TIMESTAMP_LTZ)   AS "ts_ltz",
     CAST('2026-04-28' AS DATE)                               AS "d"
   UNION ALL
   SELECT
     CAST(0.10 AS NUMBER(10,2)),
     CAST('2000-02-29 00:00:00.007' AS TIMESTAMP_NTZ),
     CAST('2000-02-29 00:00:00.007 +0000' AS TIMESTAMP_TZ),
+    CAST('2000-02-29 00:00:00.007 -0700' AS TIMESTAMP_TZ),
+    CAST('2000-02-29 00:00:00.007 -0700' AS TIMESTAMP_LTZ),
     CAST('2000-02-29' AS DATE);
 `;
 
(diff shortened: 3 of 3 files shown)


packages/cubejs-snowflake-driver/test/type-mapping.test.ts
```diff
@@ -0,0 +1,58 @@
+import { describe, expect, test } from 'vitest';
```
ovr: @npwalker let's instead of using unit tests for such place (drop this file), add asserts in the integration tests. In `stream` test, there is `tableData` that stores `types` field. Thanks

packages/cubejs-snowflake-driver/src/SnowflakeDriver.ts
```diff
@@ -53,7 +53,16 @@ const SnowflakeToGenericType: Record<string, GenericDataBaseType> = {
   // For some reason, snowflake SDK returns `fixed` type for DWH types like NUMBER(38, 15)
   // @see https://docs.snowflake.com/en/sql-reference/data-types-numeric for more info on types
   fixed: 'decimal',
-  timestamp_ntz: 'timestamp'
+  // All three TIMESTAMP variants map to the same generic timestamp. TIMESTAMP_LTZ and
```
ovr: Let's drop all comments lines excluding `@see https://docs.snowflake.com/en/sql-reference/data-types-datetime`. Thanks

packages/cubejs-snowflake-driver/test/type-mapping.test.ts
npwalker: > let's instead of using unit tests for such place (drop this file), add asserts in the integration tests. In `stream` test, there is `tableData` that stores `types` field.

Done in 2669e0fce8. Dropped `test/type-mapping.test.ts`; the `stream` integration test now asserts `tableData.types`, so all three TIMESTAMP variants are checked to arrive as the generic `timestamp`.

packages/cubejs-snowflake-driver/src/SnowflakeDriver.ts
npwalker: > Let's drop all comments lines excluding `@see https://docs.snowflake.com/en/sql-reference/data-types-datetime`.

Done in 2669e0fce8. Only the `@see` line remains.

ovr:
Thank you @npwalker for your contribution! 👍 🍰

Issue #11407, opened 2026-07-29
MSSQL: out-of-range `datetime2` value makes `MIN()` return NULL silently


marciodfg:
**Describe the bug**

SQL Server's `datetime2` range starts at year 0001. Arrow's timestamp
representation (int64 nanoseconds) starts around 1677. A single row holding a
date in between — perfectly legal in the source — produces two different
failures depending on the query:

1. **A hard error** on a rollup:
   `Database Execution Error: The conversion of a datetime2 data type to a
   datetime data type resulted in an out-of-range value.`
2. **A silent wrong answer** on an aggregate: `SELECT MIN(d)` returns **`NULL`**,
   with no error and no warning, even though the column has non-null values in
   every row.

The second is the serious one. The query succeeds, returns the right *shape*,
and the value is simply wrong — so nothing downstream can detect it. Any
row-count or schema-based health check passes while the data is incorrect.

Such rows are easy to acquire accidentally: a bad ETL default, a `0001-01-01`
sentinel, or a mis-parsed date all land in range for `datetime2`.

**To Reproduce**

1. Point Cube at a **SQL Server** data source (`CUBEJS_DB_TYPE=mssql`) with the
   schema below. It contains one in-range row and one row dated `0202-01-01`.
2. Connect to the SQL API and run:

```sql
-- 1. Hard error: "The conversion of a datetime2 data type to a datetime data type
--    resulted in an out-of-range value."
SELECT DATE_TRUNC('year', d) y, COUNT(*) n FROM OutOfRange GROUP BY 1 ORDER BY 1;

-- 2. SILENT: returns lo = NULL, no error, despite both rows having a non-null d
SELECT MIN(d) lo FROM OutOfRange;
```

3. Statement 2 returns `NULL`. The expected value is `0202-01-01`, or failing
   that, an error.

**Expected behavior**

Either:

- the out-of-range value is surfaced as an error consistently (as it already is
  in statement 1), or
- it is handled/clamped and `MIN()` returns a correct value.

What should not happen is one query erroring and another silently returning
`NULL` for the same underlying data. Silent `NULL` is worse than a failure
because it cannot be detected downstream.

**Screenshots**

N/A.

**Minimally reproducible Cube Schema**

```javascript
cube(`OutOfRange`, {
  sql: `
  select 1 as id, CAST('0202-01-01' AS DATETIME2) as d
  UNION ALL
  select 2 as id, CAST('2024-01-01' AS DATETIME2) as d
  `,
[...]

PR #11408, opened 2026-07-29
fix(cubesql): Fail instead of returning NULL for out-of-range timestamps

marciodfg:
**Check List**

#11407

`transform_response` materialises a Cube response into Arrow. Time dimensions land in `Timestamp(Nanosecond)`, whose i64 nanosecond count only spans `1677-09-21` to `2262-04-11`. A source value outside that window — legal in every database that stores dates as `date` / `datetime2` / `timestamp` — was logged server-side and appended as NULL:

```rust
if let Some(nanos) = timestamp.and_utc().timestamp_nanos_opt() {
    builder.append_value(nanos)?;
} else {
    log::error!("Unable to cast timestamp value to nanoseconds: {}", timestamp);
    builder.append_null()?;
}
```

The client sees no error. `SELECT MIN(d) FROM t` comes back NULL over a column that holds no NULLs, the row count and the schema are both what the caller expects, and nothing downstream can tell the answer apart from a real one. That is the half of #11407 worth fixing — the other symptom in that issue already fails loudly.

A value that *fails to parse* one line above already aborts the query (`parse_date_str(...)?`). This makes a value that parses but cannot be represented behave the same way, and quotes the representable range so the error says what to do about it. `pg-srv` already rejects out-of-range timestamps on the wire with `Timestamp out of range: '{}'`, so the scan path is now consistent with it.

**Compatibility**

This turns a silent NULL into a hard error, and `9999-12-31` end-of-time sentinels in SCD-2 dimension tables are common enough that some users will notice. That is the intent — today they read NULL and have no way to tell — but if you would rather ship it behind a config flag, or hold it for the `Timestamp(Microsecond)` migration that the `// TODO switch parsing to microseconds` above this branch points at, say the word and I will rework it.

Worth flagging: `test_df_cube_scan_execute` asserted the old behaviour. Its fixture fed `9999-12-31 00:00:00.000` to the timestamp column and expected `None` back, while the `Date32` column in the same fixture round-tripped the same date fine. That fixture value now uses an in-range timestamp near the upper bound, and the out-of-range cases (below and above the range) get a dedicated test.

**Failure shape**

The reasoning above covers the batch path. On the native streaming path the same error is raised per chunk and thrown at `packages/cubejs-backend-native/src/stream.rs:275`, so with `CUBESQL_STREAM_MODE=true` a client can receive several chunks and *then* the error — a partial result followed by a failure. Still better than silently wrong data, but a distinct shape that the tests do not cover.

Any query *projecting* an affected dimension now fails, including a BI tool's opening `SELECT * … LIMIT 100`, and it fails after the source database has already done the work. `COUNT(*)` over the same cube is unaffected. Users are not cornered, which is the main reason I think the hard error is shippable: the same value still succeeds as `Date32` and as `Timestamp(Millisecond)` (`scan.rs:1114`), so `SELECT d::date` keeps working where `SELECT d` now errors, and clamping the sentinel in the data model is a second way out. Both are in the docs note.

**Verification**

```
cargo fmt --all -- --check
cargo clippy --locked --workspace --all-targets --keep-going -- -D warnings
cargo test -p cubesql --lib     # 770 passed, 0 failed
```

### DIFF
--- a/docs-mintlify/reference/core-data-apis/sql-api/index.mdx
+++ b/docs-mintlify/reference/core-data-apis/sql-api/index.mdx
@@ -233,6 +233,33 @@ Example:
 SET cube_cache = 'must-revalidate';
 ```
 
+## Limitations
+
+### Timestamp range
+
+The SQL API represents timestamps with nanosecond precision, so it can only return
+values between `1677-09-21 00:12:43.145224192` and `2262-04-11 23:47:16.854775807`.
+Selecting a time dimension whose value falls outside that range fails with
+`Timestamp out of range for column '<name>'`.
+
+Values outside the range are legal in most data sources and easy to acquire
+accidentally: a `9999-12-31` end-of-time sentinel in a slowly changing dimension table,
+a bad ETL default, or a mis-parsed date. To query such a column, either read it as a
+date, which has a far wider range:
+
+```sql
+SELECT MIN(valid_to::date) FROM customers;
+```
+
+or clamp the sentinel value in the [data model][ref-data-model-concepts].
+
+<Info>
+
+Earlier versions returned an out-of-range value as `NULL` rather than raising an error,
+which made the wrong answer indistinguishable from a genuine `NULL`.
+
+</Info>
+
 ## Troubleshooting
 
 ### `Can't find rewrite`
--- a/rust/cubesql/cubesql/src/compile/engine/df/scan.rs
+++ b/rust/cubesql/cubesql/src/compile/engine/df/scan.rs
@@ -432,6 +432,12 @@ impl ColumnarValueObject for LiteralRowsValueObject {
     }
 }
 
+/// Bounds of `DataType::Timestamp(TimeUnit::Nanosecond, _)`: an i64 count of
+/// nanoseconds since the epoch. Quoted back to the user so the error says what
+/// range a value has to fall into.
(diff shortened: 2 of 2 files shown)


ovr:
@marciodfg, thank you for submitting the PR, i've triggered our review bot. Could you review the comments from it? Thanks

marciodfg:
Thanks for running the bot — all three comments are addressed in `0895a8f`:

- **Column name and value** — the error now reads ``Timestamp out of range for column 'valid_to': '9999-12-31 00:00:00.000'. …`` and closes with the "read it as a date" hint.
- **Classification** — switched to `post_processing`. Cosmetic today since `load_data` overwrites the cause at `scan.rs:824`, but `22000` is the honest one for a value that can't be represented.
- **Tests** — the assertions now pin the column name and the offending value, and cover one nanosecond outside each end; a new test pins the inclusive bounds to `i64::MIN` / `i64::MAX`.

Also added the docs note (SQL API → Limitations → Timestamp range) and updated the description with the streaming-path failure shape.

On the red checks: I reproduced `tests (postgres, false)` locally against an image built from this branch — 119 passed, 0 failed, including every `SQL API:` case, which are the ones that exercise this code. Worth knowing that every *green* cloud driver cell was green because it was **skipped** — fork PRs get no secrets — so only 16 cells actually ran and 13 of those failed, each using all three retry attempts. Could you re-run?

PR #11410, opened 2026-07-29
fix(cubestore): keep the projection schema when rewriting a rolling window

waralexrom:
## Summary

A rolling-window query whose consumer does not select one of the GROUP BY columns failed to plan in CubeStore with `Optimizer rule 'optimize_projections' failed / Schema error: No field named ...`. The error points at DataFusion, but the defect is ours: `RollingOptimizerRule` replaces a projection with a node that has a different schema.

This blocks a customer model where a `type: switch` (calc group) dimension dispatches rolling-window measures through `case` entrypoints served from rollups — the calc-group column is exactly the unselected GROUP BY column, so every such query fails in a rollup-only deployment.

## Changes

- **Do not widen the schema of the replaced node.** `RollingWindowAggregate` always emits `dimension + partition_by + rolling aggregates`, and `partition_by` comes from the aggregate's GROUP BY — so a column the projection dropped reappeared in the output. Ancestors keep resolving columns by position against their cached schema, hence the failure. The rewrite now restores the original output with a projection on top, and declines when it cannot be expressed from the rolling node.
- **Make partition keys nullable.** Planning these queries reached a second, independent defect: dimension values missing from the input still produce a row with null partition keys, but the keys inherited nullability from the input columns. A calc-group dimension stored as a literal is not nullable, so any window with an empty bucket died at execution with `declared as non-nullable but contains null values`.
- **Decline instead of building a node the rule cannot describe.** A short alias list silently truncated the schema, and `distinct` / `filter` / `order_by` / `null_treatment` were dropped from the aggregate although the executor honours none of them. Both now leave the plain aggregate/range-join in place. The destructure is exhaustive over `AggregateFunctionParams`, so a future DataFusion upgrade that adds a modifier fails to compile here rather than silently dropping it.
- **Graceful degradation** where the old code aborted planning: a `schema_from` failure now declines the rewrite instead of failing the query.

## Testing

New `rolling_window_unused_partition_by` in `cubestore-sql-tests` — four cases over a 7-point series so one bucket is genuinely empty: the partition column pruned by the consumer, omitted by the projection outright, several real partitions selected, and several partitions dropped. Expected values were checked against the window model rather than recorded from the implementation. It runs in all three configurations (in-process, multi-process, 2-node cluster), so the worker-side schema reconstruction is covered too.

- Each fix is independently pinned: reverting only `rolling_optimizer.rs` reproduces the planning failure, reverting only `rolling.rs` reproduces the nullability failure.
- Full SQL suite 185/185; `cargo fmt` clean; no clippy findings on the changed files.
- Mutation check: disabling the rule entirely fails 10 of 14 rolling tests, so a silently lost rewrite is caught by the existing result assertions.
- Traced the rule on the Tesseract switch/calc-group end-to-end tests: 13 matches → 13 rewrites, 0 bail-outs, 7 of them through the new path on the calc-group column. Those tests run each query twice — from a CubeStore rollup and from source Postgres with pre-aggregations off — and the results agree cell for cell.

Known, unrelated and out of scope: aggregate `FILTER` is broken in CubeStore generally (`SELECT sum(n) FILTER (WHERE ...) FROM t` fails on column resolution), and the multi-rolling-aggregate executor path has a pre-existing group-index bug not reachable from Cube-generated SQL.

### DIFF
--- a/rust/cubestore/cubestore-sql-tests/src/tests.rs
+++ b/rust/cubestore/cubestore-sql-tests/src/tests.rs
@@ -5770,6 +5770,37 @@ LIMIT
     Ok(())
 }
 
+/// Fails unless the physical plan for `query` mentions `expected`. Result assertions alone cannot
+/// tell an optimized plan from the fallback it degrades to, so shapes whose whole point is that a
+/// rewrite fires need to say so explicitly.
+async fn assert_plan_contains(
+    service: &Box<dyn SqlClient>,
+    query: &str,
+    expected: &str,
+) -> Result<(), CubeError> {
+    let res = service
+        .exec_query(&format!("EXPLAIN ANALYZE {}", query))
+        .await?;
+    let plans = res
+        .get_rows()
+        .iter()
+        .flat_map(|r| r.values().iter())
+        .filter_map(|v| match v {
+            TableValue::String(s) => Some(s.as_str()),
+            _ => None,
+        })
+        .collect::<Vec<_>>();
+    if !plans.iter().any(|p| p.contains(expected)) {
+        return Err(CubeError::internal(format!(
+            "`{}` not found in the plan for {}:\n{}",
+            expected,
+            query,
+            plans.join("\n")
+        )));
+    }
+    Ok(())
+}
+
 async fn rolling_window_unused_partition_by(service: Box<dyn SqlClient>) -> Result<(), CubeError> {
     service.exec_query("CREATE SCHEMA s").await?;
     service
@@ -5819,10 +5850,11 @@ ORDER BY 1 ASC, 2 ASC"
     ]);
 
     // A calc-group style constant: the source column is not nullable, but the empty bucket at day 7
-    // still has to report a null key for it.
-    let r = service
-        .exec_query(&query("'R12'", "`b`.`win` `win`,", "`q`.`day`, `q`.`num`"))
-        .await?;
+    // still has to report a null key for it. The rewrite has to restore the projection's output on
+    // top of the rolling node here, since `win` is grouped by but not selected.
+    let pruned = query("'R12'", "`b`.`win` `win`,", "`q`.`day`, `q`.`num`");
+    assert_plan_contains(&service, &pruned, "RollingWindowAgg").await?;
+    let r = service.exec_query(&pruned).await?;
     assert_eq!(to_rows(&r), without_partition);
 
     // Same, but the projection omits the partition-by column outright instead of leaving it to
@@ -5834,13 +5866,13 @@ ORDER BY 1 ASC, 2 ASC"
 
     // Several partitions, all of them asked for: the rewrite outputs exactly what the projection
     // did, so nothing has to be restored on top of it.
-    let r = service
-        .exec_query(&query(
-            "name",
-            "`b`.`win` `win`,",
-            "`q`.`day`, `q`.`win`, `q`.`num`",
-        ))
-        .await?;
+    let selected = query(
+        "name",
+        "`b`.`win` `win`,",
+        "`q`.`day`, `q`.`win`, `q`.`num`",
+    );
+    assert_plan_contains(&service, &selected, "RollingWindowAgg").await?;
+    let r = service.exec_query(&selected).await?;
     assert_eq!(
         to_rows(&r),
         rows(&[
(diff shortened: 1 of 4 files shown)


waralexrom:
Thanks — addressed 1, 3 and 4. Note 2 I checked and am pushing back on part of it; details below.

**Note 1 — fixed.** Aliases are now resolved from the aggregate they belong to (matching the projection column against the aggregate's output field name) rather than paired by position. That also removes the `len()` heuristic: an aggregate the projection does not read under its own name has no name to output, so the lookup misses and the rewrite declines. Two independent reviews flagged this and neither could construct a reachable case, but the invariant is real and this block now owns it.

**Note 4 — fixed, and it was the right call.** Added `assert_plan_contains` (following `assert_limit_pushdown_using_search_strings`) asserting `RollingWindowAgg` appears in the plan, for both branches: the shape whose output is restored by a projection, and the shape whose output already matches. Verified the assertion bites — with the rule disabled the test now fails on the plan, not just on values.

**Note 3 — fixed.** `debug_assert_eq!` on `fields.len()` pinning the dimension → partition keys → aggregates layout.

**Note 2 — the mechanism is real, the "fast path may be close to dead" part is not.** `(Some(_), None) => false` in `index_of_column_by_name` is confirmed, so a qualified probe against an unqualified field is a hard miss. But I traced the rule over the Tesseract switch/calc-group end-to-end suite, and the fast path fires **6 of 13** matches — it is alive and common. The reason is that DataFusion elides `AS win` when the alias equals the column name, leaving a bare `Expr::Column(b.win)` whose field keeps the qualifier, so both sides are qualified and compare equal.

Same trace, before and after this follow-up: **13 matches → 7 restored-projection rewrites + 6 direct, 0 bail-outs**, with aliases resolving to `sales__r3_amount` / `sales__r12_amount`. Produced rows are byte-identical to the previous run, so the name-based pairing changed no behaviour.

I did not add the unqualified fallback probe. If the probe misses, `Projection::try_new` on that same qualified column would fail too, so the decline is preventing a planning error rather than choosing a slower plan — and the fallback would add a path with no reachable test (0 declines in 13 matches). Happy to revisit if someone produces a shape that reaches it.

All three configurations green (in-process, multi-process, 2-node cluster), full suite 185/185, fmt clean, no clippy findings on the changed files.

ovr (approved):

waralexrom:
Note 1 is correct, and chasing it turned up something bigger.

**The gap is real.** Confirmed by mutation: reverting `4578583`'s name-based pairing back to a positional `zip` leaves all five plan-level cases green, exactly as you said. Every case has one aggregate, so the two schemes cannot disagree.

**Writing the two-aggregate case first meant asking whether such a plan is sound. It is not.** The executor advances one `group_index` per (aggregate, dimension point) at `rolling.rs:905`, while each aggregate accumulates into its own group space, so with two aggregates they interleave and everything past the first reads the slots the others left empty:

```sql
SELECT q.day, q.sn, q.sm FROM (
  SELECT s0.date_from day, sum(b.n) sn, sum(b.m) sm
  FROM (series 1..5) s0
  LEFT JOIN (SELECT day d, n, m FROM s.Data) b
    ON b.d > s0.date_to - 1 AND b.d <= s0.date_to
  GROUP BY 1) q ORDER BY 1
```
```
 day | sn | sm        expected 100 / 100 / 300 / 300 / 500
   1 | 10 | NULL
   2 | 10 | NULL
   3 |  3 | NULL      <- silently null, no error
```

This is ordinary SQL, not the hand-written exotica your earlier B5 note assumed — so I did not add a test pinning that two-aggregate plans get rewritten. Instead the rule now declines them, and the range join answers correctly and consistently (`NULL/NULL, 3/300, NULL/NULL, 5/500, NULL/NULL` — the literal semantics of the predicate as written). Cube gives each rolling measure its own subquery, and my trace over the Tesseract suite showed `aggs=1` for all 13 matches, so nothing reachable through the semantic layer loses an optimization. Repairing the executor's group accounting is the better fix and belongs in its own change.

That also settles note 1 rather than patching it: with at most one rolling aggregate, positional and name-based pairing cannot disagree, so there is no behaviour left to pin. I kept the name-based version — it is the more precise statement — but I am no longer claiming a test distinguishes them.

**One correction to the review.** You wrote that the distinct/filter cases bite the explicit modifier guard because `agg_col` derives the name the aggregate node actually emits. The lookup resolves against the *rebuilt* aggregate, which has its modifiers stripped and therefore a different `schema_name()`, so it misses. Verified by mutation: removing the modifier guard leaves both cases declining. Two mechanisms reach the same decline, which is what the commit message claimed — the guard still earns its place, since it is the only one left if the rebuild ever preserves modifiers.

**Note 2** — reworded to state the rationale rather than the mechanics.

Added: `declines_more_than_one_rolling_aggregate` (plan level) and `rolling_window_two_aggregates` (SQL, asserts the plan omits `RollingWindowAgg` and pins the values). Both verified to fail with the guard removed. All suites green: unit 81, in-process 186, multi-process 186, cluster 185, migration 132.

PR #11434, opened 2026-07-31
fix(cubestore): advertise INCR, not INC, in the CACHE unknown-command error

igorlukanin:
## Issue

The `CACHE` command parser accepts `incr`, but the error it emits for an unknown subcommand advertises `INC` — a keyword that doesn't parse. An operator who follows the error message gets a second error:

```
$ CACHE FOO bar;
Unknown cache command: foo, available: SET|GET|KEYS|INC|REMOVE|CLEAR

$ CACHE INC bar;
Unknown cache command: inc, available: SET|GET|KEYS|INC|REMOVE|CLEAR
```

The accepted set is exactly `SET|GET|KEYS|INCR|REMOVE|CLEAR`, so `INC` is the only inaccurate entry. This is the only enumerated-keyword error message in the parser — the other unknown-command arms emit a bare "Unknown … command" with no list.

## Description of Change

Corrects the advertised keyword to `INCR`.

### DIFF
--- a/rust/cubestore/cubestore/src/sql/parser.rs
+++ b/rust/cubestore/cubestore/src/sql/parser.rs
@@ -508,7 +508,7 @@ impl<'a> CubeStoreParser<'a> {
             "clear" => CacheCommand::Clear {},
             other => {
                 return Err(ParserError::ParserError(format!(
-                    "Unknown cache command: {}, available: SET|GET|KEYS|INC|REMOVE|CLEAR",
+                    "Unknown cache command: {}, available: SET|GET|KEYS|INCR|REMOVE|CLEAR",
                     other
                 )))
             }


rust/cubestore/cubestore/src/sql/parser.rs
```diff
@@ -1292,6 +1292,53 @@ mod tests {
         Ok(())
     }
 
+    #[test]
```
igorlukanin: Don't add a new test, the fix above is enough.

igorlukanin: Dropped the test — the PR is the one-line message fix now.

rust/cubestore/cubestore/src/sql/parser.rs
```diff
@@ -1292,6 +1292,53 @@ mod tests {
...
+                    .to_string()
+            })
+            .filter(|k| !k.is_empty())
+            .collect::<Vec<_>>();
+        assert_eq!(
+            advertised,
+            vec!["SET", "GET", "KEYS", "INCR", "REMOVE", "CLEAR"],
+            "unexpected advertised command list: {}",
+            err
+        );
```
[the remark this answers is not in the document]
igorlukanin: Moot now — the test is gone, the change is just the message.

PR #11435, opened 2026-07-31
docs: remove table chart Table options section for a tab and controls that do not exist

igorlukanin:
## Summary

The table chart page documented a **Table** tab that the chart doesn't have, listing four options under it. The chart's tabs are **Fields**, **Style**, and **Formatting**.

Of the four options in that table:

- **View names** and **Group dimensions** don't exist — no spec field, no control, no string for either anywhere in the product. Dimension grouping is unbuilt work, so that row described a feature we don't ship.
- **Header text** and **Row numbers** do exist, but as **Overflow** and a **Row numbers** toggle in the Style tab's Headers and Values sections — both already documented correctly further down the page, under *Style options*.

So the section was either wrong or duplicated, and the two correct rows named the controls by labels the UI doesn't use. Removed it. The `Column width` pointer it ended on is redundant with the **Width** row in *Column field options* immediately below, which already links to the same section.

Same class of fix as #11320, one section further up the page.

## Test plan

- Verified against the shipped chart: `TableChartSpecSchema` has no view-name or dimension-grouping field; the builder renders exactly three tabs; exhaustive search for `showViewNames` / `viewNames` / `groupDimensions` and the literal labels returns nothing for the table chart.
- Confirmed **Row numbers** and **Overflow** are documented under *Style options*, so no content is lost.
- No page linked to the removed `#table-options` anchor.
- `mint broken-links` clean; the page renders without MDX errors on a local Mintlify server.

### DIFF
--- a/docs-mintlify/docs/explore-analyze/charts/chart-types/table.mdx
+++ b/docs-mintlify/docs/explore-analyze/charts/chart-types/table.mdx
@@ -26,19 +26,6 @@ Pivot behavior:
 - Pivot columns can also be sorted by row values — click a row number to sort by that row, including the totals row.
 - Pivoted columns can be hidden, but hiding is indexed to the specific value (e.g. hiding `status: returned`), not position.
 
-## Table options
-
-The **Table** tab in the configuration panel controls layout and display settings:
-
-| Option | Description |
-|---|---|
-| **Header text** | Controls whether long column headers **truncate** or **wrap** to the next line |
-| **View names** | When enabled, the view/cube name is shown in the column header |
-| **Row numbers** | Adds a row number column on the left |
-| **Group dimensions** | Groups multiple dimensions per row into a collapsible section for subtotaling |
-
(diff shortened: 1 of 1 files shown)


igorlukanin:
Fixed both tab-name references in `03ba7de`: the totals line now points at the **Totals** section of the **Style** tab, and line 253 says **Formatting** rather than "Conditional formatting" — same wrong-tab defect, so it belongs here.

The **Display** tab flag is a bigger one than "probably fine", and it needs a scoping call rather than a rename. There is no Display tab: "Display as" is a dropdown inside each column's section under **Style → Columns**, it only appears on numeric columns, and its three values are Value / Inline bars / Sparkline. So the **Link** option that section tells you to pick doesn't exist — links are a right-click cell-menu action driven by `links:` on a dimension in the data model, not a rendering mode. Image rendering doesn't exist at all; `imageUrl` is a member format with no consumer in the table.

Rewriting that section means deciding how to present links given where they actually come from, which is why it's not in this deletion.

PR #11440, opened 2026-07-31
docs: add GPT-5.6 Sol and Terra to the predefined agent LLM models

igorlukanin:
Adds GPT-5.6 Sol and Terra to the list of predefined models for an agent's `llm` property.


igorlukanin:
Checked the cloud-side enum: both keys match exactly. `gpt_5_6_sol` and `gpt_5_6_terra` are the strings added to `PredefinedLLMProviders` in cubedevinc/cube-runtime#1112 and to the `predefined_llm` enum in cubedevinc/cubejs-enterprise#13569, so no typo to catch. This PR lands last of the three for that reason.

PR #11463, opened 2026-08-03
docs: correct the dashboard time granularity claim about custom granularities

igorlukanin:
The dashboard **Time Granularity** control docs claim that custom granularities defined in the data model show up in the switcher alongside the built-in ones. They don't — the control offers the built-in granularities only.

The grain list the widget renders is a fixed set of the standard units, and custom granularities never reach that surface in the first place: the semantic-view member type the dashboard reads has no `granularities` field, so there is nothing for the picker to merge in. Saving clamps to the same fixed set as well.

This corrects the line to describe what the control actually does, leaving "yet" in place since the capability is planned.


igorlukanin:
Keeping "in this list" — that's the wording that was signed off verbatim, and "the list" is what the section calls it throughout.

PR #11483, opened 2026-08-05
fix(tesseract): keep the query row set when a multi-stage filter directive drops a filter

waralexrom:
## Summary

A multi-stage measure whose `filter:` directive drops a query filter reported dimension values the query had filtered out. Those values are the keys the `FullKeyAggregate` joins on, so they came back as result rows carrying NULL for every other measure — the reported symptom being a share-of-total denominator returning the whole universe with a share of 0.

The directive is meant to free the *inner aggregation stage* from a filter, which is what makes a "share of total ignoring this filter" denominator possible. It must not free the rows the measure *reports*. Only the JOIN-model assembly can hold those two apart — its keys side enumerates the query grid while its measure side spans the widened set — so a member that drops a query filter is now handed to it.

## Changes

- `query_filters_dropped` detects a directive dropping a filter the query itself restricts the grid by. Membership is checked against the root query state, so a filter an upstream member introduced through its own `filter: include` does not count — that narrows the upstream member's view, not the result grid. The comparison is member-aware, since `FilterItem` equality compares a filter's operator and values but not the member it targets, and a dropped filter would otherwise be cancelled out by an unrelated look-alike that survived.
- When it fires, `use_window_path` is revoked (a window expression has one row set serving both the rows it reports and the rows it aggregates over) and the keys side is built, alongside the pre-existing grain-shrink trigger.
- Three shapes are exempt: a dimensionless parent grid is a single row no filter change can widen; a `Dimension` inode never reads `keys_input`; and a `Rank` inode ranks within whatever its source carries, so a keys side would shrink the ranked population and collapse the ranks — trading a widened row set for wrong values.
- `QueryProperties::eq_as_state`, the multi-stage CTE dedup key, is member-aware too. It compared filter vectors with the same member-blind equality, which let the keys side dedup onto the very CTE it exists to bound: a directive combining `exclude` with an `include` whose operator and values coincide on another member plans the widened measure child first, and the keys child then matched it and came back as the same CTE.
- The explicit-keys join is now null-safe, consistent with the full-join and inner-join strategies which already were. A plain `=` never matches NULL to NULL, so a key row whose dimension was NULL kept its place in the grid and lost the measure. It cannot be narrowed to the dimensions that need it: the data model carries no nullability information, and a dimension read through an outer join in the join graph arrives as NULL even from a column declared NOT NULL.
- Member-aware filter-tree equality lives in `planner/filter/tree_ops`, beside the other tree walkers, and is used by both call sites.

## Testing

Reproduced against the shape from the customer model — a grand-total denominator whose `grain.exclude` and `filter.exclude` both name a dimension the query groups by and filters on, consumed by a ratio with `coalesce`. The generated SQL matched the reported production plan: 9 rows with 6 extra reading 0, now 3 rows with correct shares.

Nine new Rust integration tests, all value-based:

- the share-of-grand-total shape, plus an assertion that no window expression is emitted
- `grain.exclude` + `filter.exclude` on a grouped and filtered dimension
- a dropped filter whose dimension the query does *not* group by, grouped by the primary key so the widening is visible: 15 rows before, 6 after
- a directive whose `include` predicate matches the dropped filter on operator and values while naming another member — the dedup shape above: 9 rows before, 6 after
- a NULL dimension key keeping its measure
- the value side of a `keep_only` segment drop, at a grain where the dropped filter changes the number (600 against 1400)
- a rolling window over a dropped date filter, at two grains. A rolling window rewrites the date range it hands to its base state, so the dropped filter is not detected there; these pin that the row set stays bounded anyway, because the window takes its rows from the time series and its values through the frame condition
- a rank measure with a filter directive, pinning the row set `Rank` keeps reporting and failing loudly if `Rank` were dropped from the exemption

Two pre-existing snapshots changed: both filtered `category = books` yet returned `clothing` and `electronics` rows with a NULL measure.

One JS Postgres integration test covers the same combination end-to-end through the schema compiler.

`cargo test --features integration-postgres -p cubesqlplanner`: 1124 passed. 16 JS Postgres suites green, including `sql-generation` (134), `pre-aggregations` (47) and `calc-groups` (24) — the dedup-key change touches every multi-stage plan, so the pre-aggregation and sql-generation suites are the relevant ones.

### Known limitations

- A rollup that does not store the dropped filter's dimension can no longer serve such a query. Before, the filter was absent from the plan entirely, so the rollup answered a query the user had not made; a rollup that does carry the dimension still matches.
- `Rank` measures keep reporting the wider row set, for the reason above. Pinned by a test rather than left implicit.
- The null-safe keys join costs hash- and merge-joinability on engines that treat the operator as unhashable (PostgreSQL; Redshift has no such operator and falls back to an equally unhashable OR-form). Behaviour on Snowflake was not measured.
- Detection compares whole filters, so a path that rewrites filter values between the root state and a CTE state reads as one the query never asked for. The rolling-window date-range rewrite is the one such path today and bounds its rows by other means; the constraint is recorded at the predicate.

### DIFF
--- a/packages/cubejs-schema-compiler/test/integration/postgres/multi-stage-grain.test.ts
+++ b/packages/cubejs-schema-compiler/test/integration/postgres/multi-stage-grain.test.ts
@@ -133,6 +133,22 @@ cubes:
             - orders.status
           include:
             - orders.id
+
+      # ── grain.exclude + filter.exclude on the same dim ────────────
+      # "Share of the whole universe" denominator: the value ignores both
+      # the query filter on status and the status partition. The result row
+      # set must stay the query's — dropping the filter may not resurrect
+      # rows the query filtered out.
+      - name: amount_grain_and_filter_exclude_status
+        sql: "{CUBE.total_amount}"
+        type: sum
+        multi_stage: true
+        grain:
+          exclude:
+            - orders.status
+        filter:
+          exclude:
+            - orders.status
     `);
 
   if (getEnv('nativeSqlPlanner')) {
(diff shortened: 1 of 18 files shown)


paveltiunov (approved):

rodrigobercini:
Thanks for implementing this fix. We can now drop the workaround we had in place, which has made our requests much cleaner. Really appreciate the quick turnaround on this.

PR #11496, opened 2026-08-06
fix(cubesql): preserve literal ORDER BY aliases

jwils:
**Check List**

CubeSQL sort pushdown can replace an aliased integer literal with the literal itself. Databases that interpret integers in ORDER BY as positional ordinals can then reject the generated SQL when the ordinal exceeds the select-list width.

- Restore the generated select-list alias when an ORDER BY expression matches a projected or grouped literal, after stripping aliases from both expressions
- Preserve sort direction, null ordering, ordinary column sorts, and existing window-expression alias handling
- Cover grouped literals with two sort keys, a projected literal without an outer GROUP BY, and literal sorting through Cube load requests

**Risk Assessment**

Low. The rewrite applies only when generating SQL text, not when building Cube load requests. It is limited to ORDER BY expressions that exactly match literal expressions in the same select after stripping aliases; all other sort expressions retain their existing behavior.

**Validation**

- cargo test -p cubesql --lib: 778 passed, 6 ignored
- cargo clippy -p cubesql --lib --tests
- cargo fmt --all -- --check

---
**Update Sep 6:** Cover projected literals and alias-wrapped expressions, and collect literal candidates once per select. Add the projected-literal regression test and tighten the grouped test’s alias matching and failure messages.

---
**Update Sep 8:** Skip literal alias rewriting on the push-to-Cube path, where generated aliases cannot resolve against scan members. Add a regression test for a projected literal with `taxful_total_price` and `LIMIT 100`; it verifies the load request keeps `hour_slot ASC` and the limit. The test fails before the guard and passes afterward.

### DIFF
--- a/rust/cubesql/cubesql/src/compile/engine/df/wrapper.rs
+++ b/rust/cubesql/cubesql/src/compile/engine/df/wrapper.rs
@@ -1759,11 +1759,44 @@ impl WrappedSelectNode {
         )
         .await?;
 
+        // Sort pushdown can replace a grouped literal's alias with the literal expression.
+        // Integer literals in ORDER BY are interpreted as select-list positions by some SQL
+        // dialects, so restore the generated alias when the literal is selected in this query.
+        let order_expr = self
+            .order_expr
+            .iter()
+            .map(|order_expr| match order_expr {
+                Expr::Sort {
+                    expr,
+                    asc,
+                    nulls_first,
+                } => {
+                    let literal_alias = flat_group_expr.iter().zip(group_by.iter()).find_map(
+                        |(selected_expr, (aliased_column, _))| {
+                            (matches!(selected_expr, Expr::Literal(_))
+                                && selected_expr == expr.as_ref())
+                            .then(|| aliased_column.alias.clone())
+                        },
+                    );
+
+                    match literal_alias {
+                        Some(alias) => Expr::Sort {
+                            expr: Box::new(Expr::Column(Column::from_name(alias))),
+                            asc: *asc,
+                            nulls_first: *nulls_first,
+                        },
+                        None => order_expr.clone(),
+                    }
+                }
+                _ => order_expr.clone(),
+            })
+            .collect::<Vec<_>>();
+
         // Sort expressions can reference window expressions computed in this same select
         // by their full DataFusion name. Those columns don't exist in the source SQL, so
         // rewrite them to the generated window aliases, which ORDER BY can reference by name.
         let order_expr = if window.is_empty() {
-            self.order_expr.clone()
+            order_expr
         } else {
             let window_columns = self
                 .window_expr
@@ -1777,9 +1810,9 @@ impl WrappedSelectNode {
                 })
                 .collect::<result::Result<HashMap<_, _>, CubeError>>()?;
             let window_columns_ref = window_columns.iter().collect();
-            self.order_expr
-                .iter()
-                .map(|e| replace_col(e.clone(), &window_columns_ref))
+            order_expr
+                .into_iter()
+                .map(|e| replace_col(e, &window_columns_ref))
                 .collect::<Result<Vec<_>>>()?
         };
 
--- a/rust/cubesql/cubesql/src/compile/engine/df/wrapper.rs
+++ b/rust/cubesql/cubesql/src/compile/engine/df/wrapper.rs
@@ -2048,8 +2048,8 @@ impl WrappedSelectNode {
             }
             expr
         }
-        // Push-to-Cube builds its order separately from self.order_expr. Generated SQL
-        // aliases are not scan members and cannot be resolved on that path.
+        // Push-to-Cube discards this `order` and builds its own from `self.order_expr`, so
+        // there is nothing to fix there, and a generated alias is not a scan member.
         let literal_aliases = if push_to_cube_context.is_some() {
             vec![]
         } else {
--- a/rust/cubesql/cubesql/src/compile/engine/df/wrapper.rs
+++ b/rust/cubesql/cubesql/src/compile/engine/df/wrapper.rs
@@ -2048,16 +2048,21 @@ impl WrappedSelectNode {
             }
             expr
         }
-        let literal_aliases = self
-            .projection_expr
-            .iter()
-            .zip(projection.iter())
-            .chain(flat_group_expr.iter().zip(group_by.iter()))
-            .filter_map(|(selected_expr, (aliased_column, _))| {
-                let expr = unalias(selected_expr);
-                matches!(expr, Expr::Literal(_)).then(|| (expr, &aliased_column.alias))
-            })
-            .collect::<Vec<_>>();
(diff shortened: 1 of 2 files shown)


rust/cubesql/cubesql/src/compile/engine/df/wrapper.rs
```diff
@@ -1759,11 +1759,44 @@ impl WrappedSelectNode {
...
+                            expr: Box::new(Expr::Column(Column::from_name(alias))),
+                            asc: *asc,
+                            nulls_first: *nulls_first,
+                        },
+                        None => order_expr.clone(),
+                    }
+                }
+                _ => order_expr.clone(),
+            })
+            .collect::<Vec<_>>();
```
MazterQyou, line 1793: Thanks for tracking this down, the diagnosis is right and the shape of the fix is the one I'd pick too. Two gaps before it lands, both of which the added test can't see:

**Only grouped literals are matched.** The `find_map` walks `flat_group_expr`/`group_by`, so a literal that is projected but not grouped never matches and still renders positionally:

```sql
WITH with_rate AS (
  SELECT customer_gender, SUM(taxful_total_price) AS hourly_rate
  FROM KibanaSampleDataEcommerce
  GROUP BY 1
)
SELECT customer_gender, 19 AS hour_slot, hourly_rate
FROM with_rate
ORDER BY hour_slot ASC, customer_gender ASC
LIMIT 100;
```

With this PR applied that still generates `ORDER BY 19 ASC, "with_rate"."customer_gender" ASC`. `SqlTemplates::select` takes projection, group_by, aggregate and window as separate lists and concatenates them into the select list, so a projection alias is just as legal an `ORDER BY` reference as a group alias; chaining `self.projection_expr.iter().zip(projection.iter())` in front covers it.

**Both sides can arrive alias-wrapped**, so `selected_expr == expr.as_ref()` under-matches. For the query above:

```
order=[Int64(19) AS hour_slot ASC NULLS LAST, #t.customer_gender ASC NULLS LAST]
proj =[#t.customer_gender, Int64(19) AS hour_slot, #t.x]
```

The raw equality only holds when both sides are bare `Expr::Literal`, which is true for group exprs and nothing else; that's exactly why the grouped test passes and the projected one doesn't. Stripping `Expr::Alias` on both sides fixes it.

While in here, hoisting the candidate set out of the per-sort-expr loop keeps the common case (nothing literal in the select list) free of per-expr work and flattens the nesting a little.

```suggestion
        // Sort push down replaces a select list alias with the expression behind it. For a
        // literal that renders a bare constant in ORDER BY, and many dialects read a bare
        // integer there as a select list position rather than a sort key. Restore the
        // generated alias when the very same literal is selected by this select.
        fn unalias(expr: &Expr) -> &Expr {
            match expr {
                Expr::Alias(expr, _) => expr.as_ref(),
                expr => expr,
            }
        }
        let literal_aliases = self
            .projection_expr
            .iter()
            .zip(projection.iter())
            .chain(flat_group_expr.iter().zip(group_by.iter()))
            .filter_map(|(selected_expr, (aliased_column, _))| {
                let expr = unalias(selected_expr);
                matches!(expr, Expr::Literal(_)).then(|| (expr, &aliased_column.alias))
            })
            .collect::<Vec<_>>();
        let order_expr = if literal_aliases.is_empty() {
            self.order_expr.clone()
        } else {
            self.order_expr
                .iter()
                .map(|order_expr| {
                    let Expr::Sort {
                        expr,
                        asc,
                        nulls_first,
                    } = order_expr
                    else {
                        return order_expr.clone();
                    };
                    let Some((_, alias)) = literal_aliases
                        .iter()
                        .find(|(literal, _)| *literal == unalias(expr))
                    else {
                        return order_expr.clone();
                    };
                    Expr::Sort {
                        expr: Box::new(Expr::Column(Column::from_name(*alias))),
                        asc: *asc,
                        nulls_first: *nulls_first,
                    }
                })
                .collect()
        };
```

The rest of the hunk is good as is, handing `order_expr` into the window-alias rewrite and the `into_iter()` / `replace_col(e, ..)` changes are both correct.

rust/cubesql/cubesql/src/compile/mod.rs
```diff
@@ -17576,6 +17576,78 @@ LIMIT {{ limit }}{% endif %}"#.to_string(),
...
+            sql
+        );
+
+        let physical_plan = query_plan.as_physical_plan().await.unwrap();
+        println!(
+            "Physical plan: {}",
+            displayable(physical_plan.as_ref()).indent()
+        );
+    }
+
```
MazterQyou, line 17650: Could you please add a companion test for the no-`GROUP BY` case? It's the half of the bug the current test doesn't cover, and it fails on this branch today:

```rust
    #[tokio::test]
    async fn test_sort_by_projected_literal_alias_sql_push_down() {
        if !Rewriter::sql_push_down_enabled() {
            return;
        }
        init_testing_logger();

        let query_plan = convert_select_to_query_plan(
            r#"
            WITH with_rate AS (
                SELECT
                    customer_gender,
                    SUM(taxful_total_price) AS hourly_rate
                FROM KibanaSampleDataEcommerce
                GROUP BY 1
            )
            SELECT customer_gender, 19 AS hour_slot, hourly_rate
            FROM with_rate
            ORDER BY hour_slot ASC, customer_gender ASC
            LIMIT 100
            "#
            .to_string(),
            DatabaseProtocol::PostgreSQL,
        )
        .await;

        let logical_plan = query_plan.as_logical_plan();
        let sql = logical_plan.find_cube_scan_wrapped_sql().wrapped_sql.sql;
        assert!(
            sql.contains(r#"ORDER BY "hour_slot" ASC, "with_rate"."customer_gender" ASC"#),
            "literal sort expression must reference the projected alias: {sql}"
        );

        let physical_plan = query_plan.as_physical_plan().await.unwrap();
        println!(
            "Physical plan: {}",
            displayable(physical_plan.as_ref()).indent()
        );
    }
```

MazterQyou (changes requested):
Confirmed the bug and the fix locally against the repro from #11588.

The approach is right: restoring the generated alias in `order_expr` before the window-alias rewrite is the correct spot, and scoping it to exact literal matches keeps the risk low. Two gaps, both detailed inline:

1. Only grouped literals are matched, so a projected-but-not-grouped literal still emits a positional `ORDER BY`.
2. Sort and select exprs can both arrive as `Expr::Alias(Literal(..), name)`, so the raw equality under-matches; the grouped case only works because group exprs happen to be bare literals.

With the inline suggestion applied and the second test added this should be good to merge.

rust/cubesql/cubesql/src/compile/engine/df/wrapper.rs
jwils: Done. Thanks

MazterQyou (approved):

MazterQyou:
@jwils thanks for the contribution! The fix will be available in v1.7.38 onward.

PR #11502, opened 2026-08-07
fix(tesseract): parenthesize member SQL spliced into filter templates

waralexrom:
## Summary

A filter template places its own operator next to the member's rendered SQL (`{{ column }} = {{ value }}`, `{{ column }} IS NOT NULL`, …). A member whose `sql` is a bare expression then re-associates: when its own top-level operator binds weaker than the template's, that operator captures only the tail of the member expression.

Reported on Athena/Trino ([CORE-726](https://linear.app/cube-d3/issue/CORE-726/member-sql-not-parenthesized-when-interpolated-into-filter-templates)):

```yaml
- name: measure_1
  sql: amount
  type: sum
- name: measure_2
  sql: "{measure_1} IS NOT NULL"
  type: boolean
```

`measure_2 equals true` rendered `HAVING (sum("orders".amount) IS NOT NULL = CAST(? AS BOOLEAN))`, which Trino rejects with `mismatched input '='`. Aggregate-typed measures were safe by accident, being wrapped in their own function call.

Two aggravating factors: it is not measure-only (a dimension `sql: "amount IS NOT NULL"` breaks the same way in `WHERE`), and where the member's top level is `AND`/`OR` the mis-parse stays **valid SQL that silently returns a different row set** — no error at all.

Tesseract only, per the ticket; the legacy planner is left as is.

## Changes

- `FilterSqlContext` renders the member as a single operand at construction time, so every filter operator receives it already pinned. The field is now private behind `member_sql()` so a new call site cannot bypass the wrapping.
- Atomicity is decided by the existing `sql_expression_scanner::is_top_level_compound` — the same scanner `ParenthesizeSqlNode` already uses for `SqlCall` arguments. Plain columns, aggregates, casts, `CASE` and already-parenthesized expressions keep their shape, so the wrapping does not spread through the filters that never had the hazard. Across the whole repo exactly one existing assertion changed: `base-query.test.ts` required `WHERE (1 = 1 = ?)`, i.e. it had encoded invalid SQL.
- New `sql_expression_scanner::ends_in_line_comment`: an expression ending in `-- …` gets its closing parenthesis on a line of its own, since on the same line the comment would swallow it. (Not a regression — such a member was already fatal in filters, `amount + 1 -- note > $1` does not parse either — but the wrapping is the natural place to close it.)

## Testing

New `packages/cubejs-schema-compiler/test/integration/postgres/filter-member-sql-parens.test.ts` — 26 tests, **22 of which failed before the fix**.

Postgres works as a full-fidelity polygon because it exhibits both failure modes:

- `a > b = c` is a **syntax error** there (comparisons are non-associative), so a member `sql: "amount > 50"` under `equals`/`notEquals` hard-fails — in `WHERE`, and in `HAVING` over `sum(...)`. Closest analogue of the reported Trino error.
- A member with a top-level `AND`/`OR` diverges **silently**: row-level assertions cover equals / notEquals / IN / NOT IN / set / notSet against real data.
- `x IS NOT NULL = FALSE` happens to parse the intended way on Postgres, so the ticket's literal shape is pinned on the emitted SQL, with a comment explaining why.
- `gt/gte/lt/lte`, the LIKE family and the date operators cannot be made to diverge on Postgres (its precedence agrees), so those assert the whole rendered predicate — which also pins the operator and wildcard shape, not just the parentheses.
- Two guards that the wrapping stays **off** atomic members (plain column dimension, aggregate measure), so the diff cannot quietly spread.

| Suite | Result |
|---|---|
| `filter-member-sql-parens` (Tesseract) | 26/26 pass (22 failed before the fix) |
| same file under legacy planner | skips cleanly via `getEnv('nativeSqlPlanner')` |
| `dist/test/integration/postgres` (Tesseract) | 46 suites, 543 passed, 21 skipped |
| `dist/test/unit` (Tesseract) | 36 suites, 716 passed |
| `cargo test -p cubesqlplanner` | 1193 passed |
| `cargo test -p cubesqlplanner --features integration-postgres` | 1193 passed |

## Note for reviewers

The same trailing-line-comment shape exists in `sql_nodes/parenthesize.rs` and `sql_templates/plan.rs::convert_tz`. Left untouched to keep this diff scoped — happy to fold in if preferred.

### DIFF
--- a/rust/cube/cubesqlplanner/cubesqlplanner/src/physical_plan/filter/operators/filter_sql_context.rs
+++ b/rust/cube/cubesqlplanner/cubesqlplanner/src/physical_plan/filter/operators/filter_sql_context.rs
@@ -2,6 +2,7 @@ use crate::cube_bridge::base_query_options::FilterValue;
 use crate::planner::query_tools::QueryTools;
 use crate::planner::sql_templates::{PlanSqlTemplates, TemplateProjectionColumn};
 use crate::planner::QueryDateTimeHelper;
+use crate::utils::sql_expression_scanner::{ends_in_line_comment, is_top_level_compound};
 use cubenativeutils::CubeError;
 use std::rc::Rc;
 
@@ -15,7 +16,7 @@ enum DateBound {
 }
 
 pub struct FilterSqlContext<'a> {
-    pub member_sql: &'a str,
+    member_sql: String,
     pub query_tools: &'a Rc<QueryTools>,
     pub plan_templates: &'a PlanSqlTemplates,
     pub use_db_time_zone: bool,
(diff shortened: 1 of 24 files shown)


rust/cube/cubesqlplanner/cubesqlplanner/src/physical_plan/filter/operators/filter_sql_context.rs
```diff
@@ -15,14 +16,55 @@ enum DateBound {
...
+    // operator binds weaker than the operator a filter template puts beside it,
+    // the bare splice re-associates and that operator captures only the tail of
+    // the member expression — a syntax error on some dialects, a silently
+    // different predicate on the rest. Parentheses pin the whole expression as
+    // one operand; an atomic expression needs none and keeps its shape.
+    fn as_operand(member_sql: &str) -> String {
+        // An expression ending in a line comment swallows whatever the template
+        // appends on that line, so it needs the wrapping — and a line of its own
+        // for the closing parenthesis — however atomic it otherwise is.
+        if ends_in_line_comment(member_sql) {
```
ovr, line 59: @claude could you showcase an example of SQL? Why is this needed?

ovr: @waralexrom, about what comments is this code about?)

```
- name: test
   sql: "amount + 1 -- this comment?"
   type: number
```

such comments? ☝️

paveltiunov (approved):

PR #11597, opened 2026-08-19
docs(monitoring): document log fields and metric labels (CUB-2879)

RusovDmitriy:
## Summary

The Monitoring Integrations page documented *inputs*, *sinks*, metric *names* and the Query History field list, but said nothing about the tags actually attached to what gets exported. This adds three things to that page:

- **Fields in exported logs** — `deployment_id`, `deployment_name`, `hostname`, `pod_name`, `service`, `level`, `status`, `origin`, plus a `#### Datadog attributes` subsection for the `datadog_logs`-only `ddsource` / `ddtags` and the **previously undocumented per-sink `ddtags = [...]` option** (its values are appended to the defaults).
- **Labels on exported metrics** — the Prometheus labels on `cube_*` metrics, which is what gives the per-API request/latency breakdown asked for in CUB-2871, including that `request_source` is one of `ai-engineer` / `other` / `unknown`.
- **Query History export** — the missing `query_history_link` row, and a note that Query History records carry a *different* field set than container logs.

Every value was read out of the Cube Cloud log and metrics export pipeline rather than inferred from the ticket wording. That pipeline is not part of this repository, so the field lists cannot be verified against this checkout.

## Two deviations from the ticket

1. **Page location.** The ticket names `docs/content/product/administration/deployment/monitoring/index.mdx`. That is the legacy Nextra site, which `CLAUDE.md` and `docs-mintlify/CLAUDE.md` both mark deprecated ("do not add or edit content there"); its last touch was the #10544 reorg, while the live equivalent — `docs-mintlify/admin/monitoring/monitoring-integrations/index.mdx` — carries the same three sections and is actively maintained (most recently #11577, also a monitoring change). The change landed on the live page. No new page was created.
2. **Metric labels are not uniform across metrics.** The ticket says every `cube_*` metric carries all four labels; in practice a label is emitted only where it is set, and the resource metrics and the request metrics carry different ones. So `cube_cpu_usage_ratio` and `cube_memory_usage_ratio` carry `pod_name` only, while `cube_requests_*` carry `deployment_id`, `api_type` and `request_source` but no `pod_name`. The table documents it per metric family rather than as one blanket list.

Two smaller details documented the same way: `ddsource` is the pod name on container logs but `query-history` on Query History records, and `trace`-severity records are exported with `level`/`status` set to `debug` while remaining selectable as `trace` via the `levels` option.

Per the ticket's out-of-scope list: no tag was added or renamed, no product code changed, and the REST API request-id docs were not touched.

## Verification

- `npx mintlify@4 broken-links --check-anchors` over the whole site: **1 broken link, in `reference/control-plane-api.mdx`** — a file this PR does not touch. Every link and anchor added here resolves, including the new `#fields-in-exported-logs` cross-reference.
- Structural check of the edited file: code fences balanced, no table column mismatches, no undefined `[ref-*]` link references.
- The `yarn build` script in `docs-mintlify/package.json` is stale — `build` is not a command in the Mintlify 4 CLI, so it could not be run. `docs-mintlify` has no lint or format script, and this repo has no `.husky` hooks, so nothing else applies to an MDX-only change.

### DIFF
--- a/docs-mintlify/admin/monitoring/monitoring-integrations/index.mdx
+++ b/docs-mintlify/admin/monitoring/monitoring-integrations/index.mdx
@@ -155,7 +155,7 @@ Before a log record reaches a sink, the following fields are attached to it:
 | `deployment_name` | Name of the deployment. |
 | `hostname` | Host name of the node that the record comes from. |
 | `pod_name` | Name of the pod that the record comes from. |
-| `service` | Name of the component that the record comes from, matching the input name: `cubejs-server`, `refresh-scheduler`, `warmup-job`, or `cubestore`. |
+| `service` | Name of the component that the record comes from, matching the input name: `cubejs-server`, `refresh-scheduler`, `warmup-job`, or `cubestore`. Records from the `query-history` input use `query-history` on `datadog_logs` sinks. |
 | `level` | Severity of the record, lower-cased: `error`, `warn`, `info`, or `debug`. Records with the `trace` severity are exported as `debug`, and records that carry no severity as `info`. |
 | `status` | Duplicate of `level`, for tools that expect the severity in a `status` field, e.g. Datadog. |
 | `origin` | Always `cube_cloud`. |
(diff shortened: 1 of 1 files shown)


RusovDmitriy:
All three review findings applied in f420810:

1. **`$DEPLOYMENT_NAME,$HOSTNAME`** — taken your wording; the collision with the `CUBE_CLOUD_MONITORING_*` section above is real, those two are substituted by Cube and can't be set.
2. **Third `[sinks.datadog]` block** — trimmed to the new option, as suggested. The two full blocks above it already cover the copy-pasteable case.
3. **`service` on Query History records** — added the clause to the `service` row, so the fifth value is visible where a reader builds the facet rather than only 250 lines later.

On the unverifiable part: correct that the four sources of truth live in `cubejs-enterprise` and can't be checked from this repo. For a reviewer who does have access, the specific claims and where they come from:

- field lists and the two branches — `packages/console-server/resources/vector/remap-final.vrl` (`.origin` is set outside the if/else, which is why it is the one field Query History records keep);
- `trace` → `debug` — same file, `if .level == "trace" { .level = "debug" }`, and the filter transform is wired *before* the remap in `VectorConfigGenerator`, so `levels = ["trace"]` still selects those records;
- `ddtags` defaults and the sink option — `VectorConfigGenerator.getRemapTransform()`;
- the per-metric label split — `MetricsService.formatLabels()` pushes each label only when its value is set, and `K8sMetrics` sets only `podName` while `RequestsMetrics` sets only `deploymentId`/`apiType`/`requestSource`;
- `request_source` values — `RequestsMetrics.normalizeRequestSource()`, allowlist `{'ai-engineer'}` → everything else `other`, empty → `unknown`.

MikeNitsenko (approved):

PR #11643, opened 2026-08-25
fix(tesseract): fold nested multi-fact join legs into one scan

waralexrom:
## Summary

A query whose measures sit at different depths of one join chain planned a separate leg
per measure, and every leg re-scanned the shared part of the join from scratch. A funnel
modelled as `sites → carts → checkouts` paid for one full scan of the base per measure,
so adding measures multiplied the cost — two measures were already timing out on Trino.

Such legs are now folded into one wherever moving a measure into the wider join tree
provably cannot change what it computes. Note that this is not the star-shaped multi-fact
case (`orders` and `returns` under `customers`), which stays split as before — it is the
case where legs differ only by how far down one chain they walk, and never needed to be
separate at all.

## Changes

- `MultiFactJoinGroups::try_new_merging_nested` folds a group into one whose join tree
  contains its own. Safe because the extra joins are `LEFT` — they only ever replicate a
  row, never drop one — and only for measures immune to that replication: a distinct
  aggregation qualifies, a plain `count` or `sum` does not and keeps its own leg.
- The wider tree is rebuilt from the union of both groups' join hints before the fan-out
  question is asked. `multiplicationFactor` is only populated for the cubes a tree's own
  hints named (`JoinGraph.ts` maps it over `cubesToJoin`), so a cube reached in transit
  reports as unmultiplied and would otherwise let an inflating measure through. The
  rebuild produces the same joins — asserted by comparing the `JoinKey`.
- `MeasureKind::survives_row_multiplication`, deliberately narrower than the existing
  `regular_in_multiplied`: a key-based count is also safe under multiplication, but only
  after switching to the distinct `MultipliedCount` form, and the render form is decided
  from the measure's own tree elsewhere.
- Merging stands down for: multi-stage measures (own CTE pipeline), member expressions
  (`COUNT(*)` names no member, so what it counts is whatever rows the join produces),
  ungrouped queries (raw rows expose the replication), pre-aggregation queries, and
  regrouping through `for_measures`.
- It also stands down when any cube read by the groups defines a pre-aggregation at all.
  A rollup is matched against one leg at a time, so a merged query needs one spanning all
  of them; a rollup outweighs the scan saved. This guard is coarse — see below.

## Testing

New fixture with a three-cube chain and a seed built so right and wrong answers differ:
two carts of one site share a `msid` (distinct ≠ count), one cart has two checkouts
(fan-out), two carts have none (`LEFT JOIN` must keep them).

- 6 integration tests on Postgres, snapshots verified by hand against the seed. The key
  one: a distinct measure returns the same numbers selected alone and selected next to a
  deeper measure, while `count`/`sum` keep their own leg and their own values.
- 2 structural tests (scan count, per-leg rollups still matching) and 5 unit tests on the
  grouping itself — nested merges, siblings do not, plain `count` does not.
- Full suite green on real Postgres: 1287 tests, no existing snapshot changed.

## Follow-up

The pre-aggregation guard answers "could a rollup exist" rather than "did one match", so
the optimisation silently switches off for models with rollups on the fact cubes. Removing
it means two-pass planning in `TopLevelPlanner` — plan unmerged, try pre-aggregations,
replan merged on a miss — which needs `QueryProperties` to carry a merge flag. Left out of
this PR deliberately.

The merge also only takes effect through `is_simple_query()`, so a query that additionally
carries a multiplied or multi-stage measure keeps its regular legs unmerged.

### DIFF
--- a/rust/cube/cubesqlplanner/cubesqlplanner/src/planner/multi_fact_join_groups.rs
+++ b/rust/cube/cubesqlplanner/cubesqlplanner/src/planner/multi_fact_join_groups.rs
@@ -325,7 +325,12 @@ impl MultiFactJoinGroups {
             .map(|key| grouped.remove(&key).unwrap())
             .collect::<Vec<_>>();
 
-        if merge_nested && !Self::any_cube_has_pre_aggregations(query_tools, &groups)? {
+        // Cheapest question first: without a nested pair there is nothing to
+        // merge, and the pre-aggregation scan crosses the bridge once per cube.
+        if merge_nested
+            && Self::has_nested_pair(&groups)
+            && !Self::any_cube_has_pre_aggregations(query_tools, &groups)?
+        {
             Self::merge_nested_groups(&mut groups, &resolve)?;
         }
 
@@ -333,6 +338,16 @@ impl MultiFactJoinGroups {
             .into_iter()
             .map(|group| (group.tree, group.measures))
             .collect())
+    }
+
+    /// Whether any group's join tree is contained in another's - the only
+    /// shape `merge_nested_groups` can do anything with.
+    fn has_nested_pair(groups: &[GroupBuild]) -> bool {
+        groups.iter().any(|group| {
+            groups
+                .iter()
+                .any(|other| group.key.is_nested_in(&other.key))
+        })
     }
 
     /// Whether any cube these groups read defines a pre-aggregation.
@@ -405,7 +420,12 @@ impl MultiFactJoinGroups {
                     }
                     let mut hints = other.hints.clone();
                     hints.extend(&group.hints);
-                    let (key, tree) = resolve(&hints)?;
+                    // Resolving is a probe: a hint set the join graph refuses
+                    // means there is no merge to make here, not that the query
+                    // the groups came from is unplannable.
+                    let Ok((key, tree)) = resolve(&hints) else {
+                        continue;
+                    };
                     if key != other.key {
                         continue;
                     }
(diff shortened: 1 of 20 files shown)


waralexrom:
Разобрал замечания, все пять по делу. Пофикшено в 902fa44140.

**Speculative `resolve` failing the query** — согласен, это был тот же класс, что уже был вылечен для `collect_multiplied_measures`, но `resolve` я пропустил. Теперь `let Ok(..) else { continue }`. Стоит отметить, что `resolve` вызывается только после `is_nested_in`, так что путь для объединения хинтов почти всегда существует — но «почти» здесь не аргумент, зонд не должен валить запрос.

**Pre-aggregation scan on every query** — прав, и вариант с `is_nested_in` лучше, чем `len() > 1`: звёздный multi-fact (`orders`/`returns`) тоже платил бы зря. Добавил `has_nested_pair` — чистое сравнение ключей, идёт первым.

**`is_nested_in` docstring** — тоже прав, это containment множества рёбер, а не пути; sibling-ветка проходит. Переписал так, чтобы комментарий не обещал больше, чем проверка делает.

**`min`/`max`** — согласен по существу: они идемпотентны к размножению и рендерятся тем же SQL. Оставляю на отдельный PR, как ты и предлагаешь — там нужна своя проверка на значениях, и я бы не расширял охват мержа в том же PR, где он только что получил три сужения по корректности. Завёл в голове как follow-up вместе со снятием грубого pre-agg гейта.

**Тесты** — добавил оба недостающих: мерж с фильтром на мелком кубе (мерж остаётся, значения сверены) и на глубоком (там деревья и так схлопываются в одно), плюс ветку `pre_aggregation_query`. Снапшоты считал руками по сиду.

Про `base_scan_count`: замечание справедливое, привязка к алиасу хрупкая. Оставил как есть осознанно — структурная часть уже покрыта юнит-тестами на `num_groups()` в `multi_fact_join_groups.rs`, а в интеграционных тестах проверяется именно отрендеренная форма, где число сканов и есть то, что PR меняет. Если алиасинг поедет, тесты упадут громко, а не тихо пройдут.

Полный прогон на реальном Postgres: 1290 тестов, зелёные.

ovr (approved):

PR #11809, opened 2026-09-09
fix(schema-compiler): reject multiple joins to the same cube

waralexrom:
## Problem

A cube may declare several joins to the same cube today, and the model still
compiles. All but one of those joins are then dropped without a word: queries
run, results look plausible, and they are produced through a join path the
author never wrote. There is no way to notice this from the outside — no error,
no warning, no difference in the API.

```yaml
cubes:
  - name: orders
    sql_table: orders_tbl
    joins:
      - name: users
        sql: "{CUBE}.user_id = {users}.id"      # silently ignored
        relationship: many_to_one
      - name: users
        sql: "{CUBE}.manager_id = {users}.id"   # the one that is used
        relationship: many_to_one
```

```sql
SELECT "users".name "users__name", count("orders".id) "orders__count"
FROM orders_tbl AS "orders"
LEFT JOIN users_tbl AS "users" ON "orders".manager_id = "users".id
GROUP BY 1
```

## Cause

`JoinGraph` keys its edges by the pair of cubes a join connects
(`${cube.name}-${join.name}`) and builds the edge map with `R.fromPairs`, so
several declarations for the same pair collapse into one entry and the last one
wins. The dropped declarations leave no trace.

## What was done

`JoinGraph.buildJoinEdges` now reports this as a model compilation error:

```
orders.yml Errors:
orders cube: Cube 'orders' declares 2 joins to 'users' (joins[0], joins[1]).
Only one join per pair of cubes is supported. Keep a single join to 'users', or
use extends to create a child cube of 'users' and join that instead
```

The message names the file, the cube, every conflicting declaration by its
position in the list, and the way out.

The check sits in the function that builds the edges, next to the collapse it
guards against and alongside the two failures already handled there the same way
(a join to a cube that does not exist, a missing primary key): report it, and
leave the join out. So the failure is scoped to the join rather than the cube —
the cube stays valid.

The strict path fails on the reported error. An embedder compiling with
`omitErrors: true` keeps a cube that serves its own members and has lost only the
ambiguous join, and *none* of the conflicting declarations is used, so a query
needing that join fails loudly with `Can't find join path` instead of quietly
picking one. Invalidating the cube instead would have made every *other* cube
joining to it report `Cube orders doesn't exist`, which is both false and
unrelated to the defect.

One corner is left open: a cube that inherits an ambiguous `joins` block through
`extends` without declaring any joins of its own still resolves through the
collapsed edge under `omitErrors: true`. The error is still reported, on the cube
that declares the duplicates; deciding it for the child means walking the
`extends` chain to find which level last declared the target, which is more
machinery than the case is worth.

The rule is documented in the joins troubleshooting page, together with the
`extends` recipe for joining the same table through two different keys.

Only the joins a cube declares itself are checked. `extends` appends a child's
joins to the inherited ones, and a child redeclaring an inherited join replaces
it — that stays valid, and duplicates a parent declares are reported once, on
the parent. The join map form (`joins: { users: { … } }`) is keyed by the joined
cube name and cannot hold duplicates in the first place.

The check deliberately stops at "one join per joined cube". Allowing several
aliased joins to the same cube needs join aliases, which do not exist yet;
relaxing the condition then means loosening a single `if` in one place.

## Compatibility

**This is a behavior change: models that compile today start failing.** The
shape that breaks is exactly one — a single cube listing the same joined cube
more than once in its own `joins` list, which is only expressible in the YAML /
array form. Every such model is already broken in a worse way: it runs through a
join path its author did not write, and the two forms are indistinguishable from
the API. Failing at model compilation, with the conflicting declarations named,
is strictly more informative than losing a join in silence.

Blast radius in this repository: **zero**. A scan of all 255 list-form `joins`
blocks across packages, examples and docs — 51 of which declare more than one
join — found no duplicate target outside the new test, and the schema-compiler
unit suite is unchanged (960 passing; the 2
remaining `error-reporter` snapshot failures are pre-existing ANSI-colour
mismatches, present on `master` too). Scaffolding is unaffected: it reduces
generated joins into a map keyed by the joined cube before rendering, so two
foreign keys to the same table already produce a single join.

Deployments that hit this will need a data model change — dropping the extra
join, or using `extends` to create a second cube over the same table.

## How it was verified

- New unit test `duplicate-cube-joins.test.ts`: 3 rejection cases and 4 shapes
  that must keep compiling (joins to different cubes, the same pair joined from
  both sides, a transitive path, an `extends` override). Without the fix the 3
  rejection cases fail, the other 4 pass.
- Both planners: the whole file passes with `CUBEJS_TESSERACT_SQL_PLANNER=true`
  and `=false` (native addon built). The error is raised while the data model is
  compiled, before any query planning, so it does not depend on the planner.
- `packages/cubejs-schema-compiler` unit suite: 960/962, the 2 failures
  pre-existing on `master`.
- Both declaration forms that can express the duplicate: the YAML list form and
  the JS array form.
- The `extends` recipe the docs now recommend: verified it compiles and emits
  two distinct `ON` clauses against the same table.

## Risks

The error fires at model compilation, so a deployment carrying this shape stops
serving until its data model is fixed, rather than serving quietly wrong join
paths. That is the intended trade-off, but it is a breaking change for anyone
who has the shape in production.

A review of this diff suggested softening it to `errorReporter.warning` for one
release, or gating it behind an env flag, so operators would see it in logs
before it turns fatal. That is deliberately not done here: a warning is still a
model that silently uses a join path nobody wrote, which is the thing this
change exists to end. Whether the rollout deserves a grace period is a release
decision rather than a code one, and the warning path stays a one-line change if
that call is made.

### DIFF
--- a/packages/cubejs-schema-compiler/src/compiler/CubeValidator.ts
+++ b/packages/cubejs-schema-compiler/src/compiler/CubeValidator.ts
@@ -1402,6 +1402,9 @@ export function functionFieldsPatterns(): string[] {
 export class CubeValidator implements CompilerInterface {
   protected readonly validCubes: Map<string, boolean> = new Map();
 
+  // cube name -> names of the cubes it declares more than one join to
+  protected readonly duplicateJoins: Map<string, Set<string>> = new Map();
+
   public constructor(
     protected readonly cubeSymbols: CubeSymbols
   ) {
@@ -1429,9 +1432,7 @@ export class CubeValidator implements CompilerInterface {
       if (!this.validateGranularitySql(cube, errorReporter)) {
         valid = false;
       }
-      if (!this.validateUniqueJoins(cube, errorReporter)) {
-        valid = false;
-      }
+      this.validateUniqueJoins(cube, errorReporter);
     }
 
     if (result.error != null) {
@@ -1483,24 +1484,18 @@ export class CubeValidator implements CompilerInterface {
     return result;
   }
 
-  // Only one join per joined cube survives into the join graph, so any extra
-  // declaration would be dropped and the model would resolve through a path its
-  // author did not write.
-  //
-  // Read from the raw definition, which holds only the joins the cube declares
-  // itself: `extends` appends a child's joins to the inherited ones, and a child
-  // redeclaring an inherited join is a supported way to replace it. Reporting
-  // over the merged list would blame the child for the replacement and point at
-  // positions it never wrote.
-  private validateUniqueJoins(cube, errorReporter: ErrorReporter): boolean {
+  private validateUniqueJoins(cube, errorReporter: ErrorReporter) {
+    // The raw definition, not `cube.joins`: `extends` merges the parent's joins
+    // in, and a child redeclaring one of them is a supported override
     const ownJoins = this.cubeSymbols.cubeDefinitions[cube.name]?.joins;
+    const duplicates = new Set<string>();
+    this.duplicateJoins.set(cube.name, duplicates);
 
     // The map form is keyed by the joined cube name and can not hold duplicates
     if (!Array.isArray(ownJoins)) {
-      return true;
+      return;
     }
 
-    let valid = true;
     const declarationsByCube = new Map<string, number[]>();
 
     ownJoins.forEach((join, index) => {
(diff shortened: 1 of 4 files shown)


packages/cubejs-schema-compiler/src/compiler/CubeValidator.ts
```diff
@@ -1402,6 +1402,9 @@ export function functionFieldsPatterns(): string[] {
 export class CubeValidator implements CompilerInterface {
   protected readonly validCubes: Map<string, boolean> = new Map();
 
+  // cube name -> names of the cubes it declares more than one join to
+  protected readonly duplicateJoins: Map<string, Set<string>> = new Map();
```
ovr: @claude I guess we should clean it after validation, do we have a place/method to do it?

packages/cubejs-schema-compiler/src/compiler/CubeValidator.ts
```diff
@@ -1478,6 +1484,41 @@ export class CubeValidator implements CompilerInterface {
     return result;
   }
 
+  private validateUniqueJoins(cube, errorReporter: ErrorReporter) {
+    // The raw definition, not `cube.joins`: `extends` merges the parent's joins
+    // in, and a child redeclaring one of them is a supported override
+    const ownJoins = this.cubeSymbols.cubeDefinitions[cube.name]?.joins;
+    const duplicates = new Set<string>();
+    this.duplicateJoins.set(cube.name, duplicates);
```
ovr: @claude is it possible to rework the PR without storing duplicateJoins in global state behind CubeValidator?

ovr (approved):

PR #11812, opened 2026-09-09
fix(tesseract): retry pre-aggregation matching within one external type

waralexrom:
## Problem

A multi-stage measure whose stages match *different* pre-aggregations gets no
pre-aggregation at all when those matches land in different external types:
`/sql` reports `preAggregations: []` and the query reads the raw source. With
both matches in the same external type the query is served normally.

Reported shape: a `brand_share` measure whose denominator lifts the brand
filter through `filter.exclude`. The numerator matches a brand-grained rollup
(`external: true`, CubeStore), the denominator a day-grained one
(`external: false`, source DB) — and the query falls through to the fact table,
the worst available outcome on a large table.

## Cause

Stages are matched independently, each taking the **first** candidate that fits
in declaration order. Once every stage has matched, a set spanning both
external types is refused — correctly, since no single query can read CubeStore
and the source database at once.

The defect is that the refusal was **terminal**. There is no retry, so a set
that could have been covered by one wider pre-aggregation in a single engine
was discarded along with everything else. Whether the query was served came
down to the order the pre-aggregations happen to be declared in: declaring the
wider one first makes every stage reach it and the query is served, at the cost
of plain queries matching the wide rollup instead of the precise one.

## What changed

A matching pass now reports *why* it failed instead of a bare `None`:

- `UnmatchedStage` — a stage matched no candidate. Returns immediately, since
  matching is monotone in the candidate set and no subset can cover it.
- `ExternalTypesSplit` — every stage matched, but across both external types.
  Matching is retried against candidates of a single external type, CubeStore
  first as the default and faster store.

This is the fallback order the reported case needs: most precise per stage,
then one engine, then the fact table. Per-stage precision is given up only in
the retry, and only where the alternative is reading the fact table.

Two properties keep this contained:

- **Plain queries are untouched.** A query that matches as a whole is rewritten
  before the multi-stage path is reached, so the retry cannot move it onto the
  wider pre-aggregation. The retry only ever fires on a set that today yields
  zero pre-aggregations.
- **No new correctness surface.** The retry re-runs the same matching predicate
  over a subset of candidates, so it can only produce plans the matcher already
  sanctioned — the same plans that reordering the model would have produced.

## How it was verified

Eight tests on three fixtures sharing one seed table. Three fail without the
fix; five hold the surrounding behaviour in both directions.

Failing without the fix:

- The reported shape — stages split across external types — now served from
  `by_brand_day` twice instead of `preAggregations: []`.
- Row-level parity against the fact table on Postgres. The denominator now
  re-aggregates across the brands stored in the brand-grained rollup, and the
  shares come back `0.30` / `0.40`, matching the fact-table plan exactly. A
  brand predicate leaking into that read would show as a share of `1.0`.
- The engine preference, on a fixture whose brand-grained rollup exists in
  **both** external types so either group can serve the query. Reversing the
  retry order fails this test; without it the preference was unexercised.

Holding the surroundings:

- One wider pre-aggregation covering every stage on its own, and the narrow one
  covering nothing.
- A whole-query match left untouched by the retry.
- A split where *neither* external type covers both stages: both retries are
  exhausted and the query reads the fact table.
- `disable_external_pre_aggregations`: externals are dropped before matching, so
  no pass can split, the retry never runs, and the outcome is unchanged.

Rebased onto current master (which does not touch this optimizer). Full
`cubesqlplanner` suite: 1420 passed, 0 failed. `cargo fmt` and `cargo clippy`
clean.

## Risks

- The retry adds up to two extra matching passes, but only on a set split
  across external types — a case that currently returns nothing. A stage that
  matched nothing still fails on the first pass, so queries with no usable
  pre-aggregation do not pay for it.
- When both external groups can cover the query, CubeStore wins. That is a
  policy choice rather than a cost estimate; there is no ranking between
  candidates anywhere in the matcher today.
- Queries in the reported shape start hitting a pre-aggregation where they
  previously read the source, so their results now depend on rollup freshness
  like any other pre-aggregated query.
- `disable_external_pre_aggregations` is unaffected: external candidates are
  filtered out before matching, so a pass can never be split and the retry
  never runs. Covered by a test rather than left as an argument.

### DIFF
--- a/rust/cube/cubesqlplanner/cubesqlplanner/src/logical_plan/optimizers/pre_aggregation/optimizer.rs
+++ b/rust/cube/cubesqlplanner/cubesqlplanner/src/logical_plan/optimizers/pre_aggregation/optimizer.rs
@@ -48,6 +48,17 @@ enum RowGrain {
     RawRows(Option<Rc<LogicalJoin>>),
 }
 
+// Outcome of one matching pass over a single set of candidate pre-aggregations.
+enum MultiStageMatch {
+    Matched(Rc<RootQuery>),
+    // A stage matched no candidate, so no subset of the candidates covers it
+    // either.
+    UnmatchedStage,
+    // Every stage matched, but the matches span both external types, so no one
+    // query can read them all.
+    ExternalTypesSplit,
+}
+
 pub struct PreAggregationOptimizer {
     query_tools: Rc<State>,
     allow_multi_stage: bool,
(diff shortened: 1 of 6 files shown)


rodrigobercini:
Thanks for working on implementing this feature! It will let pre-aggregations cover many more queries.

PR #11830, opened 2026-09-09
chore(ci): Claude Code review - small improvements

ovr:
The Claude PR review instructions lived inline in the workflow's `prompt:` block, so they could only ever run in CI. They now live in `.claude/commands/cube-review.md`, and the workflow just invokes `/cube-review <repo> <pr-number>` — so anyone can run the same review locally with `/cube-review` on a branch with an open PR. Review-thread hygiene (resolving your own stale threads, not re-posting duplicate inline comments) moved one level further out to `.claude/docs/review-threads.md`, pulled in by an `@` reference; `Read` was added to `--allowedTools` so that reference resolves even if the `@` substitution does not happen. The prompt text itself is unchanged apart from dropping the CI-only `REPO:`/`PR NUMBER:` header in favour of a target line that falls back to `gh pr view` when no arguments are passed.

**Check List**
- [x] Linter has been run for changed code (`actionlint`)

### DIFF
--- a/.claude/docs/review-threads.md
+++ b/.claude/docs/review-threads.md
@@ -12,7 +12,8 @@ preconfigured alias:
 
 For each thread where ALL of the following hold:
   - `isResolved` is false
-  - the first comment's `author.login` is `claude`
+  - the first comment's `author.login` is yours — `claude` in CI, otherwise
+    the login `gh api user -q .login` returns
   - the concern is no longer applicable in the current diff (file/line gone,
     code rewritten, issue addressed)
 
@@ -21,7 +22,9 @@ resolve it with:
   gh resolve-thread <thread-id>
 
 Do not resolve threads from human reviewers under any circumstance, even if the
-concern looks addressed — leave that decision to the reviewer. Only the two
+concern looks addressed — leave that decision to the reviewer. That includes the
+human whose login you are running under: resolve only threads you recognise as
+output of an earlier review round. Only the two
 aliases above are available; raw `gh api graphql` is not permitted.
 
 **Avoiding duplicate inline comments:**
@@ -32,7 +35,8 @@ deduplicate against your own prior comments. Before calling
 an existing thread already covers it. Skip creating a new inline comment when ALL
 of the following hold for any thread in the list:
   - `isResolved` is false
-  - the first comment's `author.login` is `claude`
+  - the first comment's `author.login` is yours — `claude` in CI, otherwise
+    the login `gh api user -q .login` returns
   - the thread is on the same `path` and `line` as the new issue you would post
   - the existing comment's body raises substantively the same concern (same root
     cause, same fix direction — wording does not need to match)
--- a/.claude/docs/review-threads.md
+++ b/.claude/docs/review-threads.md
@@ -1,0 +1,46 @@
+# Review threads
+
+Rules for handling your own prior review threads on a PR. Referenced by
+`.claude/commands/cube-review.md`.
+
+**Resolving your own stale review threads:**
+
+Before posting new comments, list existing review threads on this PR using the
+preconfigured alias:
+
+  gh list-review-threads <owner> <repo> <pr-number>
+
+For each thread where ALL of the following hold:
+  - `isResolved` is false
+  - the first comment's `author.login` is `claude`
+  - the concern is no longer applicable in the current diff (file/line gone,
+    code rewritten, issue addressed)
+
+resolve it with:
+
+  gh resolve-thread <thread-id>
+
+Do not resolve threads from human reviewers under any circumstance, even if the
+concern looks addressed — leave that decision to the reviewer. Only the two
+aliases above are available; raw `gh api graphql` is not permitted.
+
+**Avoiding duplicate inline comments:**
+
+Use the same `gh list-review-threads` output (from the resolve step above) to
+deduplicate against your own prior comments. Before calling
+`mcp__github_inline_comment__create_inline_comment` for a new issue, check whether
+an existing thread already covers it. Skip creating a new inline comment when ALL
+of the following hold for any thread in the list:
+  - `isResolved` is false
+  - the first comment's `author.login` is `claude`
+  - the thread is on the same `path` and `line` as the new issue you would post
+  - the existing comment's body raises substantively the same concern (same root
+    cause, same fix direction — wording does not need to match)
+
+When you skip, briefly note it in your top-level summary (e.g. "Re-affirmed N
+prior threads still apply") so the reader knows you considered the issue. Do not
+post a "still applies" reply on the thread — silence is fine; the unresolved
+state already communicates that.
+
+If the issue is on a different line or the fix direction has shifted (different
+root cause), post a new inline comment as usual.
--- a/.claude/docs/review-threads.md
+++ b/.claude/docs/review-threads.md
@@ -24,8 +24,8 @@ resolve it with:
 Do not resolve threads from human reviewers under any circumstance, even if the
 concern looks addressed — leave that decision to the reviewer. That includes the
 human whose login you are running under: resolve only threads you recognise as
-output of an earlier review round. Only the two
-aliases above are available; raw `gh api graphql` is not permitted.
+output of an earlier review round. Only the two aliases above are available;
+raw `gh api graphql` is not permitted.
 
 **Avoiding duplicate inline comments:**
 
--- a/.claude/docs/review-threads.md
+++ b/.claude/docs/review-threads.md
@@ -8,7 +8,7 @@ Rules for handling your own prior review threads on a PR. Referenced by
 Before posting new comments, list existing review threads on this PR using the
 preconfigured alias:
 
-  gh list-review-threads <owner> <repo> <pr-number>
+  gh list-review-threads cube-js cube <pr-number>
 
 For each thread where ALL of the following hold:
   - `isResolved` is false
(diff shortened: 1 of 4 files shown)


.github/workflows/claude-code-review.yml
```diff
@@ -72,115 +73,11 @@ jobs:
           track_progress: true
           # Allow PRs opened by coding-agent bots (e.g. cursor) to trigger reviews.
           allowed_bots: "cursor[bot]"
```
ovr, line 97: **High — the slash command will not expand; the review prompt is silently lost.**

`track_progress: true` forces the action into **tag mode** (this PR's own run confirms it: `Auto-detected mode: tag for event: pull_request`, [run 34375293075](https://github.com/cube-js/cube/actions/runs/34375293075)). In tag mode the `prompt` input is not the user message — it is appended *inside* the default tag prompt:

```ts
// claude-code-action@v1 src/create-prompt/index.ts
if (context.githubContext?.inputs?.prompt) {
  return defaultPrompt + `\n\n<custom_instructions>\n${context.githubContext.inputs.prompt}\n</custom_instructions>`;
}
```

The only path that produces a slash-command-capable message is the separate `claude-user-request.txt` block (`base-action/src/run-claude-sdk.ts`), and that file is written **only** when `extractUserRequestFromContext` finds the `@claude` trigger phrase in a comment body or the issue/PR body — never true for a `pull_request`-triggered review of a PR whose description does not contain `@claude`.

So the CLI receives one text block whose *middle* contains `/cube-review 11830`. Slash commands only expand when the message starts with them. I verified this directly:

```
$ claude -p "Context block.\n\n<custom_instructions>\n/argprobe 4242\n</custom_instructions>\n\n…does ZORKMID appear anywhere in this prompt?"
NOTSEEN     # command body never injected
$ claude -p "/argprobe 4242"   # leading position
ARGS=4242   # expands, and unreferenced args are appended even without $ARGUMENTS
```

Net effect after merge: a generic tag-mode review whose sole custom instruction is the literal string `/cube-review 11830`. Every focus area, the comment rules, the collapsed-tracking-comment format and the thread hygiene all disappear — with no error, so it will look like it worked.

This PR cannot catch it itself: the run above was skipped with *"Workflow validation failed. The workflow file must exist and have identical content to the version on the repository's default branch."* The regression only surfaces once this is on `master`.

Suggested fix — keep the single source of truth but inline the body, which works regardless of slash-command handling:

```yaml
      - name: Load review prompt
        run: |
          {
            echo 'REVIEW_PROMPT<<CUBE_REVIEW_EOF'
            sed '1{/^---$/,/^---$/d}' .claude/commands/cube-review.md
            echo 'CUBE_REVIEW_EOF'
          } >> "$GITHUB_ENV"

      - name: Review
        with:
          prompt: |
            PR NUMBER: ${{ env.PR_NUMBER }}

            ${{ env.REVIEW_PROMPT }}
```

Note that this also drops the `@`-reference: `@.claude/docs/review-threads.md` is expanded by the *CLI* at command-load time, so with the prompt inlined you would need to concatenate that file too (or fold it back into the command file).

.github/workflows/claude-code-review.yml
```diff
@@ -72,115 +73,11 @@ jobs:
...
-            prior threads still apply") so the reader knows you considered the issue. Do not
-            post a "still applies" reply on the thread — silence is fine; the unresolved
-            state already communicates that.
-
-            If the issue is on a different line or the fix direction has shifted (different
-            root cause), post a new inline comment as usual.
+          prompt: "/cube-review ${{ env.PR_NUMBER }}"
 
           # Tools for comprehensive PR review
           claude_args: |
```
ovr, line 106: **Medium (security) — adding `Read` here widens what an untrusted PR head can do.**

(Anchored here because the `checkout` step on line 60 is outside the diff.)

Not a live issue today only because of the bug flagged on line 75; it becomes live the moment the command file is actually loaded.

`refs/pull/<N>/head` is the contributor's tree, so `.claude/commands/cube-review.md` is attacker-controlled. GitHub protects the *workflow* file here — for `workflow_call` / `issue_comment` it always runs the default-branch copy, and the action adds its own guard ("must have identical content to the version on the repository's default branch"). Neither guard covers `.claude/**`. A fork PR that touches only the command file passes both.

Reachable via the `/bot-review` proxy in `claude-code.yml`: it gates on the **commenter** being OWNER/MEMBER, not on the PR author. A maintainer typing `/bot-review` on a fork PR runs attacker-authored instructions with `ANTHROPIC_API_KEY`, `pull-requests: write`, `Bash(gh pr comment:*)`, `Bash(gh resolve-thread:*)` and (new here) unrestricted `Read`. That is enough to post arbitrary content as `claude[bot]`, mass-resolve review threads, and read runner-local files.

Before this PR the prompt was pinned in the workflow file and therefore always the trusted default-branch copy — so this is a regression, not a pre-existing gap.

Mitigation — restore `.claude/` from the base ref after checkout:

```yaml
      - name: Restore trusted review command
        run: |
          git fetch --depth 1 origin "${BASE_SHA}"
          git checkout FETCH_HEAD -- .claude/
        env:
          BASE_SHA: ${{ github.event.pull_request.base.sha || github.sha }}
```

Worth double-checking the widened `Read` too: the reviewer's job is served by `gh pr diff`, and `Read` is what turns "can post a comment" into "can read anything on the runner and post it".

.claude/docs/review-threads.md
```diff
@@ -0,0 +1,46 @@
...
+
+Rules for handling your own prior review threads on a PR. Referenced by
+`.claude/commands/cube-review.md`.
+
+**Resolving your own stale review threads:**
+
+Before posting new comments, list existing review threads on this PR using the
+preconfigured alias:
+
+  gh list-review-threads cube-js cube <pr-number>
```
ovr, line 11: Low — these command lines are indented by 2 spaces, which is not a code block in Markdown (4 spaces, or a fence). It did not matter inside the YAML block scalar; now that this is a real `.md` GitHub renders it, both alias invocations collapse into the preceding paragraph and `<pr-number>` / `<thread-id>` get eaten as HTML tags. Same on line 21.

ovr (commented):
Nice refactor in intent — one source of truth for the review prompt, runnable locally. But I believe the CI half does not actually work; details inline.

Nit not worth an inline thread: `cube-js/cube` is now hardcoded in the command while `gh pr diff` / `gh pr comment` still resolve against the checkout's remote. Identical in CI, divergent in a fork clone.

ovr:
**Review complete — 1 high, 1 medium, 1 low.** The refactor's intent is right, but the CI half almost certainly does not work: `/cube-review` never expands under `track_progress: true`.

<details>
<summary>Full review</summary>

- [x] Resolve PR and fetch diff
- [x] Check existing review threads (none)
- [x] Code quality
- [x] Security
- [x] Performance
- [x] Testing
- [x] Documentation
- [x] Comments
- [x] Post inline findings

### Findings

| # | Sev | File | What |
|---|-----|------|------|
| 1 | High | `.github/workflows/claude-code-review.yml:75` | `track_progress: true` forces tag mode, which buries `prompt` inside `<custom_instructions>` in the middle of the message. Slash commands only expand in leading position, so the entire review prompt is silently dropped. |
| 2 | Medium | `.github/workflows/claude-code-review.yml:79` | The prompt now originates from the PR head checkout, which is attacker-controlled on fork PRs and covered by neither GitHub's default-branch workflow rule nor the action's own workflow-content guard. `Read` widens the blast radius. |
| 3 | Low | `.claude/docs/review-threads.md:11,21` | 2-space indentation is not a Markdown code block; the alias examples render as prose and `<pr-number>` / `<thread-id>` are swallowed as HTML. |

### How finding 1 was established

1. This PR's own review run logs `Auto-detected mode: tag for event: pull_request`.
2. `claude-code-action@v1`, `src/create-prompt/index.ts` — tag mode returns `defaultPrompt + "\n\n<custom_instructions>\n" + inputs.prompt + "\n</custom_instructions>"`.
3. The slash-command-capable multi-block path (`base-action/src/run-claude-sdk.ts`) reads `claude-user-request.txt`, written only when `extractUserRequestFromContext` finds the `@claude` trigger phrase in a comment or the issue/PR body. A `pull_request`-triggered review of a PR without `@claude` in its description produces `null`.
4. Verified against the CLI directly: a `/command` in leading position expands (and unreferenced arguments are appended even without `$ARGUMENTS`); the same command nested inside `<custom_instructions>` does not.

The PR cannot self-test this — its run was skipped with *"Workflow validation failed. The workflow file must exist and have identical content to the version on the repository's default branch."* The failure only appears after merge, and it fails open: a generic review still gets posted.

### Checked and clean

- **Arguments** — `argument-hint` without a matching `$1`/`$ARGUMENTS` in the body is fine; unreferenced args are appended to the prompt. Verified empirically.
- **`Read` in `--allowedTools`** — correctly required: the `@.claude/docs/review-threads.md` reference is resolved through the `Read` tool at command-load time.
- **Prompt fidelity** — the moved text is byte-identical to the old inline block apart from the target line and the extracted thread-hygiene section.
- **`actionlint`** — no new findings; YAML structure unchanged.
- **Comments** — nothing to flag; the workflow's existing comments all explain non-obvious *why* and stay within budget.
- Performance / testing / API docs: not applicable to this change.

### Nit

`cube-js/cube` is hardcoded in the command and `cube-js cube` in the docs, while `gh pr diff` / `gh pr comment` still resolve against the checkout's remote. Identical in CI; divergent in a fork clone, where the thread-hygiene step would query upstream PR numbers.

</details>

waralexrom (approved):

Issue #11545, opened 2026-08-13
Cube Store silently returns ~2x-inflated results for multi_stage measures whose stage references two measures, on longer date ranges


connor-wayne-mcelroy:
## Describe the bug

With Tesseract pre-aggregation serving enabled (`CUBEJS_TESSERACT_SQL_PLANNER=true`, `CUBEJS_TESSERACT_PRE_AGGREGATIONS=true`), a `multi_stage` measure whose stage SQL references **two** measures returns **silently inflated (~2x) values** for some months when the query's date range is long enough. No error — plausible-looking wrong numbers, served fast from Cube Store.

Minimal model shape (MS SQL source; a per-order "ticket count" allocated fractionally):

```yaml
# fact cube
- name: ticket_fraction     # line's share of its order's positive sales; sums to <= 1 per order
  sql: percent_of_total_positive_ticket_sales
  type: sum

# order-dimension cube joined many_to_one from the fact
- name: order_slice_gated
  multi_stage: true
  type: number
  sql: "CASE WHEN {fact.sale} > 0 THEN {fact.ticket_fraction} END"   # <-- TWO measures
  add_group_by: [id]

- name: tickets
  multi_stage: true
  type: sum
  sql: "{order_slice_gated}"
```

One rollup on the fact covering `sale`, `ticket_fraction`, the order key (`orders.id`), the query dimensions, and `time_dimension` on a conformed calendar cube, granularity day.

## Symptom

Same tenant, same month cell, month granularity — the answer depends on the surrounding date range:

| query range | value for July |
|---|---|
| Jul alone | 671.99 ✅ |
| Jun–Jul | 671.99 ✅ |
| Mar–Jul | 671.99 ✅ |
| **Apr–Jul** | **1341.99** ❌ |
| **Feb–Jul** | **1341.99** ❌ |

- The tenant had **692 orders with positive sales in July** — `tickets` is a sum of per-order fractions each ≤ 1, so 692 is a hard ceiling. 1341.99 is arithmetically impossible under the model's semantics.
- WHICH months inflate depends on the exact range: April is wrong in a 6-month query but right in 4- and 5-month queries; May is the reverse. Stable across repeated queries and rebuilt rollup tables for a given shape.
[...]

PR #11833, opened 2026-09-09
test(cubestore): failing test for the result corruption in #11545

connor-wayne-mcelroy:
Failing test for #11545, as requested.

`multi_stage_gated_join_with_sort_and_limit` in `rust/cubestore/cubestore/src/sql/mod.rs` is a self-contained Cube Store test — one table, plain `INSERT`s, no Cube, no pre-aggregation matching, no source database. It currently **fails**.

## What it asserts

A single rollup-shaped table holds 1100 orders in each of two months. The query is the SQL Cube's Tesseract planner emits verbatim for a `multi_stage` measure that gates one base measure on another (`CASE WHEN {sale} > 0 THEN {ticket_fraction} END` with `add_group_by`), with only the table name substituted: two per-key leaf aggregations over the same table, a `DISTINCT` key set built from their `UNION ALL`, a `LEFT JOIN` back to each leaf, and a top-level `ORDER BY ... LIMIT`.

Each order contributes exactly `1` to `ticket_fraction`, so every month must return `1100`.

| query | 2026-02 | 2026-03 |
|---|---|---|
| February alone | 1100 ✅ | — |
| March alone | — | 1100 ✅ |
| February–March | 1100 ✅ | **1401** ❌ |

Same table, same query text, same physical rows — only the two date literals differ. 1401 is larger than the number of distinct keys that feed the sum, which this query's algebra cannot produce: the key set is `SELECT DISTINCT`, each leaf is pre-grouped to one row per `(order, month)`, and the joins are on the full composite key.

## What narrows it down

While reducing this from a production model, three things turned out to be load-bearing:

- **The top-level `ORDER BY` and `LIMIT` together.** Removing either one — the `ORDER BY`, or the `LIMIT` — makes the same query return correct values. That points at the sort/limit pushdown into `ClusterSend` (`pull_up_cluster_send`'s `LogicalPlan::Sort` branch and the `worker_sort_and_limit` descriptor) rather than at the join. `LIMIT 10000` is far above the 2-row result, so it is the pushdown, not the limit value.
- **Crossing a record batch.** At ≤ 2048 distinct `(order, month)` keys the query is always correct; at 2049+ it corrupts. The test sits just over that line at 2200 keys. This is why the bug is invisible on short date ranges and why *which* months are wrong shifts with the width of the range asked for.
- **Scan order disagreeing with the leaves' `ORDER BY`.** Order ids interleave across the two months, so index order (by id) and the leaves' `ORDER BY month` disagree, as they do for real order ids. With ids assigned month-by-month (so the two orders agree) the same query returns correct values.

Row counts are right throughout — `SELECT count(*)` over the key set and over each join stage all return 2200. Only the summed values are wrong, so nothing duplicates rows; the join pairs keys with values that do not belong to them.

## One note on the test harness

The test runs its runtime on an explicitly sized thread. Planning this query shape recurses deeply enough to overflow libtest's default 2 MiB stack in a debug build — it aborts the whole test binary with `fatal runtime error: stack overflow` before reaching any assertion, which is why the explicit stack is there. That looks unrelated to the corruption (release frames are much smaller, and production gives select workers 4 MiB via `CUBESTORE_SELECT_WORKER_STACK_SIZE`), but it may be the same deep-recursion path as the "recursion limit reached" crash, so flagging it rather than burying it.

Current output:

```
assertion failed: `(left == right)`

Diff < left / right > :
 [
     1100,
<    1401,
>    1100,
 ]
```

## Versions

Reproduces identically on `cubejs/cubestore:v1.7.4`, `v1.7.19` and `latest`, single-node and router+workers, on freshly built tables. Decimal measures behave the same as the ints used here; the test uses ints to keep the assertions readable.

Signed-off-by: Connor McElroy <connor.mcelroy@xyzz.dev>

### DIFF
--- a/rust/cubestore/cubestore/src/sql/mod.rs
+++ b/rust/cubestore/cubestore/src/sql/mod.rs
@@ -3667,6 +3667,194 @@ mod tests {
         Ok(())
     }
 
+    /// Regression test for https://github.com/cube-js/cube/issues/11545.
+    ///
+    /// Cube's Tesseract planner emits this shape for a `multi_stage` measure whose SQL
+    /// gates one base measure on another (`CASE WHEN {a} > 0 THEN {b} END` with
+    /// `add_group_by`): two per-key leaf aggregations over the same rollup table, a
+    /// DISTINCT key set built from their UNION ALL, a LEFT JOIN back to each leaf, and a
+    /// top-level `ORDER BY ... LIMIT`.
+    ///
+    /// Each single-month query returns the correct 1100. Widening the range so the join
+    /// input crosses a record batch boundary (> 2048 keys) makes the *same* query over the
+    /// *same* table return 1401 for March -- a sum larger than the number of keys that
+    /// feed it, which is impossible under this query's algebra.
+    ///
+    /// The corruption requires BOTH the top-level ORDER BY and the LIMIT. Removing either
+    /// one returns correct values, which points at the sort/limit pushdown into
+    /// ClusterSend rather than at the join itself. Reproduced on v1.7.4, v1.7.19 and the
+    /// current `latest` image; decimal measures behave identically to the int ones here.
+    #[test]
+    fn multi_stage_gated_join_with_sort_and_limit() {
+        // Planning this query recurses deeply enough to overflow libtest's default 2 MiB
+        // stack in a debug build (production gives select workers 4 MiB via
+        // CUBESTORE_SELECT_WORKER_STACK_SIZE, and release frames are much smaller), so run
+        // it on a thread sized for debug frames. Unrelated to the corruption below.
+        std::thread::Builder::new()
+            .stack_size(32 * 1024 * 1024)
+            .spawn(|| {
+        tokio::runtime::Builder::new_multi_thread()
+            .enable_all()
+            .thread_stack_size(32 * 1024 * 1024)
+            .build()
+            .unwrap()
+            .block_on(async {
+        Config::test("multi_stage_gated_join_with_sort_and_limit")
+            .update_config(|mut c| {
+                // Keep the rollup in a single partition, as a real unpartitioned
+                // pre-aggregation table is.
+                c.partition_split_threshold = 1000000;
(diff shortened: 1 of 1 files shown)


paveltiunov:
Superseded by #11886, which merged on Sep 16 and fixes #11545. Released in **v1.7.40**.

Your test landed in master as part of that PR, unchanged and with your `Signed-off-by` preserved. I ran it on the current master just now:

```
test sql::tests::multi_stage_gated_join_with_sort_and_limit ... ok
test result: ok. 1 passed; 0 failed
```

**The root cause, and one correction to the analysis in this PR.** `ChooseIndexContext` propagated the query's top-level `ORDER BY`/`LIMIT` down the whole logical plan without resetting at any boundary, so the limit was attributed to aggregates nested arbitrarily deep — inside subqueries, CTEs and join branches. In your shape that reordered the inner CTEs to `(month, id)` while the merge above them still merged by `id`, so the streaming `DISTINCT` stopped seeing equal keys adjacent and emitted 301 duplicate keys — hence 1401 instead of 1100.

So the `DISTINCT` was emitting duplicate rows; the join was not mispairing values. Your "row counts are correct throughout" observation was measured at the join stages rather than at the dedup itself, which is what made it read as a pairing bug. Your three ingredients were all correct and load-bearing, and the `pull_up_cluster_send` / sort-and-limit pointer was the right place to look.

Two things worth knowing for your deployment:

- The blast radius was wider than `multi_stage`: any query whose plan deduplicates a key set under the `ORDER BY 1 ASC LIMIT 10000` that Cube always emits could hit this, not just multi-stage measures. Worth a re-check of any aggregate you validated on short ranges only, since ≤2048 keys never corrupts.
- #11886 adds `CUBESTORE_LIMIT_PUSHDOWN` (default on) as an escape hatch, if you ever want to rule the optimization out without rolling back.

Closing this as superseded. Thank you for the reduction to a standalone test — no join, no Cube, no pre-aggregation matching — it is what made this findable, and it is now permanent regression coverage for a shape Cube Store's suite had none of.

---

